PR #14, opened 2021-04-27
Send event every 1 sec instead of 100 millis in Kafka

Sgitario:


Sgitario:
Closing as this issue is still being happening.

PR #39, opened 2021-05-20
Add sql-db/sql-app devServices scenario

pjgg:
- Add `sql-db/sql-app` devServices scenario

### DIFF
--- a/sql-db/sql-app/src/test/java/io/quarkus/ts/openshift/sqldb/DevModePostgresqlIT.java
+++ b/sql-db/sql-app/src/test/java/io/quarkus/ts/openshift/sqldb/DevModePostgresqlIT.java
@@ -1,0 +1,19 @@
+package io.quarkus.ts.openshift.sqldb;
+
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.junit.DisabledOnNativeImage;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.DevModeQuarkusApplication;
+
+@QuarkusScenario
+@DisabledOnNativeImage
+public class DevModePostgresqlIT extends AbstractSqlDatabaseIT {
+
+    @DevModeQuarkusApplication
+    static RestService app = new RestService().withProperties("postgresql.properties");
+
+    @Override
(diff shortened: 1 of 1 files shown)


Sgitario (changes requested):
The error in the actions will be fixed in main.
However, I would not cover dev services in multiple-pus as we would be losing the Native coverage (dev services is not compatible with Native).
I would simply add a new scenario in the `sql-db/sql-app` module.

sql-db/sql-app/src/test/java/io/quarkus/ts/openshift/sqldb/PostgresqlDevServicesIT.java
```diff
@@ -0,0 +1,19 @@
+package io.quarkus.ts.openshift.sqldb;
+
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.junit.DisabledOnNativeImage;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.DevModeQuarkusApplication;
+
+@QuarkusScenario
```
Sgitario, line 8: I've proposed to name the tests using DevMode starting with `DevMode`. See https://github.com/quarkus-qe/quarkus-test-suite/pull/41/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R87.
What do you think?

.github/workflows/daily.yaml
```diff
@@ -68,6 +68,9 @@ jobs:
         run: mvn -s .github/mvn-settings.xml clean install -pl .
       - name: Build Quarkus QE Test Framework
         run: git clone https://github.com/quarkus-qe/quarkus-test-framework && cd quarkus-test-framework && mvn -B -s .github/mvn-settings.xml clean install -Pframework
+        # Install parent as other modules might need it
+      - name: Install Test Suite Parent
+        run: mvn -s .github/mvn-settings.xml clean install -pl .
```
Sgitario, line 73: This is duplicated, It's not needed it.

Sgitario (changes requested):

Sgitario (approved):

PR #40, opened 2021-05-21
Upgrade quarkus to 2.0.0.Alpha3

Sgitario:


Sgitario:
@pjgg I've dropped the tests in http-advanced that verified the redirection of non app endpoints as this is no longer supported. Let me know if you prefer to change them or do something different.

pjgg (approved):

PR #66, opened 2021-07-19
Misc cleanup in the code

rsvoboda:
Misc cleanup in the code

- Fix JavaDoc link
- Drop unnecessary .toString()
- Use long to avoid Long.valueOf() method call


Sgitario (approved):
lgtm!

Sgitario:
Failures will be fixed in a separate PR.
Merging.

Issue #82, opened 2021-07-21
Migrate 011-quarkus-panache-rest-flyway and 014-quarkus-panache-with-transactions-xa from the Quarkus Beefy Test Suite


Sgitario:
The new module should be part of sql-db parent and both modules should be one in the new test suite.

PR #130, opened 2021-07-28
Migrate 011 and 014 into sql-db

Sgitario:
Migrate 011-quarkus-panache-rest-flyway and 014-quarkus-panache-with-transactions-xa into sql-db/panache-flyway.
Module that test whether we can setup a REST API using ORM Panache with Flyway and a MySQL database. Moreover, it covers XA transactions.

Topics covered here:
- https://quarkus.io/guides/hibernate-orm-panache
- https://quarkus.io/guides/rest-data-panache
- https://quarkus.io/guides/flyway

Base application:
- Define multiple datasources (https://quarkus.io/guides/datasource#multiple-datasources): default (transactions enabled), `with-xa` (transactions xa).
- Define Panache entity `ApplicationEntity` and its respective REST resource `ApplicationResource` (https://quarkus.io/guides/rest-data-panache).
- Define a REST resource `DataSourceResource` that provides info about the datasources.

Additional tests:
- Rest Data with Panache test according to https://github.com/quarkus-qe/quarkus-test-plans/blob/main/QUARKUS-976.md  
Additional UserEntity is a simple JPA entity that was created with aim to avoid inheritance of PanacheEntity methods
and instead test the additional combination of JPA entity + PanacheRepository + PanacheRepositoryResource, where
PanacheRepository is a facade class. Facade class can override certain methods to change the default behaviour of the
PanacheRepositoryResource methods.

- AgroalPoolTest, will cover how the db pool is managed in terms of IDLE-timeout, max connections and concurrency.

Fix https://github.com/quarkus-qe/quarkus-test-suite/issues/82

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,29 @@ Verifies that the application can connect to multiple SQL databases and persist 
 The application also uses RESTEasy to expose a RESTful API, Jackson for JSON serialization, and Hibernate Validator to validate inputs.
 The multiple persistence units are defined in `application.properties`.
 
+### `sql-db/panache-flyway`
+
+Module that test whether we can setup a REST API using ORM Panache with Flyway and a MySQL database. Moreover, it covers XA transactions.
+
+Topics covered here:
+- https://quarkus.io/guides/hibernate-orm-panache
+- https://quarkus.io/guides/rest-data-panache
+- https://quarkus.io/guides/flyway
+
+Base application:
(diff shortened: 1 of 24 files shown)


Sgitario:
I didn't configure any OpenShift scenario as I don't think this coverage is relevant in OpenShift.
I left AgroalPoolTest as a `@QuarkusTest` as this is a performance test instead of a functional test.
cc @mjurc

mjurc (approved):
LGTM!

Issue #78, opened 2021-07-21
Migrate 005-quarkus-and-maven-profiles from the Quarkus Beefy Test Suite


Sgitario:

rsvoboda:
I will cover this one.

PR #152, opened 2021-08-05
Quarkus and Maven profiles coverage

rsvoboda:
Quarkus and Maven profiles coverage

Enhancing `lifecycle-application` module to save execution time.

Fixes https://github.com/quarkus-qe/quarkus-test-suite/issues/78 / migrates https://github.com/quarkus-qe/beefy-scenarios/tree/main/005-quarkus-and-maven-profiles

### DIFF
--- a/lifecycle-application/pom.xml
+++ b/lifecycle-application/pom.xml
@@ -10,6 +10,11 @@
     <artifactId>lifecycle-application</artifactId>
     <packaging>jar</packaging>
     <name>Quarkus QE TS: Lifecycle Application</name>
+    <properties>
+        <!-- Using invalid Quarkus version to make sure the one defined in profile gets picked -->
+        <!-- Quarkus maven plugin had troubles to use stuff defined in profile -->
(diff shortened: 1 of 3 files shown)


lifecycle-application/pom.xml
```diff
@@ -10,6 +10,9 @@
     <artifactId>lifecycle-application</artifactId>
     <packaging>jar</packaging>
     <name>Quarkus QE TS: Lifecycle Application</name>
+    <properties>
+        <quarkus.platform.version>9999999-SNAPSHOT</quarkus.platform.version>
```
Sgitario, line 14: Can you add a comment here to state the purpose of setting this value here?

Sgitario (changes requested):

lifecycle-application/pom.xml
rsvoboda: Updated

Sgitario (approved):

PR #153, opened 2021-08-06
Improve Native time on Daily action

Sgitario:
++ Run only http/http-minimum in Windows native only
++ Disable AnnotationScheduledJobsMySqlQuartzIT on Native as it takes around 30 min to finish
++ There were some duplicated test cases in QuarkusCliVersionIT and QuarkusCliCreateJvmApplicationIT

### DIFF
--- a/quarkus-cli/pom.xml
+++ b/quarkus-cli/pom.xml
@@ -27,6 +27,7 @@
                 </property>
             </activation>
             <properties>
+                <!-- To not build the module on Native -->
                 <quarkus.package.type>fast-jar</quarkus.package.type>
             </properties>
         </profile>
--- a/quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/QuarkusCliVersionIT.java
+++ b/quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/QuarkusCliVersionIT.java
@@ -1,23 +1,15 @@
 package io.quarkus.ts.quarkus.cli;
 
-import static io.quarkus.test.utils.AwaitilityUtils.untilAsserted;
-import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-
-import java.util.List;
-import java.util.stream.Stream;
 
 import javax.inject.Inject;
 
-import org.apache.http.HttpStatus;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.condition.EnabledIfSystemProperty;
+import org.junit.jupiter.api.condition.DisabledIfSystemProperty;
 
 import io.quarkus.builder.Version;
 import io.quarkus.test.bootstrap.QuarkusCliClient;
-import io.quarkus.test.bootstrap.QuarkusCliRestService;
 import io.quarkus.test.scenarios.QuarkusScenario;
 import io.quarkus.test.scenarios.annotations.DisabledOnQuarkusVersion;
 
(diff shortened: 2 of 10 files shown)


Sgitario:
These are the scenario that took longer:

```
Time elapsed: 432.972 s - in io.quarkus.ts.sqldb.sqlapp.Postgresql10DatabaseIT
Time elapsed: 826.375 s - in io.quarkus.ts.sqldb.sqlapp.MariaDB102DatabaseIT
Time elapsed: 328.757 s - in io.quarkus.qe.sqldb.panacheflyway.ApplicationResourceIT
Time elapsed: 354.225 s - in io.quarkus.qe.sqldb.panacheflyway.UserResourceIT
Time elapsed: 1,941.858 s - in io.quarkus.qe.scheduling.quartz.AnnotationScheduledJobsMySqlQuartzIT
Time elapsed: 384.336 s - in io.quarkus.ts.quarkus.cli.QuarkusCliCreateApplicationIT
Time elapsed: 1,217.108 s - in io.quarkus.ts.quarkus.cli.QuarkusCliSpecialCharsIT
Time elapsed: 392.366 s - in io.quarkus.ts.quarkus.cli.QuarkusCliVersionIT
```

The CLI scenarios are building the apps always on Native and that's why they took so much. For now, I think we're good to not run these tests on Native.

rsvoboda (changes requested):
++ Not run CLI jobs on Native
add parallel run for CLI tests

++ Disable AnnotationScheduledJobsMySqlQuartzIT on Native as it takes around 30 min to finish
ok for now, but we should think about a way how to run the tests which were disabled just because they took too much time. Please create issue for that

Sgitario:
> ++ Not run CLI jobs on Native
> add parallel run for CLI tests
> 
> ++ Disable AnnotationScheduledJobsMySqlQuartzIT on Native as it takes around 30 min to finish
> ok for now, but we should think about a way how to run the tests which were disabled just because they took too much time. Please create issue for that

> add parallel run for CLI tests

Let's speak about it in a private chat.

> ok for now, but we should think about a way how to run the tests which were disabled just because they took too much time. Please create issue for that

I will tag those tests and create an issue to think about how to run these scenarios.

Sgitario:
> > ++ Not run CLI jobs on Native
> > add parallel run for CLI tests
> > ++ Disable AnnotationScheduledJobsMySqlQuartzIT on Native as it takes around 30 min to finish
> > ok for now, but we should think about a way how to run the tests which were disabled just because they took too much time. Please create issue for that
> 
> > add parallel run for CLI tests
> 
> Let's speak about it in a private chat.
> 
> > ok for now, but we should think about a way how to run the tests which were disabled just because they took too much time. Please create issue for that
> 
> I will create an issue to think about how to run these scenarios: #154

About the CLI tests, the problem is that when building on Native, all the builds were done in Native but this was not meant to behave like this. Instead, all the tests should be run on JVM but a few, to verify build native. 
I've already updated the scenario to cover JVM and Native in the same time.

quarkus-cli/pom.xml
```diff
@@ -27,8 +27,26 @@
                 </property>
             </activation>
             <properties>
+                <!-- To not build the module on Native -->
                 <quarkus.package.type>fast-jar</quarkus.package.type>
             </properties>
+            <!-- Skip on Native as JVM run will cover the Native build too -->
```
rsvoboda, line 33: This can be a bit problem for Jenkins jobs when I expect to have JVM and Native variants of the jobs.

This way we would need to specify Mandrel image for JVM job too ?
And there would be no cli tests executed on NATIVE job?

Sgitario: No CLI tests executed on Native job.
But good point about the Mandrel image.. we're losing this coverage

quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/QuarkusCliVersionIT.java
```diff
@@ -42,77 +31,4 @@ public void shouldVersionMatchQuarkusVersion() {
         // Using shortcut
         assertEquals("Client Version " + Version.getVersion(), cliClient.run("-v").getOutput());
     }
-
-    @Test
-    public void shouldCreateApplicationOnJvm() {
```
Sgitario, line 47: I've just realised that these scenarios were duplicated. We should keep the ones in QuarkusCliCreateJvmApplicationIT

rsvoboda (approved):

rsvoboda:
waiting for CI before merging

PR #161, opened 2021-08-09
Fix properties module by removing the Person field

Sgitario:
Quarkus upstream upgraded SmallRye Config to 2.4.3 and it seems that there is no longer possible to use the annotation @ConfigMapping on classes, only in interfaces.
As we're already covering the interface mapping, I removed the field coverage.
Related issue: https://github.com/quarkusio/quarkus/issues/19298


Sgitario:
Similar to https://github.com/quarkus-qe/beefy-scenarios/pull/246

kshpak (approved):

Issue #170, opened 2021-08-12
Add coverage for `quarkus-rest-client-jaxb` and `quarkus-rest-client-jsonb`


Sgitario:
Related to https://github.com/quarkus-qe/quarkus-test-suite/issues/157

PR #171, opened 2021-08-12
Add coverage for `rest-client-jaxb` and `rest-client-jsonb`

Sgitario:
Verifies Rest Client configuration using `quarkus-rest-client-jaxb` (XML support) and `quarkus-rest-client-jsonb` (JSON support).
This module will setup a very minimal configuration (only `quarkus-resteasy`) and have four endpoints:
- Two endpoints to get a book in JSON and XML formats.
- Two endpoints to get the value of the previous endpoints using the rest client interface.

Fix https://github.com/quarkus-qe/quarkus-test-suite/issues/170

### DIFF
--- a/http/rest-client/src/main/java/io/quarkus/ts/http/restclient/RestInterface.java
+++ b/http/rest-client/src/main/java/io/quarkus/ts/http/restclient/RestInterface.java
@@ -1,0 +1,27 @@
+package io.quarkus.ts.http.restclient;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+import org.eclipse.microprofile.rest.client.annotation.RegisterClientHeaders;
+import org.eclipse.microprofile.rest.client.inject.RegisterRestClient;
+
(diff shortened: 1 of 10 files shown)


http/rest-client/src/main/java/io/quarkus/ts/http/restclient/RestInterface.java
```diff
@@ -0,0 +1,27 @@
...
+import javax.ws.rs.core.MediaType;
+
+import org.eclipse.microprofile.rest.client.annotation.RegisterClientHeaders;
+import org.eclipse.microprofile.rest.client.inject.RegisterRestClient;
+
+import io.quarkus.ts.http.restclient.xml.Book;
+
+@RegisterRestClient
+@Path("/book")
+@RegisterClientHeaders
```
jsmrcka, line 15: Just a question: does this annotation do anything if there are no headers specified by a `org.eclipse.microprofile.rest.client.propagateHeaders` property? Based on the way we use it, I assume it by default propagates all incoming headers, but the documentation does not seem entirely clear to me.

jsmrcka (approved):
LGTM. ~~Please rebase.~~ Thx.

Issue #98, opened 2021-07-21
Migrate 603-spring-web-smallrye-openapi from the Quarkus Beefy Test Suite


Sgitario:

jsmrcka:
https://github.com/quarkusio/quarkus/issues/16321 has been fixed => resolve this TODO when migrating the module:  
https://github.com/quarkus-qe/beefy-scenarios/pull/162/files#diff-741ead5554ce7e064b608c90448dd2ad6e71d8386fe06e7e5c13b9fa3a02ae91R71

PR #177, opened 2021-08-15
Migrate 603-spring-web-smallrye-openapi from BeefyTS

jsmrcka:
Closes https://github.com/quarkus-qe/quarkus-test-suite/issues/98


spring/spring-web/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkus.ts.qe</groupId>
+        <artifactId>parent</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+        <relativePath>../..</relativePath>
+    </parent>
+    <artifactId>spring-web</artifactId>
+    <packaging>jar</packaging>
+    <name>Quarkus QE TS: spring-web</name>
```
Sgitario, line 12:
```suggestion
    <name>Quarkus QE TS: Spring: Spring Web</name>
```

Sgitario (changes requested):
The windows action is failing because this module is using testcontainers which is not supported in Windows.
You need to add this profile in the pom.xml of this new module: https://github.com/quarkus-qe/quarkus-test-suite/blob/909dfb5188baf4deb823a25b9a8dc738ed0be683/sql-db/panache-flyway/pom.xml#L43

spring/spring-web/pom.xml
jsmrcka: Thanks for pointing it out. I also fixed the name for previously migrated `spring/spring-data` module and added it to this PR.

Sgitario (approved):

Sgitario:
run tests

Issue #175, opened 2021-08-13
Fix test: PostgresPoolIT.checkDbPoolTurnover in JVM build


Sgitario:
This test is randomly failing when running on JVM.

Originally, this test was not meant to be built on Native (check it out in the Beefy repository)

Sgitario:
Example of random failure: https://github.com/quarkus-qe/quarkus-test-suite/actions/runs/1133658704

Sgitario:
fyi @pjgg and @fedinskiy

fedinskiy:
@Sgitario according to linked logs the test fails on non-native(JVM build), but this is a problem anyway. I guess, the best solution would be to keep that test in beefy TS and remove it from the new one.

Sgitario:
> @Sgitario according to linked logs the test fails on non-native(JVM build), but this is a problem anyway. I guess, the best solution would be to keep that test in beefy TS and remove it from the new one.

This test is not running on Beefy TS on Native (only in JVM) and it's not related to your migrated module (that's why I copied also to @pjgg )
In this test suite, what I did is to disable the test only on Native, so now beefy TS and this TS are behaving equally. 
But we need to figure out why is randomly failing.

fedinskiy:
> In this test suite, what I did is to disable the test only on Native, so now beefy TS and this TS are behaving equally.
We can disable it on Native, yes, but it will fail on JVM anyway, just like it did in provided link [1]. 

To port that test from beefy to the new TS, I had to put a part of this test to a separate application, and it looks like communication over the network between the test and the application breaks some assumptions this test was based on. So we either have to change the test framework to allow injection in test classes, or move the test back to beefy.

[1] https://github.com/quarkus-qe/quarkus-test-suite/actions/runs/1133658704

Sgitario:
> > In this test suite, what I did is to disable the test only on Native, so now beefy TS and this TS are behaving equally.
> > We can disable it on Native, yes, but it will fail on JVM anyway, just like it did in provided link [1].
> 
> To port that test from beefy to the new TS, I had to put a part of this test to a separate application, and it looks like communication over the network between the test and the application breaks some assumptions this test was based on. So we either have to change the test framework to allow injection in test classes, or move the test back to beefy.
> 
> [1] https://github.com/quarkus-qe/quarkus-test-suite/actions/runs/1133658704

What we need is to clearly identify the reasons why this test is randomly failing:
[...]

PR #181, opened 2021-08-16
Fix test: PostgresPoolIT.checkDbPoolTurnover

Sgitario:
The problem was caused by calling the Reactive Route using RESTAssured instead of using the WebClient instance. 
Calling using RestAssured was not waiting for the call to be ended and hence sometimes the response was empty (the number of active connections).
The rest of the changes are cosmetic, so this test looks more similar to what we have in Beefy.
I did check that now it's working on Native and JVM.

Fix https://github.com/quarkus-qe/quarkus-test-suite/issues/175

### DIFF
--- a/sql-db/vertx-sql/src/main/java/io/quarkus/qe/vertx/sql/PoolApplication.java
+++ b/sql-db/vertx-sql/src/main/java/io/quarkus/qe/vertx/sql/PoolApplication.java
@@ -29,7 +29,7 @@ public class PoolApplication {
                 .transform(iterator -> iterator.hasNext() ? iterator.next().getLong("active_con") : null);
     }
 
-    @Route(methods = Route.HttpMethod.PUT, path = "/connect")
+    @Route(methods = Route.HttpMethod.GET, path = "/connect")
     public Uni<Long> newConnection() {
         // Make just one extra query and Hold "Idle + 1 sec" in order to release inactive connections.
         return postgresql.preparedQuery("SELECT CURRENT_TIMESTAMP").execute()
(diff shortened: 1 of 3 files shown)


sql-db/vertx-sql/src/test/java/io/quarkus/qe/vertx/sql/dbpool/PostgresPoolIT.java
```diff
@@ -127,27 +120,45 @@ public void checkIdleExpirationTime() throws InterruptedException {
...
+            if (connectionsAmount == 1) {
+                doneIdleExpired.countDown();
+            }
+        });
+
+        doneIdleExpired.await(TIMEOUT_SEC, TimeUnit.SECONDS);
+        assertEquals(doneIdleExpired.getCount(), 0, "Missing doneIdleExpired query.");
+    }
+
+    private Uni<Long> activeConnections() {
```
Sgitario, line 139: @fedinskiy fyi, this is the change that is making the test worked in JVM and Native both.

sql-db/vertx-sql/src/main/java/io/quarkus/qe/vertx/sql/PoolApplication.java
```diff
@@ -29,7 +29,7 @@
                 .transform(iterator -> iterator.hasNext() ? iterator.next().getLong("active_con") : null);
     }
 
-    @Route(methods = Route.HttpMethod.PUT, path = "/connect")
+    @Route(methods = Route.HttpMethod.GET, path = "/connect")
```
fedinskiy, line 32: We change state of the object with that request. It's not recommended to use GET requests for that

Sgitario: We are not changing the state of the application, but just doing two queries: one to get the current timestamp, another one to get the number of active connections.

fedinskiy: Yes, and "getting current timestamp" is also "adding another connection to count".

Issue #154, opened 2021-08-06
Investigate approach about how to run tests that requires high execution times


Sgitario:
An example of a scenario that requires a lot of execution time to run is: https://github.com/quarkus-qe/quarkus-test-suite/blob/main/scheduling/quartz/src/test/java/io/quarkus/qe/scheduling/quartz/AnnotationScheduledJobsMySqlQuartzIT.java

When running this test on Native, it will build the Native app three times (because we're selecting the classes to include).

Therefore, we need to find an approach to run these scenarios.

PR #183, opened 2021-08-17
Split Native job to run selected modules

Sgitario:
The idea would be to run the modules that take the most of the time in individual jobs, so the jobs do not exceed the timeout.

++ Update Mandrel image to use 21.2
++ Update reason of @DisabledOnNative for S2i deployments. Caused by: https://github.com/quarkus-qe/quarkus-test-framework/issues/206
++ Exclude DevMode tests on Native by Maven configuration
++ Refactor scheduling/quartz module to run tests on Native without build
++ Avoid to use RandomStringUtils in micrometer/prometheus, fails on Native

Fix https://github.com/quarkus-qe/quarkus-test-suite/issues/154

### DIFF
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -76,7 +76,7 @@ jobs:
           chmod +x ./quarkus-dev-cli
       - name: Test in JVM mode
         run: |
-          mvn -fae -V -B -s .github/mvn-settings.xml -fae clean verify -Dinclude.quarkus-cli-tests -Dts.quarkus.cli.cmd="${PWD}/quarkus-dev-cli"
+          mvn -fae -V -B -s .github/mvn-settings.xml -fae clean verify -Dall-modules -Dinclude.quarkus-cli-tests -Dts.quarkus.cli.cmd="${PWD}/quarkus-dev-cli"
       - name: Zip Artifacts
         if: failure()
         run: |
@@ -94,7 +94,12 @@ jobs:
     strategy:
       matrix:
         java: [ 11 ]
-        image: [ "ubi-quarkus-native-image:21.0-java11", "ubi-quarkus-mandrel:21.0-java11" ]
+        image: [ "ubi-quarkus-native-image:21.2-java11", "ubi-quarkus-mandrel:21.2-java11" ]
+        profiles: [ "!http-modules,!security-modules,!sql-db-modules,!messaging-modules",
+                   "http-modules",
+                   "security-modules",
+                   "sql-db-modules",
+                   "messaging-modules"]
     steps:
       - uses: actions/checkout@v1
       - uses: actions/cache@v1
@@ -129,10 +134,7 @@ jobs:
           chmod +x ./quarkus-dev-cli
       - name: Test in Native mode
         run: |
-          # Skip module due to high time execution on Native
-          MODULES="-pl '!sql-db/sql-app'"
-
-          mvn -fae -V -B -s .github/mvn-settings.xml $MODULES -fae clean verify -Dnative \
+          mvn -fae -V -B -s .github/mvn-settings.xml -P ${{ matrix.profiles }} -fae clean verify -Dnative \
             -Dquarkus.native.builder-image=quay.io/quarkus/${{ matrix.image }} \
             -Dinclude.quarkus-cli-tests -Dts.quarkus.cli.cmd="${PWD}/quarkus-dev-cli"
       - name: Zip Artifacts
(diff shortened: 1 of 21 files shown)


Sgitario:
Running Daily run test in https://github.com/Sgitario/quarkus-test-suite/runs/3347279091?check_suite_focus=true

.github/workflows/daily.yaml
```diff
@@ -94,7 +94,9 @@ jobs:
     strategy:
       matrix:
         java: [ 11 ]
-        image: [ "ubi-quarkus-native-image:21.0-java11", "ubi-quarkus-mandrel:21.0-java11" ]
+        image: [ "ubi-quarkus-native-image:21.0-java11", "ubi-quarkus-mandrel:21.2-java11" ]
```
rsvoboda, line 97:
```suggestion
        image: [ "ubi-quarkus-native-image:21.2-java11", "ubi-quarkus-mandrel:21.2-java11" ]
```

.github/workflows/quarkus-test-framework-snapshot.yaml
```diff
@@ -95,7 +95,7 @@ jobs:
     strategy:
       matrix:
         java: [ 11 ]
-        image: [ "ubi-quarkus-native-image:21.0-java11", "ubi-quarkus-mandrel:21.0-java11" ]
+        image: [ "ubi-quarkus-native-image:21.0-java11", "ubi-quarkus-mandrel:21.2-java11" ]
```
rsvoboda, line 98:
```suggestion
        image: [ "ubi-quarkus-native-image:21.2-java11", "ubi-quarkus-mandrel:21.2-java11" ]
```

external-applications/src/test/java/io/quarkus/ts/external/applications/OpenShiftQuickstartUsingS2iIT.java
```diff
@@ -9,7 +9,7 @@
 import io.quarkus.test.scenarios.annotations.DisabledOnQuarkusSnapshot;
 import io.quarkus.test.services.GitRepositoryQuarkusApplication;
 
-@DisabledOnNative
+@DisabledOnNative(reason = "S2i is not supported on Native yet: https://github.com/quarkus-qe/quarkus-test-framework/issues/206")
```
rsvoboda, line 12: I think S2I for native needed some special Quarkus image which was supported just on community side. On product side native build using S2I wasn't supported. 

We used that community image for Native S2I in OpenShiftTS. @mjurc can you point to the details?
Ideally update https://github.com/quarkus-qe/quarkus-test-framework/issues/206

rsvoboda (changes requested):

.github/workflows/quarkus-test-framework-snapshot.yaml
Sgitario: Updated

.github/workflows/daily.yaml
Sgitario: Updated

Sgitario:
@rsvoboda this is ready to be reviewed now.
The main implications are that we have moved the modules into profiles, so we can now easily select and unselect groups quite. After these changes, the daily run is finally finished ok in time.
The disadvantage is that when we select other profiles (for example, openshift or native), we need to explicitly add the property `-Dinclude.all-modules`. 
I didn't update the documentation or update the Jenkins jobs because I wanted to know what you think about the current solution.

rsvoboda (changes requested):
`-Dinclude.all-modules` next to `-Dinclude.quarkus-cli-tests` doesn't feel right to me. I would just use `-Dall-modules`.
I think the split is needed and I don't think Maven gives us better ways how to handle combinations of profiles with default ones.

Please update the README

https://github.com/quarkus-qe/quarkus-test-framework/pull/208 was merged, `@DisabledOnNative` for S2i deployments could be removed imho.

Sgitario:
> `-Dinclude.all-modules` next to `-Dinclude.quarkus-cli-tests` doesn't feel right to me. I would just use `-Dall-modules`.
> I think the split is needed and I don't think Maven gives us better ways how to handle combinations of profiles with default ones.
> 
> Please update the README
> 
> [quarkus-qe/quarkus-test-framework#208](https://github.com/quarkus-qe/quarkus-test-framework/pull/208) was merged, `@DisabledOnNative` for S2i deployments could be removed imho.

Changed `-Dinclude.all-modules` to `-Dall-modules`.
I didn't remove the disabled on native for S2i deployments as this should be done after upgrading the test framework to 0.0.7. Moreover, we need to map the `quarkus.s2i.base-jvm-image` and `quarkus.s2i.base-native-image` properties to work properly.

Just a note, we could improve the situation by removing the rest of the profiles (openshift, native, validate-format, quarkus-cli-tests, operator-scenarios, include.serverless, redhat-registry). These profiles are only about to update some properties, so we could provide these properties directly to the Maven command instead of using a property to activate the Maven profile. This way we don't need to explicitly add the `-Dall-modules` almost all the times.

rsvoboda:
I would keep openshift, native, validate-format, quarkus-cli-tests, operator-scenarios, include.serverless, redhat-registry profiles, they are handy and sort of abstraction if we need to adjust the underlying properties

rsvoboda (approved):

Issue #184, opened 2021-08-17
Add coverage for `quarkus-spring-cloud-config-client`


Sgitario:
Related to https://github.com/quarkus-qe/quarkus-test-suite/issues/157

PR #185, opened 2021-08-17
Add module for `spring/spring-cloud-config`

Sgitario:
Verifies that we can use an external Spring Cloud Server to inject configuration in our Quarkus applications.
Fix https://github.com/quarkus-qe/quarkus-test-suite/issues/184

### DIFF
--- a/spring/spring-cloud-config/src/test/java/io/quarkus/ts/spring/cloud/config/SpringCloudConfigIT.java
+++ b/spring/spring-cloud-config/src/test/java/io/quarkus/ts/spring/cloud/config/SpringCloudConfigIT.java
@@ -1,0 +1,34 @@
+package io.quarkus.ts.spring.cloud.config;
+
+import static org.hamcrest.Matchers.is;
+
+import org.apache.http.HttpStatus;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.Container;
+import io.quarkus.test.services.QuarkusApplication;
+
+@QuarkusScenario
+public class SpringCloudConfigIT {
+
+    @Container(image = "${spring.cloud.server.image}", port = 8888, expectedLog = "Started ConfigServer")
+    static RestService spring = new RestService()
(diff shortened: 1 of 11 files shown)


Sgitario:
@jsmrcka would you mind taking a look into this PR?

spring/spring-cloud-config/src/test/java/io/quarkus/ts/spring/cloud/config/SpringCloudConfigIT.java
```diff
@@ -0,0 +1,34 @@
...
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.Container;
+import io.quarkus.test.services.QuarkusApplication;
+
+@QuarkusScenario
+public class SpringCloudConfigIT {
+
+    @Container(image = "${spring.cloud.server.image}", port = 8888, expectedLog = "Started ConfigServer")
+    static RestService spring = new RestService()
+            .withProperty("SPRING_PROPERTY_FILE", "resource::/config/application-SpringCloudConfigIT.properties")
```
jsmrcka, line 19: I was unable to find any mention of `SPRING_PROPERTY_FILE` anywhere. Can you perhaps point me in the right direction?

jsmrcka (approved):

spring/spring-cloud-config/src/test/java/io/quarkus/ts/spring/cloud/config/SpringCloudConfigIT.java
Sgitario: This is used just to map the external file from the classpath to the container

jsmrcka: But how is it used by the container?
I mean, it obvoiusly works, I just don't see how.

Sgitario: The property is ignored by the container, but the framework will detect that there is a resource that needs to be mapped into the container.

jsmrcka: I see. I didn't realize it's the external resources feature of the framework. Thanks for the explanation! :)

Issue #186, opened 2021-08-17
Add coverage for `quarkus-spring-scheduled`


Sgitario:
Related to https://github.com/quarkus-qe/quarkus-test-suite/issues/157

PR #187, opened 2021-08-17
Add module `scheduling/spring` to cover `quarkus-spring-scheduled` ext

Sgitario:
We cover the following scenario from guide: https://quarkus.io/guides/spring-scheduled with some adjustments to use only Spring API.

Fix https://github.com/quarkus-qe/quarkus-test-suite/issues/186


Sgitario:
@jsmrcka would you mind taking a look into this PR?

jsmrcka (approved):

PR #192, opened 2021-08-18
Feature/reactive panache

fedinskiy:
Tests for hibernate-reactive in general and Panache in particular.
Test plans for these changes can be found here: https://github.com/quarkus-qe/quarkus-test-plans/blob/main/QUARKUS-1079.md

Progress:
- [x] join commits into one
- [x] description in README
- [x] decide, if it's worth to create a new module instead of changing the existing sql-db-app
- [x] Add some OpenShift scenarios
- [x] Coverage for remove method
- [x] openSession, withSession, transaction
- [x] Query method:  setMaxResults
- [x] validation of values without panache(Looks like it's not working now, will need to create a reproducer)
- [x] DTO projection
- [x] Attribute converters
- [x] Stateless sessions and their methods
- [x] Queries: named, native


fedinskiy:
TODO:
- [ ] join commits into one

README.md
```diff
@@ -229,6 +229,14 @@ Container images used in the tests are:
   - version 10.3: `registry.redhat.io/rhscl/mariadb-103-rhel7` (only if `ts.redhat.registry.enabled` is set)
 - MSSQL: `mcr.microsoft.com/mssql/rhel/server`
 
+### `sql-db/sql-reqactive`
```
Sgitario, line 232: Typo here

sql-db/sql-reactive/src/test/java/io/quarkus/ts/openshift/reactive/Postgresql12DatabaseIT.java
```diff
@@ -0,0 +1,36 @@
...
+
+import io.quarkus.test.bootstrap.DefaultService;
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.Container;
+import io.quarkus.test.services.QuarkusApplication;
+
+@QuarkusScenario
+//@EnabledIfSystemProperty(named = "ts.redhat.registry.enabled", matches = "true")
+public class Postgresql12DatabaseIT extends AbstractReactiveDatabaseIT {
```
Sgitario, line 11: You're using Postgres SQL 10, so you don't need the RH registry and this comment can be out, plus rename the class name to Postgresql10DatabaseIT.

sql-db/sql-reactive/src/test/java/io/quarkus/ts/openshift/reactive/MySQLDatabaseIT.java
```diff
@@ -0,0 +1,36 @@
+package io.quarkus.ts.openshift.reactive;
+
+import io.quarkus.test.bootstrap.DefaultService;
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.Container;
+import io.quarkus.test.services.QuarkusApplication;
+
+@QuarkusScenario
+//@EnabledIfSystemProperty(named = "ts.redhat.registry.enabled", matches = "true")
```
Sgitario, line 10: Remove comment

sql-db/sql-reactive/src/test/java/io/quarkus/ts/openshift/reactive/MySQLDatabaseIT.java
```diff
@@ -0,0 +1,36 @@
...
+public class MySQLDatabaseIT extends AbstractReactiveDatabaseIT {
+
+    private static final String MYSQL_USER = "quarkus_test";
+    private static final String MYSQL_PASSWORD = "quarkus_test";
+    private static final String MYSQL_DATABASE = "quarkus_test";
+    private static final int MYSQL_PORT = 3306;
+
+    @Container(image = "registry.access.redhat.com/rhscl/mysql-80-rhel7", port = MYSQL_PORT, expectedLog = "Only MySQL server logs after this point")
```
Sgitario, line 18: Use `{mysql.80.image}`

sql-db/sql-reactive/src/test/java/io/quarkus/ts/openshift/reactive/MsSQLDatabaseIT.java
```diff
@@ -0,0 +1,97 @@
...
+                    Assertions.assertEquals("Slovník", "title");
+                })
+                .onFailure(Assertions::fail);
+    }
+
+    @Test
+    @Override
+    @Disabled
+    public void testUniEndpoint() {
+        // MSSQL hibernate has some problems with extended Latin
```
Sgitario, line 88: This sounds odd to me. Can you describe a bit more the problems or add a link to the issue to get more information? Same as below.

sql-db/sql-reactive/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy-reactive</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.microsoft.sqlserver</groupId>
+            <artifactId>mssql-jdbc</artifactId>
+            <version>9.4.0.jre11</version>
+            <scope>test</scope>
```
Sgitario, line 50: Why do we need this dependency?

Sgitario (changes requested):
Besides the comments, why not adding some OpenShift scenarios too?
Adding @pjgg as reviewer since it did the Vert.x SQL module and has experience with reactive.

sql-db/sql-reactive/pom.xml
fedinskiy: For reproducer tests in MsSQLDatabaseIT. Probably, I'll drop both these tests and this dependency before the merge.

sql-db/sql-reactive/src/test/java/io/quarkus/ts/openshift/reactive/MsSQLDatabaseIT.java
fedinskiy: Yes, tests testUniEndpoint and testMultiEndpoint fail on retrieval of word "Slovník" from the database. The sixth letter(from extended Latin set) is replaced with some unreadable symbol. There is no issue AFAIK, since I found this behaviour while making this test.

sql-db/sql-reactive/src/test/java/io/quarkus/ts/openshift/reactive/MySQLDatabaseIT.java
fedinskiy: removed

sql-db/sql-reactive/src/test/java/io/quarkus/ts/openshift/reactive/MySQLDatabaseIT.java
fedinskiy: fixed

README.md
fedinskiy: fixed

sql-db/sql-reactive/src/test/java/io/quarkus/ts/openshift/reactive/Postgresql12DatabaseIT.java
fedinskiy: You're right, fixed

sql-db/sql-reactive/pom.xml
fedinskiy: Dependency removed

sql-db/sql-reactive/src/test/java/io/quarkus/ts/openshift/reactive/MsSQLDatabaseIT.java
fedinskiy: Added issue link to the code.

README.md
```diff
@@ -239,6 +239,14 @@ Container images used in the tests are:
   - version 10.3: `registry.redhat.io/rhscl/mariadb-103-rhel7` (only if `ts.redhat.registry.enabled` is set)
 - MSSQL: `mcr.microsoft.com/mssql/rhel/server`
 
+### `sql-db/sql-reactive`
```
pjgg, line 242: Could be renamed to `sql-db/sql-reactive-panache` because there are several way to work with SQL in reactive way

pom.xml
```diff
@@ -449,6 +449,7 @@
                 <module>sql-db/multiple-pus</module>
                 <module>sql-db/panache-flyway</module>
                 <module>sql-db/vertx-sql</module>
+                <module>sql-db/sql-reactive</module>
```
pjgg, line 452: I think that could be renamed to something like `sql-db/sql-reactive-panache`

sql-db/sql-reactive/src/test/java/io/quarkus/ts/openshift/OpenShiftDB2IT.java
```diff
@@ -0,0 +1,8 @@
+package io.quarkus.ts.openshift;
```
pjgg, line 1: I am not sure If Openshift test should have its own Packages, or share package with Reactive test.

Sgitario: Well spotted! Yes, please rename the packages to rule out the openshift name.

fedinskiy: What's wrong with openshift name in package name? Personally, I would prefer to keep OCP and ordinary tests separately, to not make packages bigger than 5-7 classes.

pom.xml
fedinskiy: Module renamed

README.md
fedinskiy: Good point, I renamed the module.

sql-db/sql-reactive/src/test/java/io/quarkus/ts/openshift/OpenShiftDB2IT.java
Sgitario: We are running the module on baremetal, with some optional OCP tests. That's why that having all the classes under `*.openshift.*` does not state the reality about what the module is being executed.
If you want to move the OpenShift tests into a `.openshift` package, fine with me.

fedinskiy: @Sgitario I am not sure if I understand you correctly. Tests which run on OCP are in package "io.quarkus.ts.openshift". Tests, which do not,  are in package "io.quarkus.ts.reactive". I would like to keep these tests in separate packages. Is this ok with you? If it's not, which package names would you prefer?

Sgitario: It should be "io.quarkus.ts.reactive.openshift" or simply in "io.quarkus.ts.reactive".

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Book.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public static Multi<Book> all() {
+        return streamAll();
+    }
+
+    public static Uni<Book> byId(Integer id) {
+        return findById(id);
+    }
+
+    public static Uni<Book> create(Integer author, String name) {
```
pjgg, line 48: Why create is a static method? 

could be part of a Book instance and by this way be able to persist a book.
For example, 

```
 public Uni<Book> create() {
        return this.persistAndFlush();
    }
```

Now imagine that you have a Rest method that persists a book

```
@PUT
    @Path("books/{author}/{name}")
    public Uni<Response> createBook(Integer author, String name) {
 Book book = new Book();
        book.setAuthor(author);
        book.setTitle(name);
         book.create()
```

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Library.java
```diff
@@ -0,0 +1,113 @@
...
+                .map(nothing -> Response.status(Response.Status.CREATED))
+                .onFailure().recoverWithItem(error -> Response.status(Response.Status.BAD_REQUEST).entity(error.getMessage()))
+                .map(Response.ResponseBuilder::build);
+    }
+
+    @GET
+    @Path("books/author/{name}")
+    public Multi<String> search(String name) {
+        return authors.byName(name)
+                .flatMap(author -> Multi.createFrom().iterable(author.getBooks()))
```
pjgg, line 58: I prefer to use `.onItem.transform(...` instead of Maps or flapmaps
Have a look at methods as `transformToMulti`, `transformToIterable`...

For example in this case we are returning a list of titles, so why not do something like: 

```
static List<String> toBookList(Author author) {
        return author.getBooks().stream().map(b -> b.getTitle()).collect(Collectors.toList());
    }
    
     @GET
    @Path("books/author/{name}")
    public Multi<String> search(String name) {
        return authors.byName(name)
                .onItem()
                .transformToIterable(Library::toBookList);
}
```

And then ...I think that this helper method that I have created (`toBookList`) should be part of Author or book domain, not part of the endpoint (from a conceptual point of view). Is like...if you have a method `toJson` or `toString` is part of the object that is going to be serialized.

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Library.java
```diff
@@ -0,0 +1,113 @@
...
+                .map(Book::getTitle);
+    }
+
+    @GET
+    @Path("author")
+    public Multi<String> author() {
+        return authors.streamAll()
+                .map(author -> author.getId() + " " + author.getName());
```
pjgg, line 66: I prefer `onItem.tranform` Also, have you think into have a toJson Method in your domain
something like:

```

public String toJsonStringify() {
        return Json.encode(this);
    }
```

And then later you could use it in your endpoints:

```
@GET
    @Path("author")
    public Multi<Response> author() {
        return authors.streamAll()
        .onItem().transform(author -> Response.ok(author.toJsonStringify()).build());
```
something like this...

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/AbstractReactiveDatabaseIT.java
```diff
@@ -0,0 +1,175 @@
...
+        Response response = given()
+                .when().get("/library/books/1");
+        String title = response.body().asString();
+        Assertions.assertEquals(HttpStatus.SC_OK, response.statusCode());
+        Assertions.assertEquals("Slovník", title);
+    }
+
+    @Test
+    public void testFind() {
+        String title = given()
```
pjgg, line 34:
```
given()
                .when().get("/library/books/2")
                .then()
                .statusCode(HttpStatus.SC_OK)
                .extract().body().asString();
```

This code is repeated coulbe refactor and reused in a private method

pjgg (changes requested):

sql-db/sql-reactive/src/test/java/io/quarkus/ts/openshift/OpenShiftDB2IT.java
fedinskiy: Changed.

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Book.java
fedinskiy: >Why create is a static method
By definition, there is no object to have instance methods before we create one.

In your example, we do not need any "create" method at all, we can just call persistAndFlush directly.

And with static method, your example become much more compact `Book.create(author, name)`

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Library.java
fedinskiy: Well, this method is needed only for smoke testing and for debugging purposes, so I see no reason to put much effort into it. Especially, since resteasy is not yet able to produce valid json from Multi anyway[1]

[1] https://github.com/quarkusio/quarkus/issues/18043

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/AbstractReactiveDatabaseIT.java
fedinskiy: It will, but I don't see any gains in it.
On the over side, if we will make our tests as simple as possible and with plain control flow, they will be easier to understand and to find a source of failure when something will go wrong.

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Library.java
fedinskiy: Well, onItem() and friends are more verbose and aren't used anywhere outside Mutiny.
Map and flatmap, on the other side, are much more compact and very common both in JVM world[1],[2] and outside of it [3], [4].

Good idea about putting this into helper method, I changed that.

[1] https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html
[2] https://spark.apache.org/docs/latest/api/java/index.html?org/apache/spark/sql/Dataset.html
[3] https://doc.rust-lang.org/std/iter/trait.Iterator.html
[4] https://hexdocs.pm/elixir/Enum.html

sql-db/hibernate-reactive/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkus.ts.qe</groupId>
+        <artifactId>parent</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+        <relativePath>../..</relativePath>
+    </parent>
+    <artifactId>hibernate-reactive</artifactId>
+    <packaging>jar</packaging>
+    <name>Quarkus QE TS: SQL Database: Reactive</name>
```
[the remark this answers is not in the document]
fedinskiy: Yes, will change the name

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Library.java
pjgg: could you add this issue as a // TODO comment in our method

```
@GET
    @Path("author")
// TODO quarkusio/quarkus#18043
    public Multi<Response> author() {
        return authors.streamAll()
        .onItem().transform(author -> Response.ok(author.toJsonStringify()).build());

```

fedinskiy: Added the description

rsvoboda:
@fedinskiy how long do the newly added tests run in JVM and Native mode ?

fedinskiy:
@rsvoboda 
On my machine:
34:26 min with DB2
02:12 min without

kshpak:
Review in process

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/GroundedEndpoint.java
```diff
@@ -0,0 +1,94 @@
...
+
+import org.hibernate.reactive.mutiny.Mutiny;
+
+import io.quarkus.arc.Arc;
+import io.smallrye.mutiny.Uni;
+
+@Path("/hibernate")
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public class GroundedEndpoint {
```
kshpak, line 20: Just a suggestion to stick with Resource for class name.
What is the purpose of this class?
Does it lack Multi(Response) methods?

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/GroundedEndpoint.java
```diff
@@ -0,0 +1,94 @@
...
+
+    Mutiny.SessionFactory factory;
+
+    public GroundedEndpoint() {
+        factory = Arc.container().instance(Mutiny.SessionFactory.class).get(); //TODO for some reason, @Inject fails. Need to investigate
+    }
+
+    @GET
+    @Path("books/{id}")
+    public Uni<Response> openSession(Integer id) {
```
kshpak, line 30: change the method name to reflect the method purpose

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/GroundedEndpoint.java
```diff
@@ -0,0 +1,94 @@
...
+                        .map(book -> book == null
+                                ? Response.status(Response.Status.NOT_FOUND)
+                                : Response.ok(book.getTitle()))
+                        .map(Response.ResponseBuilder::build)
+                        .eventually(session::close));
+    }
+
+    @GET
+    @Path("books/author/{authorId}")
+    public Uni<Response> findAll(Integer authorId) {
```
kshpak, line 42: why findAll? change the method name to reflect the method purpose

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/GroundedEndpoint.java
```diff
@@ -0,0 +1,94 @@
...
+            Author author = new Author();
+            author.setName(name);
+            return session.insert(author);
+        })
+                .map(ignored -> Response.status(Response.Status.CREATED))
+                .map(Response.ResponseBuilder::build);
+    }
+
+    @PUT
+    @Path("books/{authorName}/{name}")
```
kshpak, line 75: this is not the best endpoint specification IMO. replace it with URL parameters if possible

kshpak: why @PUT if it's a create method? Should be @POST

sql-db/hibernate-reactive/src/main/resources/application.properties
```diff
@@ -0,0 +1,5 @@
+quarkus.hibernate-orm.database.charset=utf-8
+quarkus.hibernate-orm.database.generation=drop-and-create
+
+quarkus.hibernate-orm.provider=org.hibernate.reactive.provider.ReactivePersistenceProvider
+quarkus.hibernate-orm.class=io.quarkus.ts.reactive.Book, io.quarkus.ts.reactive.Author
```
kshpak, line 5: what are these properties doing? can you please attach here the link where they described?  
If it's for purpose, then please create an issue that following props are missing in https://quarkus.io/guides/all-config

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/BookDescription.java
```diff
@@ -0,0 +1,13 @@
...
+import io.quarkus.runtime.annotations.RegisterForReflection;
+
+@RegisterForReflection
+public class BookDescription {
+    public final String name;
+
+    public BookDescription(@ProjectedFieldName("title") String title) {
+        this.name = title;
+    }
+}
```
kshpak, line 13: please state -> why this class is needed?

sql-db/hibernate-reactive/src/test/resources/test.properties
```diff
@@ -0,0 +1,2 @@
+ts.db2.container.privileged-mode=true
+ts.db2.startup.timeout=10m
```
kshpak, line 2: why these props are needed?

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/AuthorRepository.java
```diff
@@ -0,0 +1,25 @@
...
+        return findById(id);
+    }
+
+    public Uni<Author> create(String name) {
+        Author author = new Author();
+        author.setName(name);
+        return persistAndFlush(author);
+    }
+
+    public Multi<Author> byName(String name) {
```
kshpak, line 22:
```suggestion
    public Multi<Author> findByName(String name) {
```

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/GroundedEndpoint.java
```diff
@@ -0,0 +1,94 @@
...
+            return session.createNativeQuery("Select * from books where author='" + authorId + "'", Book.class)
+                    .setMaxResults(1)
+                    .getResultList()
+                    .map(books -> Response.ok(books).build());
+        });
+    }
+
+    @GET
+    @Path("books/starts_with/{prefix}")
+    public Uni<Response> findNamed(String prefix) {
```
kshpak, line 53: change the method name to reflect the method purpose.

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/BookDescription.java
kshpak: GENERAL: Please use field name `title` for book, not `name`

kshpak (changes requested):
This is not a complete review, but there are already enough things to change. Please respond to the questions and update the PR. I will take a look once again after the PR is updated

sql-db/hibernate-reactive/src/test/resources/db2_import.sql
```diff
@@ -0,0 +1,11 @@
+DROP TABLE IF EXISTS authors;
```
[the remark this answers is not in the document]
fedinskiy: The idea was tried, but it looks like DROP is not handled and create is no redundant.

sql-db/hibernate-reactive/src/test/resources/test.properties
fedinskiy: Db2 requires, that container is started in privileged mode and it has a long start-up phase, so we use a long possible timeout.

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Book.java
```diff
@@ -0,0 +1,83 @@
...
+
+import io.quarkus.hibernate.reactive.panache.PanacheEntityBase;
+import io.smallrye.mutiny.Multi;
+import io.smallrye.mutiny.Uni;
+
+@Entity
+@Table(name = "books")
+@NamedQuery(name = "title_prefix", query = "from Book where title like :prefix")
+public class Book extends PanacheEntityBase {
+    private static final int REASONABLY_LONG_TEXT = 100;
```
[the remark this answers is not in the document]
fedinskiy: The name changed

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/BookDescription.java
fedinskiy: This class is needed to test a DTO projection feature[1]. The name changed.

[1] https://quarkus.io/guides/hibernate-reactive-panache#query-projection

sql-db/hibernate-reactive/src/main/resources/application.properties
fedinskiy: Looks like this property is not needed for Panache. Removed it.

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Book.java
```diff
@@ -0,0 +1,83 @@
...
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    @Column(name = "id", unique = true, nullable = false)
+    private Integer id;
+
+    @NotNull
+    @Size(max = REASONABLY_LONG_TEXT)
+    private String title;
+
+    @NotNull
+    private Integer author;
```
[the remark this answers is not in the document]
fedinskiy: It's an id from another database

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Book.java
```diff
@@ -0,0 +1,83 @@
...
+        this.title = title;
+    }
+
+    public static Multi<Book> all() {
+        return streamAll();
+    }
+
+    public static Uni<Book> byId(Integer id) {
+        return findById(id);
+    }
```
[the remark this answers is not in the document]
fedinskiy: Yes, but due to the way generics implemented in Java it should be Book.<Book>findById, which is not very convenient.

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Book.java
```diff
@@ -0,0 +1,83 @@
...
+    private long isbn;
+
+    public Book() {
+    }
+
+    public Book(String title) {
+        this.title = title;
+    }
+
+    public static Multi<Book> all() {
```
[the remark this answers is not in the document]
fedinskiy: Same thing, as with findById method.

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/ISBNConverter.java
```diff
@@ -0,0 +1,32 @@
...
+            throw new IllegalStateException("ISBN " + s + " has a wrong length: " + number);
+        }
+        int prefixLength = 13 - s.length();
+        result.insert(0, "0".repeat(prefixLength));
+        result.insert(3, '-');
+        result.insert(5, '-');
+        result.insert(8, '-');
+        result.insert(15, '-');
+        return result.toString();
+    }
```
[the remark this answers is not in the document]
fedinskiy: 1) This conversion works before we write into database, and before checks are made.
2, 3) changed(although ISBN standard calls this thing "prefix")
4) Suggestion declined, but added short description of the formatting

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/GroundedEndpoint.java
fedinskiy: It doesn't use Panache methods, hence the pun in the name(ideas for better name are welcomed, though). I am not sure, that Resource can be a good name for an endpoint class.

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/GroundedEndpoint.java
fedinskiy: It was meant to solve a strange behaviour of PUT endpoints in Quarkus. Changed.

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/GroundedEndpoint.java
fedinskiy: fixed

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Library.java
```diff
@@ -0,0 +1,134 @@
...
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+import io.smallrye.mutiny.Multi;
+import io.smallrye.mutiny.Uni;
+
+@Path("/library")
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public class Library {
```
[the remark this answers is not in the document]
fedinskiy: Name changed

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Book.java
```diff
@@ -0,0 +1,83 @@
...
+import javax.validation.constraints.Size;
+
+import io.quarkus.hibernate.reactive.panache.PanacheEntityBase;
+import io.smallrye.mutiny.Multi;
+import io.smallrye.mutiny.Uni;
+
+@Entity
+@Table(name = "books")
+@NamedQuery(name = "title_prefix", query = "from Book where title like :prefix")
+public class Book extends PanacheEntityBase {
```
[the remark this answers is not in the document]
kshpak: what about this one?

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/ConverterTest.java
```diff
@@ -0,0 +1,36 @@
+package io.quarkus.ts.reactive;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class ConverterTest {
```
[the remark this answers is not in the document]
kshpak: We don't need it IMO. @pjgg what do you think?

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/ISBNConverter.java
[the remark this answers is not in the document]
kshpak: ok, this is not that crucial part

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/GroundedEndpoint.java
kshpak: findAllBooksByAuthor() -> findBookByAuthor
This method returns only one item

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Book.java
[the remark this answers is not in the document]
fedinskiy: As I understand it, Quarkus generates getters and setters, which can be used by Hibernate, but cannot be used by client code. Since we need to use these getters and setters in client code, we have to define them manually.

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/PanacheEndpoint.java
```diff
@@ -0,0 +1,135 @@
...
+@Path("/library")
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public class PanacheEndpoint {
+
+    @Inject
+    AuthorRepository authors;
+
+    @GET
+    @Path("books")
```
kshpak, line 26:
```suggestion
    @Path("/books")
```
applies to all endpoints in this class

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Book.java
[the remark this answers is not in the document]
kshpak: you can map it by OneToMany <-> ManyToOne, this shouldn't be a problem

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/PanacheEndpoint.java
```diff
@@ -0,0 +1,135 @@
...
+    public Multi<String> all() {
+        return Book.all()
+                .map(Book::getTitle);
+    }
+
+    @GET
+    @Path("books/{id}")
+    public Uni<Response> find(Integer id) {
```
kshpak, line 34:
```suggestion
    public Uni<Response> findById(Integer id) {
```

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/PanacheEndpoint.java
```diff
@@ -0,0 +1,135 @@
...
+    public Multi<String> search(String name) {
+        return authors.findByName(name)
+                .flatMap(Author::getBooks)
+                .map(Book::getTitle);
+    }
+
+    @GET
+    @Path("author")
+    // TODO: Return Json, when this will be fixed: https://github.com/quarkusio/quarkus/issues/18043
+    public Multi<String> author() {
```
kshpak, line 86:
```suggestion
    public Multi<String> getAllAuthors() {
```
please go through method names and adjust their names

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/PanacheEndpoint.java
```diff
@@ -0,0 +1,135 @@
...
+                .map(isDeleted -> isDeleted
+                        ? Response.status(Response.Status.NO_CONTENT)
+                        : Response.status(Response.Status.NOT_FOUND))
+                .map(Response.ResponseBuilder::build);
+    }
+
+    @GET
+    @Path("dto")
+    public Uni<Response> dtoQuery() {
+        return Book.find("author", 4).project(BookDescription.class).list()
```
kshpak, line 126: We need to avoid the use of "magic numbers" in the code. Find more appropriate filter parameter

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/IdGenerator.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.hibernate.reactive.id.ReactiveIdentifierGenerator;
+import org.hibernate.reactive.session.ReactiveConnectionSupplier;
+
+import io.quarkus.runtime.annotations.RegisterForReflection;
+
+@RegisterForReflection
+public class IdGenerator implements ReactiveIdentifierGenerator<Integer> {
+    private final AtomicInteger lastId = new AtomicInteger(4);
```
kshpak, line 14: We need to avoid the use of "magic numbers" in the code. Specify this number as constant for example. Probably you start the id generation from 4 because of /import.sql, this doesn't seem the great way to do it.

kshpak (changes requested):
changes required

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/IdGenerator.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.hibernate.reactive.id.ReactiveIdentifierGenerator;
+import org.hibernate.reactive.session.ReactiveConnectionSupplier;
+
+import io.quarkus.runtime.annotations.RegisterForReflection;
+
+@RegisterForReflection
+public class IdGenerator implements ReactiveIdentifierGenerator<Integer> {
```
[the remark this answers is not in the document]
fedinskiy: This class is required to verify that `@GenericGenerator` annotation works.

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Book.java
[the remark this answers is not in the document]
kshpak: You can  retrieve the book before the **return** statement
Uni<Book> book = Book.findById(id);
return book.flatMap...

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Book.java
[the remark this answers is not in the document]
kshpak: the same as findById()
Multi<Book> books = Book.streamAll();

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/PanacheEndpoint.java
fedinskiy: But why? This path is a relative one, so a slash in the beginning would be misleading.

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Book.java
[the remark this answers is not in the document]
kshpak: Agree

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/PanacheEndpoint.java
fedinskiy: Again, why? We never call this method directly, and usage of id is easily seen from the method signature and URl template

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/IdGenerator.java
fedinskiy: Magic number removed, thank you!

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/PanacheEndpoint.java
fedinskiy: Added id to url

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/PanacheEndpoint.java
fedinskiy: Method renamed

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/PanacheEndpoint.java
kshpak: It is not misleading at all, look how other endpoints define it in other modules

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Book.java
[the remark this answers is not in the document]
fedinskiy: Well, it's already mapped on the other side and I don't see any problem with this field at the moment. Do you?

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Book.java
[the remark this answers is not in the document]
fedinskiy: I can. Also I can just write a single method and use fluent API. Is there some problem with that?

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/GroundedEndpoint.java
fedinskiy: Name changed.

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/PanacheEndpoint.java
fedinskiy: Well, I can fix this in other modules too, if such uniformity is required :D

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/PanacheEndpoint.java
kshpak: It's just for readability purposes, you are searching "by" some criteria, you are not trying to find integer

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/PanacheEndpoint.java
kshpak: :D Ok, both styles can be used. Just a matter of preference, resolving this and similar threads

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Book.java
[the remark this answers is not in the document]
kshpak: This is exactly the case why we need to define CREATE Table with foreign key for every import.sql file. Let's see what @pjgg will say, personally I'm against unidirectional association at this case.

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/Book.java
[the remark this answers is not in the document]
kshpak: I don't see a problem in a generics syntax anyway. And again readability, "byId" could be: update, create, delete, find. @pjgg your thoughts? I still think this method is redundant

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/MariaDbDatabaseIT.java
```diff
@@ -0,0 +1,36 @@
...
+@QuarkusScenario
+public class MariaDbDatabaseIT extends AbstractReactiveDatabaseIT {
+
+    private static final String MYSQL_USER = "quarkus_test";
+    private static final String MYSQL_PASSWORD = "quarkus_test";
+    private static final String MYSQL_DATABASE = "quarkus_test";
+    private static final int MYSQL_PORT = 3306;
+
+    @Container(image = "${mariadb.102.image}", port = MYSQL_PORT, expectedLog = "Only MySQL server logs after this point")
+    static DefaultService database = new DefaultService()
```
pjgg, line 18: Please instead of `DefaultService` use `MariaDbService`

pjgg (changes requested):

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/MariaDbDatabaseIT.java
fedinskiy: I have doubts about that, since Quarkus Hibernate doesn't yet support maria db(see comment on line 28). We will need to override parameters anyway, so I suggest we will do it later, when we have a proper support. WDYT?

pjgg: Then add a //TODO

kshpak: "Quarkus Hibernate doesn't yet support maria db" what is not supported specifically? + to add a mention to include OpenShiftMariaDb scenario once resolved with TODO comment

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/ConverterTest.java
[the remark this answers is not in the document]
kshpak: Agree to have a converter test that tests integration between hibernate<->converter on a different PR.

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/MariaDbDatabaseIT.java
fedinskiy: TODO added

fedinskiy:
run tests

pjgg:
@fedinskiy 
`DB2DatabaseIT.java' has not been previously formatted.  Please format file and commit before running validation! `

rsvoboda:
run tests

pjgg:
run tests

fedinskiy:
@pjgg OpenShift tests are now run as a part of standard checks. If you want, you can rerun them there: https://github.com/quarkus-qe/quarkus-test-suite/runs/4334102321?check_suite_focus=true

pjgg (approved):
LGTM, Thank you for your contribution!

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/AbstractReactiveDatabaseIT.java
```diff
@@ -0,0 +1,243 @@
...
+        String result = given()
+                .when().get("library/books/author/Vern")
+                .then()
+                .statusCode(HttpStatus.SC_OK)
+                .extract().body().asString();
+        Assertions.assertEquals("[Around_the_World_in_Eighty_Days]", result);
+    }
+
+    @Test
+    public void deletion() {
```
kshpak, line 171: Please take a look at test names. Try to make uniform naming. The same applies for test: dto, get*, set*, validationWithoutPanache

kshpak:
run tests

kshpak (changes requested):
Please resolve the last two comments && it's ready to be merged

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/MsSQLDatabaseIT.java
```diff
@@ -0,0 +1,44 @@
...
+            .withProperty("quarkus.datasource.reactive.url", database::getReactiveUrl);
+
+    @Override
+    protected RestService getApp() {
+        return app;
+    }
+
+    @Test
+    @Override
+    @Disabled("https://github.com/quarkusio/quarkus/issues/19539")
```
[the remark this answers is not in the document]
fedinskiy: Added TODOs to all applicable classes in the module

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/AbstractReactiveDatabaseIT.java
fedinskiy: Changed test names to the single style.

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/Postgresql12DatabaseIT.java
```diff
@@ -0,0 +1,36 @@
...
+
+import io.quarkus.test.bootstrap.PostgresqlService;
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.Container;
+import io.quarkus.test.services.QuarkusApplication;
+
+@QuarkusScenario
+@EnabledIfSystemProperty(named = "ts.redhat.registry.enabled", matches = "true")
+public class Postgresql12DatabaseIT extends AbstractReactiveDatabaseIT {
```
kshpak, line 13: This test will never be run in the bare metal cause of this property + there is no respective OpenshiftTest for Postgresql12.
Please add the OpenShift test for it

kshpak (changes requested):
Missed this one before

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/AbstractReactiveDatabaseIT.java
```diff
@@ -0,0 +1,243 @@
...
+import io.quarkus.test.bootstrap.RestService;
+import io.restassured.http.ContentType;
+import io.restassured.path.json.JsonPath;
+import io.restassured.response.Response;
+
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
+public abstract class AbstractReactiveDatabaseIT {
+    @Test
+    public void getAll() {
+        getApp().given()
```
kshpak, line 21: Validate returned books + remove getApp() reference in the rest of the tests

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/AbstractReactiveDatabaseIT.java
```diff
@@ -0,0 +1,243 @@
...
+    @Test
+    public void namedQuery() {
+        Response response = given()
+                .when().get("/hibernate/books/starts_with/Thinking");
+        String result = response.getBody().asString();
+        Assertions.assertTrue(result.contains("Thinking fast and slow"));
+    }
+
+    @Test
+    @Order(3)
```
kshpak, line 102: Put orders on top of the class + sort them

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/AbstractReactiveDatabaseIT.java
```diff
@@ -0,0 +1,243 @@
...
+        String result = given()
+                .when().get("/library/author/7")
+                .then()
+                .statusCode(HttpStatus.SC_OK)
+                .extract().body().asString();
+        Assertions.assertEquals("Plato", result);
+    }
+
+    @Test
+    public void getAuthor() {
```
kshpak, line 117:
```suggestion
    public void getAuthorById() {
```

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/AbstractReactiveDatabaseIT.java
```diff
@@ -0,0 +1,243 @@
...
+        String result = given()
+                .when().get("library/books/author/Vern")
+                .then()
+                .statusCode(HttpStatus.SC_OK)
+                .extract().body().asString();
+        Assertions.assertEquals("[Around_the_World_in_Eighty_Days]", result);
+    }
+
+    @Test
+    public void deleteAuthor() {
```
kshpak, line 171:
```suggestion
    public void deleteAuthorById() {
```

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/AbstractReactiveDatabaseIT.java
```diff
@@ -0,0 +1,243 @@
...
+        given().delete("library/author/1")
+                .then()
+                .statusCode(HttpStatus.SC_NO_CONTENT);
+        Response response = given()
+                .when().get("/library/author/1");
+        Assertions.assertEquals(HttpStatus.SC_NOT_FOUND, response.statusCode());
+    }
+
+    @Test
+    public void useDTO() {
```
kshpak, line 181: Expand DTO meaning pls

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/AbstractReactiveDatabaseIT.java
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    @Test
+    public void setConvertedZeroValue() {
+        Response change = getApp().given().put("/library/isbn/1/5170261586");
+        Assertions.assertEquals(HttpStatus.SC_OK, change.statusCode());
+        Response lookup = getApp().given().get("/library/isbn/1");
+        Assertions.assertEquals(HttpStatus.SC_OK, lookup.statusCode());
+        Assertions.assertEquals("5170261586", lookup.body().asString());
+    }
```
kshpak, line 232: Add test to check the raw value in the database (basically after conversion). ++ Drop ConverterTest and reflect it's logic in these IT tests

kshpak (changes requested):
Discussed changes

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/AbstractReactiveDatabaseIT.java
fedinskiy: expanded

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/AbstractReactiveDatabaseIT.java
fedinskiy: Implemented

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/AbstractReactiveDatabaseIT.java
fedinskiy: done

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/AbstractReactiveDatabaseIT.java
fedinskiy: Validation added, getApp added to all methods of the class

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/Postgresql12DatabaseIT.java
fedinskiy: Postgresql test coverage changed.

pjgg:
Kyrylo comments were already reflected.

kshpak:
> Kyrylo comments were already reflected.

Thanks for the final merge :)

Issue #77, opened 2021-07-21
Migrate 004-quarkus-HHH-and-HV from the Quarkus Beefy Test Suite


Sgitario:

rsvoboda:
004 uses H2, migrated app could use dev service for bare metal use cases.
PostgreSQL looks like good candidate on OpenShift.

PR #201, opened 2021-08-30
[#77] Quarkus Hibernate and Hibernate Validator

mjurc:
* Migrating 004-quarkus-HHH-and-HV from beefy-scenarios


pjgg (changes requested):
I saw that the test is using `@QuarkusTest` instead of `@QuarkusScenario` . If the main problem to do that is `H2` (that I think is not covered by @QuarkusScenario), did you think to move on to PostgreSQL instead?

pjgg (approved):
There is an issue on CI but is not related to these changes

PR #207, opened 2021-09-01
CLI: Fix using stream test by getting stream ID from version

Sgitario:
The stream ID 2.0 is gone from the platform site and this might eventually happen in the future for other versions.
So, in order to fix this test and prevent future failures, we need to infer the stream ID from the current Quarkus version.


Sgitario:
@rsvoboda can you take a look into this PR? This issue is causing that the Daily action and other PRs fail.

rsvoboda (approved):

PR #202, opened 2021-08-30
Reactive DevServices postgresql

pjgg:
Reactive Postgresql / Dev services scenario

PR #208, opened 2021-09-01
Add reactive/mysql DevService Scenario

pjgg:
Verifies Reactive-mysql / DevServices integration


Sgitario (changes requested):
This module is very similar to https://github.com/quarkus-qe/quarkus-test-suite/pull/202/files. I think both should be merged in one.

Sgitario (commented):
I would prefer to have one endpoint /book/postgresql and another endpoint /book/mysql (instead of notebook), so it's clearer that the purpose of each endpoint is to use one datasource or another.

Sgitario (changes requested):
The BookResource and NotebookResource is almost identical, pls refactor the solution so we get rid of the duplicity.

Sgitario (approved):

Sgitario (changes requested):
CI is failing

Sgitario:
Blocked by https://github.com/quarkus-qe/quarkus-test-suite/pull/236.

Sgitario (approved):

PR #209, opened 2021-09-01
Add HTTP Caching scenario to jaxrs-reactive

jsmrcka:


http/jaxrs-reactive/src/main/resources/application.properties
```diff
@@ -0,0 +1,5 @@
+quarkus.openshift.route.expose=true
```
jsmrcka, line 1: Honestly, I don't really know why this property is needed, but the scenario fails without it.

Sgitario: I think it makes sense as otherwise the route is not exposed and hence you can't verify your endpoints.

Sgitario (approved):
simple and beautiful!

http/jaxrs-reactive/src/main/resources/application.properties
jsmrcka: That's the weird thing, the endpoints work fine without it, because the test framework itself contains this setting. But some of the tests fail because of https://access.redhat.com/solutions/4820731.

Sgitario:
run tests

PR #217, opened 2021-09-02
[2.2] CLI: Fix using stream test by getting stream ID from version

Sgitario:
The stream ID 2.0 is gone from the platform site and this might eventually happen in the future for other versions.
So, in order to fix this test and prevent future failures, we need to infer the stream ID from the current Quarkus version.

Backport of https://github.com/quarkus-qe/quarkus-test-suite/pull/207


Sgitario:
Will be done in https://github.com/quarkus-qe/quarkus-test-suite/pull/219

PR #227, opened 2021-09-06
Use Quarkus version 2.2.1.Final

Sgitario:
Due to this issue: https://github.com/quarkusio/quarkus/issues/19875, we can't use 999-SNAPSHOT until is fixed, therefore Main will be based on 2.2.1.Final.

++ Daily actions and Quarkus Ecosystem will still use 999-SNAPSHOT to check out the status with 999-SNAPSHOT
++ Disable HttpCachingResourceIT.shouldGetNoCacheUnqualified due to https://github.com/quarkusio/quarkus/issues/19822

### DIFF
--- a/.github/quarkus-ecosystem-test
+++ b/.github/quarkus-ecosystem-test
@@ -2,7 +2,7 @@
 set -e
 
 # run the tests on JVM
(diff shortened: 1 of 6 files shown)


rsvoboda (approved):
OK, waiting for CI before merge

PR #229, opened 2021-09-07
Add REST Client Reactive multipart scenario

jsmrcka:
Based on https://github.com/quarkusio/quarkus/issues/19892.
Serves as a verification for related JIRA:
https://issues.redhat.com/browse/QUARKUS-1225


Sgitario (approved):

Sgitario:
Good addition! Thanks for this PR. 
Will merge when CI is green. I don't think we need to backport it to 2.2 unless the issue is fixed and backported in upstream which is unlikely to happen.

PR #233, opened 2021-09-07
[2.2] Maintain Quarkus Ide Config independently

Sgitario:
As this dependency is not part of RHBQ, we need to use the upstream version always.
Dependabot will maintain it.

Back port of https://github.com/quarkus-qe/quarkus-test-suite/pull/232


rsvoboda (approved):

Sgitario:
The JVM failure is a sneaky test only in Windows, will investigate it separately.

PR #238, opened 2021-09-08
Adapt main to Quarkus 999-SNAPSHOT

Sgitario:
This change is caused by https://github.com/quarkusio/quarkus/issues/19875 which as far I understood, it's not going to be reverted or fixed.

Guides updates in https://github.com/quarkusio/quarkus/pull/19995


sql-db/panache-flyway/src/test/java/io/quarkus/qe/sqldb/panacheflyway/DataSourceIT.java
```diff
@@ -27,6 +24,6 @@ public void shouldDataSourceBeProperlyConfigured() {
     }
 
     private static final RequestSpecification dataSourcePath() {
-        return given().accept(MediaType.APPLICATION_JSON).contentType(ContentType.JSON).when().basePath(DATA_SOURCE_PATH);
+        return given().when().basePath(DATA_SOURCE_PATH);
```
Sgitario, line 27: This was wrong and using reactive fails which I think it's expected. It's now corrected.

jsmrcka:
`lifecycle-application` still uses 2.2.2.Final
https://github.com/Sgitario/quarkus-test-suite/blob/aab9b720f89ffcd8067503c9f0cbe7a90452382e/lifecycle-application/pom.xml#L42

spring/spring-data/pom.xml
```diff
@@ -11,10 +11,6 @@
     <packaging>jar</packaging>
     <name>Quarkus QE TS: Spring: Spring Data</name>
     <dependencies>
-        <dependency>
-            <groupId>io.quarkus</groupId>
```
fedinskiy, line 15: Is that dependency was giving some problems or it's just not used?

spring/spring-data/src/test/java/io/quarkus/ts/spring/data/primitivetypes/CommonsHeadersIT.java
```diff
@@ -63,6 +65,8 @@ public void testRequestScope() {
     //This is for regression test for https://github.com/quarkusio/quarkus/pull/12234
     @Test
     @DisplayName("session scope")
+    @DisabledOnQuarkusVersion(version = "2.3.*", reason = "quarkus-spring-data-rest is now reactive")
```
fedinskiy, line 68: Is it temporary fix? Because it looks like https://github.com/quarkusio/quarkus/issues/12233 is about spring DI and should be orthogonal to spring-data

Sgitario: As it now uses reactive approach, I would say it's a permanent fix. Wdyt @jsmrcka ?

spring/spring-data/pom.xml
Sgitario: Resteasy Reactive does not work well with Undertow (see https://github.com/quarkusio/quarkus/issues/14997). Spite there are workarounds, this is not necessary anylonger.

Sgitario:
> `lifecycle-application` still uses 2.2.2.Final
> https://github.com/Sgitario/quarkus-test-suite/blob/aab9b720f89ffcd8067503c9f0cbe7a90452382e/lifecycle-application/pom.xml#L42

Updated

spring/spring-data/src/test/java/io/quarkus/ts/spring/data/primitivetypes/CommonsHeadersIT.java
jsmrcka: TBH, I don't understand why this doesn't work with reactive approach.

And also, why is it disabled just for `2.3.*`? I suppose it's not going to work for any future version.

Sgitario: This is a good point... as I dropped the undertow dependency, this is not going to work ever, so we should delete this test case. 

It does not work because the reactive approach does not propagate the session id.

jsmrcka (approved):

fedinskiy (approved):

PR #243, opened 2021-09-09
Fix Hibernate module

Sgitario:
Changes:
- Rename the Dev Mode to follow the name convention in the test suite: https://github.com/Sgitario/quarkus-test-suite#name-convention

- Add coverage for JVM and Native:

This module was migrated from Beefy: https://github.com/quarkus-qe/beefy-scenarios/blob/main/004-quarkus-HHH-and-HV. 
The problem is that in Beefy we were covering JVM and Native both, and the migrated module was only running it on Dev Mode.

- Move all the tests in a single scenario to improve the execution time (don't need to create N deployments, just one)

- Rename the module description to not use the word "features":

The framework is watching the app logs for the word "features" and if found, it thinks the app has started. This is a bit weak and it's going to be fixed in https://github.com/quarkus-qe/quarkus-test-framework/pull/250. However, for the time being, I've changed the module description from "features" to "exploratory" to avoid having to put a workaround in the test code.


pjgg (approved):

mjurc (approved):
Thanks for the fix!

PR #253, opened 2021-09-09
REST resource extending abstract class fail on POST/PUT scenario

pjgg:
Reproducer for the issue: https://github.com/quarkusio/quarkus/issues/20008


security/keycloak-authz/pom.xml
```diff
@@ -13,7 +13,7 @@
     <dependencies>
         <dependency>
             <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-resteasy</artifactId>
+            <artifactId>quarkus-resteasy-reactive</artifactId>
```
Sgitario, line 16: My only concern here is that we're losing the coverage of RestEasy Classic. 
Not sure if we should eventually start splitting the coverage into classic and reactive.

Sgitario (commented):
The changes are ok for me.
But I've a concern (see my only comment) that I would like the oppinion from @mjurc and/or @rsvoboda

security/keycloak-authz/pom.xml
rsvoboda: This is concerning me too, let's discuss options on chat.

rsvoboda (changes requested):
The conclusion from chat discussion is to use duplicated approach, please update the PR to have security/keycloak-authz-classic ++ security/keycloak-authz-reactive

pom.xml
```diff
@@ -397,14 +397,15 @@
...
-                <module>security/keycloak-authz</module>
+                <module>security/keycloak-authz-classic</module>
                 <module>security/keycloak-jwt</module>
                 <module>security/keycloak-webapp</module>
                 <module>security/keycloak-oauth2</module>
                 <module>security/keycloak-multitenant</module>
                 <module>security/keycloak-oidc-client-basic</module>
                 <module>security/keycloak-oidc-client-extended</module>
                 <module>security/vertx-jwt</module>
+                <module>security/keycloak-authz-reactive-resteasy</module>
```
Sgitario, line 408: I would drop the word `resteasy`. Simply: reactive. And also group both modules together, pls.

security/keycloak-authz-reactive-resteasy/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkus.ts.qe</groupId>
+        <artifactId>parent</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+        <relativePath>../..</relativePath>
+    </parent>
+    <artifactId>security-keycloak-authz-reactive-resteasy</artifactId>
+    <packaging>jar</packaging>
+    <name>Quarkus QE TS: Security: Keycloak + Authorization + reactive resteasy</name>
```
Sgitario, line 12: Same here, I would drop Resteasy.
++ Also, update the classic description with + Classic.

Sgitario (changes requested):

Sgitario (approved):

Sgitario:
run tests

PR #256, opened 2021-09-10
[2.2] REST resource extending abstract class fail on POST/PUT scenario

pjgg:
backport of: #253


Sgitario:
The fix was not backported to 2.2, let's on hold this PR until is done.

Issue #259, opened 2021-09-13
Fix OpenShift Keycloak tests


Sgitario:
Affected tests:
- OpenShiftRhSso73WebappSecurityIT
- OpenShiftRhSso74WebappSecurityIT
- OpenShiftRhSso73MultiTenantSecurityIT
- OpenShiftRhSso74MultiTenantSecurityIT 
- OpenShiftRhSso73OidcJwtSecurityIT
- OpenShiftRhSso74OidcJwtSecurityIT

It fails with `org.opentest4j.AssertionFailedError: Login page title should display application realm ==> expected: <Sign in to test-realm> but was: <Log in to test-realm>`

PR #260, opened 2021-09-13
Fix OpenShift Keycloak + Sql Server scenarios

Sgitario:
The problem is that in RH SSO 7.3 is expected `Log in to test-realm` and in RH SSO 7.4 is `Sign in to test-realm`.

I fixed it by using a regular expression.

Fix https://github.com/quarkus-qe/quarkus-test-suite/issues/259


Sgitario:
run tests

Sgitario:
run tests

sql-db/sql-app/src/test/resources/test.properties
```diff
@@ -1,6 +1,3 @@
 ts.app.log.enable=true
 ts.database.log.enable=true
-ts.database.openshift.use-internal-service-as-url=true
-ts.mssql.log.enable=true
-ts.mssql.openshift.template=/mssql.yaml
```
jsmrcka, line 5: Is dropping this custom template fixing anything, or has it just become obsolete?

jsmrcka (approved):

jsmrcka:
JVM failures are caused by missing RH registry configuration. Not related to the PR.

sql-db/sql-app/src/test/resources/test.properties
Sgitario: Yes, it's fixing the OpenShift Mssql scenario. This scenario without this configuration works fine.

jsmrcka: I see. I didn't notice the scenario failing. Thanks for the explanation.

PR #261, opened 2021-09-13
Add OpenTelemetry OpenShift scenario

kshpak:
Test plan: https://github.com/quarkus-qe/quarkus-test-plans/blob/main/QUARKUS-280.md


Sgitario (changes requested):
Overall, the changes look good, but I'm missing the baremetal scenario.

kshpak:
PR updated

kshpak:
CI should be green now, disabled bare metal on windows

opentelemetry/pom.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkus.ts.qe</groupId>
+        <artifactId>parent</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+        <relativePath>..</relativePath>
```
pjgg, line 8: @kshpak @Sgitario 
WDYT if we move on all the monitoring modules (as this one) to a "monitoring" folder as security?

opentelemetry/src/test/java/io/quarkus/ts/opentelemetry/OpentelemetryIT.java
```diff
@@ -0,0 +1,84 @@
...
+import io.restassured.response.Response;
+
+@QuarkusScenario
+@DisabledOnOs(value = OS.WINDOWS, disabledReason = "Windows does not support Linux Containers / Testcontainers")
+public class OpentelemetryIT {
+    private static final int GRPC_COLLECTOR_PORT = 14250;
+
+    private Response resp;
+
+    @JaegerContainer(image = "jaegertracing/all-in-one:latest", restPort = GRPC_COLLECTOR_PORT, expectedLog = "\"Health Check state change\",\"status\":\"ready\"")
```
pjgg, line 32: `GRPC_COLLECTOR_PORT` -> is using the default port, so I think that is not really necessary. 

`jaegertracing/all-in-one:latest` -> I don't mind be pointing to the latest version, but I think that is more stable to be pointing to the latest tagged version or just use the default one that comes from `@JaegerContainer`. I mean there is any motivation to use the latest jaeger version? 

If finally, you need one specific version please don't hardcore the version in the test, instead write down [in the pom](https://github.com/quarkus-qe/quarkus-test-suite/blob/main/pom.xml#L196)

pjgg (changes requested):

pjgg:
Apart from the minor changes that I suggested looks great to me. I have tested locally (bare-metal and Openshift) and works as expected.

opentelemetry/pom.xml
Sgitario: +1

Sgitario:
run tests

opentelemetry/src/test/java/io/quarkus/ts/opentelemetry/OpentelemetryIT.java
kshpak: GRPC_COLLECTOR_PORT is needed because the default collector port is REST.
No reason for latest really, fallback on default

opentelemetry/pom.xml
kshpak: module moved under /security/monitoring/opentelemetry
profile micrometer was renamed into monitoring
@Sgitario wdyt?

Sgitario: Not under `security`, simply `monitoring`.
Pablo's point was that the same we have the folder `security` with the modules `keycloak`, `keycloak-webapp`, ... We should have a `monitoring` folder to group all the modules related to monitoring. At the moment, we already have a module `microprofile` that should be moved to the new `monitoring` folder.

kshpak: Changed. Separate commit for microprofile module migration done

Sgitario (commented):
Update the microprofile reference in the README.md.
Otherwise, it looks ok for me.

kshpak:
Readme adjusted.

Sgitario (approved):

kshpak:
PR updated one more time. Micrometer modules were moved under the monitoring folder (this is kinda expected because they were under monitoring profile)

pjgg (approved):
LGTM

Sgitario:
run tests

kshpak:
This PR is not related to OCP failures, can it be merged?

Sgitario:
> This PR is not related to OCP failures, can it be merged?

Please, investigate the OCP failures and spot what could cause the failures.

PR #270, opened 2021-09-14
Many Extensions: Disable health extension

Sgitario:
The OpenShift deployment needs the healthcheck to be UP, but it was:

```
{"status":"DOWN","checks":[{"name":"Database connections health check","status":"UP"},{"name":"Artemis JMS health check","status":"DOWN"}]}
```

I think the purpose of this module is to cover as many extensions as possible, so I decided to exclude all the checks via property.

++ Building the Quarkus Test Suite parent separately:
This is a consequence that we're building only the changed modules in the PR workflow.
Also, the test framework does not deal well when the parent is not in the parent folder of the current module.


Sgitario:
run tests

rsvoboda (approved):

PR #273, opened 2021-09-14
Feature/websocket

fedinskiy:
Coverage of sending messages over WebSocket in our new TS.
Commits will be joined before the merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -715,3 +715,9 @@ a collection using a filter and a projection. All tests are performed using:
 ### `nosql-db/mongodb-reactive`
 
 Reactive equivalent of `nosql-db/mongodb`. Uses reactive ReactiveMongoClient (without codecs)
+
+### `websockets/quarkus-websockets`
+Coverage for sending messages over websockets
+
+### `websockets/websockets-client`
+Coverage for sending messages over websockets with only a client library
--- a/pom.xml
+++ b/pom.xml
@@ -491,6 +491,19 @@
             </modules>
         </profile>
         <profile>
+            <id>websockets</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+                <property>
+                    <name>all-modules</name>
+                </property>
+            </activation>
+            <modules>
+                <module>websockets/quarkus-websockets</module>
+                <module>websockets/websockets-client</module>
+            </modules>
+        </profile>
+        <profile>
             <id>native</id>
             <activation>
                 <property>
--- a/websockets/quarkus-websockets/pom.xml
+++ b/websockets/quarkus-websockets/pom.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkus.ts.qe</groupId>
+        <artifactId>parent</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+        <relativePath>../..</relativePath>
+    </parent>
+    <artifactId>quarkus-websockets</artifactId>
+    <name>Quarkus QE TS: Messaging: Websockets producer</name>
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-websockets</artifactId>
+        </dependency>
+    </dependencies>
+</project>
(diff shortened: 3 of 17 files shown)


messaging/websockets/src/test/java/io/quarkus/ts/ChatTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.QuarkusApplication;
+
+@QuarkusScenario
+public class ChatTest {
```
Sgitario, line 24: It should be renamed to ChatIT

Sgitario (changes requested):
One rename and some missing scenarios:
- Cover the usage of websockets-client dependency in isolation: This is to have two Quarkus applications, one using websockets server and another one websockets client (this is only one example)
- Add OpenShift scenarios.

messaging/websockets/src/test/java/io/quarkus/ts/ChatTest.java
fedinskiy: Class renamed

messaging/websockets-client/pom.xml
```diff
@@ -0,0 +1,30 @@
...
+        <artifactId>parent</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+        <relativePath>../..</relativePath>
+    </parent>
+    <artifactId>websockets-client</artifactId>
+    <name>Quarkus QE TS: Messaging: Websockets client</name>
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-websockets</artifactId>
```
Sgitario, line 15: This is not correct as this dependency brings both server and client. And according this test: https://github.com/quarkus-qe/quarkus-test-suite/pull/273/files#diff-ea69dcb1d9ecd506ef36c6ffd128ff24529c5c1aa43b2c0bd820417d54d51843R13, you want to verify the client isolately.

Sgitario (changes requested):

messaging/websockets-client/pom.xml
fedinskiy: Added "provided" scope

Sgitario (changes requested):
Why having two modules for this? I don't think it's necessary, we can isolatelly cover the websocket client directly in the test, see my commit: https://github.com/Sgitario/quarkus-test-suite/commit/b7c5ebbc7924d0db29a7f15e79c1aae9905cb2f2

This way, we don't need to use the forced dependencies and the solution will work for both JVM and Native.

fedinskiy:
@Sgitario the solution you're suggesting, does not check the client in isolation in JVM[1] and doesn't check client at all in Native mode or on Openshift[2]. Also it is already implemented in the websockets module.

Existing solution currently works in both jvm and Native module[3].

P.S. I would also want to humbly point to the fact, that the usage of two applications and forced dependencies was your own suggestion(see [4] and gchat thread).

[1] It doesn't isolate server's quarkus-websocket dependency from tests.
[2] Since our tests are not compiled to Native code and aren't deployed to Openshift.
[3] https://github.com/quarkus-qe/quarkus-test-suite/actions/runs/1265836456
[4] https://github.com/quarkus-qe/quarkus-test-suite/pull/273#pullrequestreview-757462171

messaging/websockets-client/src/main/java/io/quarkus/ts/Pusher.java
```diff
@@ -0,0 +1,34 @@
...
+    private Session session;
+
+    @OnOpen
+    public void onOpen(Session session) {
+        this.session = session;
+        //todo messages are not sent on opening of session. Need to investigate and create reproducer
+    }
+
+    @OnMessage
+    public void onMessage(String message) throws ExecutionException, InterruptedException, TimeoutException {
```
pjgg, line 27: the message is not used. Could you log the message (just for debugging purposes)

messaging/websockets/src/main/java/io/quarkus/ts/Pusher.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @OnOpen
+    public void onOpen(Session session) {
+        this.session = session;
+        //todo messages are not sent on opening of session. Need to investigate and create reproducer
+    }
+
+    @OnMessage
+    public void onMessage(String message) throws ExecutionException, InterruptedException, TimeoutException {
+        RemoteEndpoint.Async remote = session.getAsyncRemote();
```
pjgg, line 28: the message is not used. Could you log the message (just for debugging purposes)

pjgg:
I have proposed some cosmetic changes. 

I think that WebSockets module(s) should be under "HTTP" module
`http/websockets`

WDYT @fedinskiy ?

I mean, Websockets is using under the hood HTTP protocol, so could make sense to move on these items into HTTP module.

pjgg (changes requested):

pjgg:
Over all looks good to me

pjgg (approved):

PR #300, opened 2021-09-22
Support JDK 17

pjgg:
Add JDK 17 to GitHub CI

- Quarkus Upstream is still using JDK 11 
- JVM test (Linuz/Win) runs over JDK17
- Native runs over JDK 11

### DIFF
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -43,7 +43,7 @@ jobs:
     needs: build-dependencies
     strategy:
       matrix:
-        java: [ 11 ]
+        java: [ 11, 17 ]
         profiles: [ "root-modules,http-modules,security-modules,monitoring-modules,spring-modules,test-tooling-modules",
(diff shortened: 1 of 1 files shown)


Sgitario (changes requested):
We should not build Quarkus on both 11 and 17, only 11 ++ execute the JVM tests in 11 and 17 both

.github/workflows/ci.yml
```diff
@@ -41,7 +41,7 @@ jobs:
     needs: build-dependencies
     strategy:
       matrix:
-        java: [ 11 ]
+        java: [ 11, 17 ]
```
Sgitario, line 44: I don't think we should validate the format twice, one for 11 and another one for 17.

Sgitario (changes requested):
What about quarkus-snapshot.yaml? I think we should also extend the execution to use 17 in the Quarkus Ecosystem workflow.

PR #332, opened 2021-09-30
Refactor Infinispan to delete cluster when startup failures

Sgitario:
++ To move all the common functionality into a base class.


Sgitario:
run tests

PR #346, opened 2021-10-11
Log min level must overwrite

pjgg:
We need to create a new scenario called `DefaultMinLogLevelIT` because the existing one is forcing the log level to `info`

```
 @BeforeEach
    public void setup() {
        // Reset level to INFO
        setLogLevelTo(Logger.Level.INFO);
    }
```

And in our scenario is important to check the default `quarkus.log.min-level`  is set to `DEFAULT`


Sgitario:
@pjgg this one must be backported to 2.2, right?

PR #351, opened 2021-10-15
Correct usage of MP `@ConfigurationProperties`

Sgitario:
These changes are based on the https://github.com/quarkusio/quarkus/issues/20610 issue and the https://github.com/quarkusio/quarkus/wiki/Migration-Guide-2.4#microprofile-config clarification in the migration guides.


kshpak:
Thanks for update. I will prepare the same for beefy now

PR #362, opened 2021-11-01
Add separate SQL module for Oracle

jsmrcka:
Original RFE: https://issues.redhat.com/browse/QUARKUS-1164.
Related bug: https://issues.redhat.com/browse/QUARKUS-1305.

### DIFF
--- a/sql-db/sql-app-oracle/src/test/java/io/quarkus/ts/sqldb/sqlapp/oracle/AbstractSqlDatabaseIT.java
+++ b/sql-db/sql-app-oracle/src/test/java/io/quarkus/ts/sqldb/sqlapp/oracle/AbstractSqlDatabaseIT.java
@@ -1,0 +1,197 @@
+package io.quarkus.ts.sqldb.sqlapp.oracle;
+
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.hamcrest.Matchers.hasSize;
+
+import org.apache.http.HttpStatus;
+import org.junit.jupiter.api.MethodOrderer;
+import org.junit.jupiter.api.Order;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestMethodOrder;
+
+import io.restassured.http.ContentType;
+
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
+public abstract class AbstractSqlDatabaseIT {
+
+    private static final int EXPECTED_SIZE = 7;
+
+    private static final int VALID_ID = 8;
+    private static final int INVALID_ID = 999;
+
+    @Test
+    @Order(1)
+    public void getAll() {
+        given()
+                .get("/book")
+                .then()
+                .statusCode(HttpStatus.SC_OK)
+                .body("", hasSize(EXPECTED_SIZE));
+    }
+
+    @Test
+    @Order(2)
+    public void get() {
+        given()
+                .get("/book/7")
+                .then()
+                .statusCode(HttpStatus.SC_OK)
+                .body("title", equalTo("Perdido Street Station"))
+                .body("author", equalTo("China Miéville"));
+    }
+
+    @Test
+    @Order(3)
+    public void create() {
+        Book book = new Book();
+        book.title = "Neuromancer";
+        book.author = "William Gibson";
+
+        given()
+                .contentType(ContentType.JSON)
+                .body(book)
+                .post("/book")
+                .then()
+                .statusCode(HttpStatus.SC_CREATED)
+                .body("id", equalTo(VALID_ID))
+                .body("title", equalTo("Neuromancer"))
+                .body("author", equalTo("William Gibson"));
+
+        given()
+                .get("/book/8")
+                .then()
+                .statusCode(HttpStatus.SC_OK)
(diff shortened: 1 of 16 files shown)


fedinskiy:
@jsmrcka the most important question: why create a separate module instead of updating an existing OracleDatabaseIT in sql-app?

And if you still want to create a new module, please add some description into README.md.

jsmrcka:
> @jsmrcka the most important question: why create a separate module instead of updating an existing OracleDatabaseIT in sql-app?
> 
> And if you still want to create a new module, please add some description into README.md.

We need a separate module in order to cover the native scenario, which is not possible in the existing `sql-app` due to the bug linked in the description.
Thanks for spotting the missing README, I'll add a description in it ASAP.

mjurc:
run tests

sql-db/sql-app-oracle/src/test/java/io/quarkus/ts/sqldb/sqlapp/oracle/AbstractSqlDatabaseIT.java
```diff
@@ -0,0 +1,197 @@
...
+    public void update() {
+        Book book = new Book();
+        book.id = Long.valueOf(VALID_ID);
+        book.title = "Schismatrix";
+        book.author = "Bruce Sterling";
+
+        given()
+                .contentType(ContentType.JSON)
+                .body(book)
+                .put("/book/8")
```
fedinskiy, line 111: Could this(as well as similar code on line 137) be replaced with `.get("/book/" + VALID_ID)`? Right now, one need to go to the start of the file to understand, what is happening in these two methods.

sql-db/sql-app-oracle/src/test/java/io/quarkus/ts/sqldb/sqlapp/oracle/AbstractSqlDatabaseIT.java
```diff
@@ -0,0 +1,197 @@
...
+                .get("/book/8")
+                .then()
+                .statusCode(HttpStatus.SC_NOT_FOUND)
+                .body("code", equalTo(HttpStatus.SC_NOT_FOUND))
+                .body("error", equalTo("book '8' not found"));
+    }
+
+    @Test
+    @Order(11)
+    public void deleteWithUnknownId() {
```
fedinskiy, line 189: If I am not mistaken, this test checks that Book.findById(id) properly returns null. Since it doesn't check any DELETE-specific code, I suggest to use GET method here, for the sake of clarity.

sql-db/sql-app-oracle/src/test/java/io/quarkus/ts/sqldb/sqlapp/oracle/AbstractSqlDatabaseIT.java
jsmrcka: So the point is to use the `VALID_ID` constant (and possibly other constants) wherever possible, to avoid confusion, right?
If so, I think I can replace it in a few more occurences.
However, the `put` method is IMO correct here.

sql-db/sql-app-oracle/src/test/java/io/quarkus/ts/sqldb/sqlapp/oracle/AbstractSqlDatabaseIT.java
jsmrcka: You're right, let me replace it with `getWithUnknownId`. I think the same goes for `updateWithUnknownId`.

sql-db/sql-app-oracle/src/test/java/io/quarkus/ts/sqldb/sqlapp/oracle/AbstractSqlDatabaseIT.java
fedinskiy: Yes, "put" is definitely belongs here :)

No, you don't need to use VALID_ID constant  everywhere, just use(or don't use) it consistently inside the same method. When we create book with id=VALID_ID, but access it through "books/8" URL, it's not really clear, that these ids are connected.

fedinskiy:
@jsmrcka I approve the changes, but there are some formatting problems, fix them, please.

fedinskiy (approved):

jsmrcka:
> @jsmrcka I approve the changes, but there are some formatting problems, fix them, please.

@fedinskiy Thanks. The failed checks are being addressed in https://github.com/quarkus-qe/quarkus-test-suite/pull/366.

jsmrcka:
run tests

jsmrcka:
run tests

mjurc:
run tests

mjurc:
Native failures are unrelated

@jsmrcka Thanks for the feature work. Merging!

mjurc:
@jsmrcka please backport to 2.2 branch, thank you

jsmrcka:
The native check didn't pass and it appears that the workaround in [this commit](https://github.com/quarkus-qe/quarkus-test-suite/pull/362/commits/5b8319a41b612d65f80312702aa6794eb0390994) is no good anyway. The reason of the failure is being investigated in https://github.com/quarkus-qe/quarkus-test-suite/issues/372.
Sorry for the mess, we should have sorted this out before the merge.

I propose to create a PR with a revert of the commit. Then, I can backport without the commit and its revert.
@mjurc WDYT?

PR #386, opened 2021-11-18
Add reactive counterpart of `http-advanced`

jsmrcka:
Replaced RESTEasy Classic extensions with RESTEasy Reactive extensions.
Removed unused extensions.
Switched to reactive execution model wherever it makes sense.
Test cases' names changed to `*ReactiveIT`.

PR #388, opened 2021-11-18
Introduce reactive counterparts of http-minimum and rest-client modules under /http

kshpak:
Main things to look at:
1. Parent pom.xml references
2. New reactive package in counterparts
3. Test classes name should reflect they're Reactive
4. Update endpoints from blocking to non-blocking signature
5. Update dependencies in the module's POM

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,8 +136,14 @@ It also verifies multiple deployment strategies like:
 - Using OpenShift quarkus extension
 - Using OpenShift quarkus extension and Docker Build strategy
 
+### `http/http-minimum-reactive`
+Reactive equivalent of the http/http-minimum module
+
 ### `http/http-advanced`
 Verifies Server/Client http_2/1.1, Grpc and http redirections.
+
+### `http/http-advanced-reactive`
+Reactive equivalent of the http/http-advanced module
 
 ### `http/http-static`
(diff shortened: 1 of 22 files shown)


kshpak:
Looking into native CI fail with rest-client-reactive

jsmrcka (changes requested):
The two new modules are missing from the `README.md`. I think that something like this would be enough:
> Reactive equivalent of `http/http-minimum`.
> Reactive equivalent of `http/rest-client`.

Moreover, I am afraid I also forgot to add `http-advanced` to the README in https://github.com/quarkus-qe/quarkus-test-suite/pull/386. Could you please do it in this PR?
> Reactive equivalent of `http-advanced`.

As for the lack of XML serialization support in reactive rest client: is it documented somewhere? I was not able to find any explicit info on this. We should at least mention in the README, and add a link to the docs if possible.
If it is an open issue in quarkus (i.e. the support is expected to be added in foreseeable future), we should keep the XML-related code in the module, just disable the test and tag it with a link to the issue.

kshpak:
@jsmrcka README updated

jsmrcka (approved):

pjgg (approved):

PR #392, opened 2021-11-22
DockerHub limits: Moves hyness/spring-cloud-config-server to quay

pjgg:
We have reached dockerHub request limits on Jenkins. 

This PR moves `hyness/spring-cloud-config-server` image to `quay.io/quarkusqeteam/spring-cloud-config-server`


jsmrcka (approved):
Btw. what's the strategy here? Do we move images in an ad hoc way, upon hitting the limit? Or do we want to preemptively move as many as possible?

pjgg:
> Do we move images in an ad hoc way, upon hitting the limit

Many as possible. We should not depend on docker hub rate limits.

rsvoboda:
> Many as possible. We should not depend on docker hub rate limits.

Not priority for this or next week, pls create trello card to prioritize the effort. Plus add candidates for the move so we can discuss it.

PR #403, opened 2021-12-02
Update github actions versions for Setup-java and Checkout

kshpak:
Setup-java@2 supports now caching, which behaves by default the same way as it was configured in our actions/cache@v1.
Performance improvement:
1. We don't need caching on **Build Dependencies** and **Validate Format**. **Build Dependencies** already performs zipping of the **~/.m2/repository** folder after the build and this zipped archive is unzipped in subsequent steps. Caching at this stage will only create a new cache key with only dependencies from Quarkus Build which we will hit all the time in the next steps (if any of the pom.xml files _was not_ changed!).
Performing caching at the stage of `Test Suite Build` will create a new cache key if any of the pom.xml was changed and will save to this cache all the resolved dependencies during the build (this was not the case before). This way we will have cached dependencies for the next CI run (e.x. when PR failed; the next Daily run)
Previously we were basically zipping and unzipping only dependencies resolved during the build and caching action was not helpful.
*Action/Cache creates a new key that _cannot be updated_.
2. Setup-java@2 now supports caching on our Daily build (basically on `scheduled` event)

Check the last 3 actions in https://github.com/kshpak/quarkus-test-suite/actions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,16 +10,10 @@ jobs:
         java: [ 11 ]
     steps:
       - uses: actions/checkout@v1
-      - uses: actions/cache@v1
-        with:
-          path: ~/.m2/repository
-          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
-          restore-keys: |
-            ${{ runner.os }}-maven-
       - name: Reclaim Disk Space
         run: .github/ci-prerequisites.sh
       - name: Install JDK {{ matrix.java }}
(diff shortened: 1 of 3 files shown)


kshpak:
THE CI failures are unrelated to this PR

rsvoboda (approved):
Changes look fine to me.

For @pjgg - https://github.com/actions/setup-java#caching-packages-dependencies describes the caching, initial input to look into this topic comes from https://twitter.com/rieckpil/status/1460291056751235078?s=27

jsmrcka (approved):
Looks good. Thanks for the changes and the explanation.

PR #423, opened 2021-12-17
Use a non default devService Kafka version on user experience tests

pjgg:
The daily build is failing because the user-experience Kafkadevservice scenario is failing. 

I could not reproduce the error locally but this scenario is checking the default log messages logged by kafka devservices. More in detail is checking the default messages when redpanda docker image doesn't exist. The scenario remove redpanda docker image as a previous test step. In order to avoid some conflicts with other scenarios that also are using this default image, let's use a custom image for this case. 

`kafkaContainerShouldBeStarted` test was failing because the started log has changed


messaging/kafka-streams-reactive-messaging/src/test/java/io/quarkus/ts/messaging/kafka/DevModeRedPandaDevServiceUserExperienceIT.java
```diff
@@ -15,7 +15,7 @@
 @QuarkusScenario
 public class DevModeRedPandaDevServiceUserExperienceIT {
 
-    private static final String RED_PANDA_VERSION = "latest";
+    private static final String RED_PANDA_VERSION = "v21.10.3";
```
rsvoboda, line 18: We are checking just logs, not concerned about the latest docker image being used

rsvoboda (approved):

PR #434, opened 2021-12-23
Oracle JDBC extension is now able to handle presence of other JDBC drivers in native

rsvoboda:
Oracle JDBC extension is now able to handle presence of other JDBC drivers in native

Resolved in https://github.com/quarkusio/quarkus/pull/22255


rsvoboda:
Failure is not related to this change, needs investigation though.

rsvoboda:
Second run of CI is green

PR #437, opened 2022-01-03
Quarkus app should startup without domain socket

pjgg:
Quarkus application should start even with property `quarkus.http.host-enabled=false`


jsmrcka (approved):
Looks good.

jsmrcka:
The CI failures are caused by Jenkins infrastructure issues, not related to this PR.

jsmrcka:
@pjgg I did not notice it before the merge, but the commit could use a better subject.

PR #438, opened 2022-01-03
Add docker image name with empty spaces scenario

pjgg:
issue Ref: https://github.com/quarkusio/quarkus/pull/19958

A Quarkus app that uses the extension `container-image-docker` fails if the docker image name (`quarkus.application.name`) contains some spaces at the end. 

The issue has been fixed.  Verified by this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -182,8 +182,9 @@ jobs:
         run: |
           if [[ -n ${MODULES_ARG} ]]; then
             echo "Running modules: ${MODULES_ARG}"
-            mvn -fae -V -B -s .github/mvn-settings.xml -fae -Dall-modules -pl $MODULES_ARG clean verify -Dnative \
-                        -Dinclude.quarkus-cli-tests -Dts.quarkus.cli.cmd="${PWD}/quarkus-dev-cli"
+            mvn -fae -V -B -s .github/mvn-settings.xml -fae -Dall-modules -Dquarkus.container-image.build=false \
+                        -pl $MODULES_ARG clean verify -Dnative -Dinclude.quarkus-cli-tests \
+                        -Dts.quarkus.cli.cmd="${PWD}/quarkus-dev-cli"
           fi
       - name: Zip Artifacts
         if: failure()
@@ -229,7 +230,7 @@ jobs:
             MODULES_MAVEN_PARAM="-pl ${MODULES_ARG}"
           fi
 
-          mvn -fae -s .github/mvn-settings.xml clean verify -Dall-modules $MODULES_MAVEN_PARAM
+          mvn -fae -s .github/mvn-settings.xml clean verify -Dall-modules -Dquarkus.container-image.build=false $MODULES_MAVEN_PARAM
       - name: Zip Artifacts
         shell: bash
         if: failure()
--- a/README.md
+++ b/README.md
@@ -385,6 +385,9 @@ Tests:
 Checks that the application can read configuration from a ConfigMap and a Secret.
 The ConfigMap/Secret is exposed by mounting it into the container file system or the Kubernetes API server.
 
+### `docker-build`
+Checks that an application can generate a Docker image based on some configuration parameters.
+
 ### `lifecycle-application`
 Verifies lifecycle application features like `@QuarkusMain` and `@CommandLineArguments`.
 Also ensures maven profile activation with properties and additional repository definition propagation into Quarkus maven plugin.
--- a/docker-build/src/main/docker/Dockerfile.legacy-jar
+++ b/docker-build/src/main/docker/Dockerfile.legacy-jar
@@ -1,0 +1,37 @@
+####
+# This Dockerfile is used in order to build a container that runs the Quarkus application in JVM mode
+#
+# Before building the container image run:
+#
+# ./mvnw package -Dquarkus.package.type=legacy-jar
+#
+# Then, build the image with:
+#
+# docker build -f src/main/docker/Dockerfile.legacy-jar -t quarkus/getting-started-legacy-jar .
+#
+# Then run the container using:
+#
+# docker run -i --rm -p 8080:8080 quarkus/getting-started-legacy-jar
+#
+# If you want to include the debug port into your docker image
+# you will have to expose the debug port (default 5005) like this :  EXPOSE 8080 5005
+#
+# Then run the container using :
+#
+# docker run -i --rm -p 8080:8080 quarkus/getting-started-legacy-jar
+#
+###
+FROM registry.access.redhat.com/ubi8/openjdk-11-runtime:1.10
+
+ENV LANG='en_US.UTF-8' LANGUAGE='en_US:en'
+
+# Configure the JAVA_OPTIONS, you can add -XshowSettings:vm to also display the heap size.
+ENV JAVA_OPTIONS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager"
+
+COPY target/lib/* /deployments/lib/
+COPY target/*-runner.jar /deployments/quarkus-run.jar
+
+EXPOSE 8080
+USER 185
+
+ENTRYPOINT [ "java", "-jar", "/deployments/quarkus-run.jar" ]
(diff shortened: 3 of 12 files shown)


docker-build/src/test/java/io/quarkus/ts/docker/DockerBuildIT.java
```diff
@@ -0,0 +1,32 @@
...
+import org.junit.jupiter.api.Test;
+
+import com.github.dockerjava.api.model.Image;
+
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.scenarios.annotations.DisabledOnNative;
+import io.quarkus.test.utils.DockerUtils;
+
+@QuarkusScenario
+@DisabledOnNative // in order to avoid extra test time. In my opinion native mode doesn't add value for this scenario
```
fedinskiy, line 15: If the test is disabled on Native, do we still need https://github.com/quarkus-qe/quarkus-test-suite/pull/438/files#diff-873490bdd3f9d04392641dd37b052242fb566f181517f6f5fabeabdce22e85a7 ?

docker-build/src/test/java/io/quarkus/ts/docker/DockerBuildIT.java
```diff
@@ -0,0 +1,32 @@
...
+public class DockerBuildIT {
+
+    private static final String SPACE = "      ";
```
fedinskiy, line 18: I suggest to create a check for a tabulation('\t') symbol

docker-build/src/test/java/io/quarkus/ts/docker/DockerBuildIT.java
```diff
@@ -0,0 +1,32 @@
...
+    private static final String DOCKER_IMG_VERSION = "1.0.0";
+
+    @AfterAll
+    public static void tearDown() {
+        DockerUtils.removeImage(PADDED_DOCKER_IMG_NAME, DOCKER_IMG_VERSION);
+    }
+
+    @Test
+    public void buildDockerImage() {
+        Image dockerImg = DockerUtils.getImage("qe/hello-world-app", "1.0.0");
```
fedinskiy, line 29: I suggest to also check for PADDED_<...>_NAME. And definitely to use DOCKER_IMG_VERSION.

README.md
```diff
@@ -385,6 +385,9 @@ Tests:
 Checks that the application can read configuration from a ConfigMap and a Secret.
 The ConfigMap/Secret is exposed by mounting it into the container file system or the Kubernetes API server.
 
+### `docker-build`
+Checks that an application can generate a Docker image based in some configuration parameters.
```
fedinskiy, line 389: I presume that "in" is a typo, and it should be "on" instead.

docker-build/src/test/java/io/quarkus/ts/docker/DockerBuildIT.java
```diff
@@ -0,0 +1,32 @@
...
+        assertFalse(dockerImg.getId().isEmpty(), "Docker images was not created");
```
fedinskiy, line 30: I can see only one image being created. Am I missing something?

fedinskiy (changes requested):

README.md
pjgg: these files are generated by `mvn io.quarkus.platform:quarkus-maven-plugin:2.6.1.Final:create` command by default. It's true that we are not going to use it, but I prefer to leave it as it is.

docker-build/src/test/java/io/quarkus/ts/docker/DockerBuildIT.java
pjgg: these files are generated by `mvn io.quarkus.platform:quarkus-maven-plugin:2.6.1.Final:create` command by default. It's true that we are not going to use it, but I prefer to leave it as it is.

docker-build/src/test/java/io/quarkus/ts/docker/DockerBuildIT.java
pjgg: No, is fine. We are only creating a single image `qe/hello-world-app:1.0.0`, which also is removed at the end of the test.

fedinskiy: In that case it's probably "image" and not "images" :)

fedinskiy (changes requested):
... and the tabulation tests was not commited

fedinskiy:
@pjgg you check the tabulation(and I have doubts, that it needs to be shielded in this case) but do not check spaces. Are you sure about that?

pjgg:
> @pjgg you check the tabulation(and I have doubts, that it needs to be shielded in this case) but do not check spaces. Are you sure about that?

Tabulation and space on the test are not required anymore in order to reproduce the issue. 

In order to reproduce you can downgrade the Quarkus version to 2.0.0.Final and run a `mvn clean verify`

fedinskiy (approved):

.github/workflows/ci.yml
```diff
@@ -182,8 +182,9 @@ jobs:
         run: |
           if [[ -n ${MODULES_ARG} ]]; then
             echo "Running modules: ${MODULES_ARG}"
-            mvn -fae -V -B -s .github/mvn-settings.xml -fae -Dall-modules -pl $MODULES_ARG clean verify -Dnative \
-                        -Dinclude.quarkus-cli-tests -Dts.quarkus.cli.cmd="${PWD}/quarkus-dev-cli"
+            mvn -fae -V -B -s .github/mvn-settings.xml -fae -Dall-modules -Dquarkus.container-image.build=false \
```
rsvoboda, line 185: Why is `-Dquarkus.container-image.build=false` needed to be added to the GH actions here and in the other places?

docker-build/src/main/docker/Dockerfile.legacy-jar
```diff
@@ -0,0 +1,37 @@
+####
```
rsvoboda, line 1: Why is this file added when not used?

docker-build/src/main/docker/Dockerfile.native
```diff
@@ -0,0 +1,27 @@
+####
```
rsvoboda, line 1: Why is this file added when not used?

docker-build/src/main/docker/Dockerfile.native-distroless
```diff
@@ -0,0 +1,23 @@
+####
```
rsvoboda, line 1: Why is this file added when not used?

docker-build/src/main/resources/application.properties
```diff
@@ -0,0 +1,5 @@
+quarkus.container-image.build=true
```
rsvoboda, line 1: BTW, in GH action this get overridden to false

.github/workflows/ci.yml
rsvoboda: Comment above the `mvn` with some explanation should be present?

Is it because GH Action runner is container and you can't invoke docker build from it?

docker-build/src/main/docker/Dockerfile.native
rsvoboda: Seems it's used by `NativeConfigurationIT` added after this module. So I guess this file can stay.

docker-build/src/test/java/io/quarkus/ts/docker/DockerBuildIT.java
```diff
@@ -0,0 +1,33 @@
...
+import org.junit.jupiter.api.Test;
+
+import com.github.dockerjava.api.model.Image;
+
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.scenarios.annotations.DisabledOnNative;
+import io.quarkus.test.utils.DockerUtils;
+
+@QuarkusScenario
+@DisabledOnNative(reason = "Due to reduce native execution time")
```
rsvoboda, line 17: `NativeConfigurationIT` is executed in native, the native binary gets produced in native run of the TS, probably no harm to have the test enabled.

.github/workflows/ci.yml
rsvoboda: If this is about windows not having containers available, the property should be set in pom.xml of the module, there is already profile for Windows

PR #440, opened 2022-01-04
Add coverage of `quarkus-test-security` extension

jsmrcka:
Verifies that the `quarkus.security.jaxrs.deny-unannotated-endpoints`
property and  `@TestSecurity` annotation work correctly together and in
combination with Basic Auth.

### DIFF
--- a/security/basic/src/test/java/io/quarkus/ts/openshift/security/basic/SecurityTest.java
+++ b/security/basic/src/test/java/io/quarkus/ts/openshift/security/basic/SecurityTest.java
@@ -1,0 +1,71 @@
+package io.quarkus.ts.openshift.security.basic;
+
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.Matchers.containsString;
+
+import org.apache.http.HttpStatus;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.junit.QuarkusTest;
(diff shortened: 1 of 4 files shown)


security/basic/src/test/java/io/quarkus/ts/openshift/security/basic/SecurityTest.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Test
+    @TestSecurity(authorizationEnabled = false)
+    public void allowUnannotated() {
+        given().get("/unannotated")
+                .then()
+                .statusCode(HttpStatus.SC_OK)
+                .body(equalTo("Hello!"));
+    }
```
pjgg, line 48: I am missing an scenario that cover this feature:

```
@TestSecurity(user = "isaac", roles = { "user" })
```

Basically, we need to add another test that validates that we are setting a user and a role by the above annotation. 

The test could be something like:
```
@Test
    @TestSecurity(user = "testUser", roles = { "user" })
    public void getUser() {
        given().get("/user")
                .then()
                .statusCode(HttpStatus.SC_OK)
                .body(containsString("testUser"));
    }
```
and / or

```
@Test
    @TestSecurity(user = "testUser", roles = { "admin" })
    public void getUser() {
        given().get("/user")
                .then()
                .statusCode(HttpStatus.SC_FORBIDDEN);
    }
```

pjgg (changes requested):

pjgg (approved):

PR #442, opened 2022-01-04
Remove some RegisterForReflection annotation that are not required any more

pjgg:
This PR talks about native mode

Some `RegisterForReflection` annotations are not required based on this commit: https://github.com/quarkusio/quarkus/pull/20225

Also, we have added a new scenario with a data structure as is described on this issue:
https://github.com/quarkusio/quarkus/issues/20200

### DIFF
--- a/config/src/main/java/io/quarkus/ts/configmap/api/server/Hello.java
+++ b/config/src/main/java/io/quarkus/ts/configmap/api/server/Hello.java
@@ -1,8 +1,5 @@
 package io.quarkus.ts.configmap.api.server;
 
-import io.quarkus.runtime.annotations.RegisterForReflection;
-
-@RegisterForReflection
 public class Hello {
     private final String content;
 
--- a/http/http-minimum-reactive/src/test/java/io/quarkus/ts/http/minimum/reactive/HttpMinimumReactiveIT.java
+++ b/http/http-minimum-reactive/src/test/java/io/quarkus/ts/http/minimum/reactive/HttpMinimumReactiveIT.java
@@ -2,8 +2,10 @@ package io.quarkus.ts.http.minimum.reactive;
 
 import static io.restassured.RestAssured.given;
 import static org.hamcrest.CoreMatchers.is;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import org.apache.http.HttpStatus;
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
 import io.quarkus.test.scenarios.QuarkusScenario;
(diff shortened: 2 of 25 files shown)


http/http-minimum-reactive/src/test/java/io/quarkus/ts/http/minimum/reactive/HttpMinimumReactiveIT.java
```diff
@@ -15,4 +16,15 @@
     public void httpServer() {
         given().get("/api/hello").then().statusCode(HttpStatus.SC_OK).body("content", is("Hello, World!"));
     }
+
+    @Test
+    public void httpClusterConfig() {
```
jsmrcka, line 21: What is the added value of this test, compared to the `httpServer` test? If it is just the coverage of hierarchical data structure and Jackson annotations, then the data structure seems unnecessarily complicated to me.

Why is the test method called `httpClusterConfig`? I would name it accordingly to the point of the test (see previous question).

Do we want it also in the non-reactive module `http/http-minimum`?

config/src/main/java/io/quarkus/ts/configmap/api/server/Hello.java
```diff
@@ -1,8 +1,5 @@
 package io.quarkus.ts.configmap.api.server;
 
-import io.quarkus.runtime.annotations.RegisterForReflection;
-
-@RegisterForReflection
```
jsmrcka, line 5: Could you elaborate (perhaps in the commit message) on which uses of the annotation can we get rid of? I don't see any pattern.
If it's just me not getting something that should be obvious from e.g. looking at the https://github.com/quarkusio/quarkus/pull/20225 PR, then a simple hint in this discussion would do. Thanks. :)

pjgg: Is not obvious. There are some exceptions as you can see [on this commit ](quarkusio/quarkus#20225)  but in all the other cases I have tested it manually and I realize that the register for reflection annotation was redundant. 

Take a look to [this documentation](https://quarkus.io/guides/writing-native-applications-tips#registering-for-reflection):

```
When building a native executable, GraalVM operates with a closed world assumption. It analyzes the call tree and removes all the classes/methods/fields that are not used directly.
```

So seems that Mandrel is clever enough to see that this code is used directly, so is not removed (and then doesn't need the annotation "@RegisterForReflection"). 

In most of the cases, you will not need this annotation, but in some other cases as "Reflexion" you will need to tag this code by the annotation "@RegisterForReflection"

http/http-minimum-reactive/src/test/java/io/quarkus/ts/http/minimum/reactive/HttpMinimumReactiveIT.java
pjgg: [The issue](https://github.com/quarkusio/quarkus/issues/20200) only talks about reactive, so doesn't apply to classic scenarios. 

As you are pointing the additional value is the data structure more in detail the list serialization os a "complex" pojo. I prefer to leave the data structure as it is because is a mirror of the origin reproducer.

Tag was added and test renamed to `nativeListSerialization`

config/src/main/java/io/quarkus/ts/configmap/api/server/Hello.java
jsmrcka: Ok, thanks for the clarification.

jsmrcka (approved):

PR #445, opened 2022-01-05
native coverage: reflections registration of constructors

pjgg:
original issue Ref: https://github.com/quarkusio/quarkus/issues/19711

A new issue was created: https://github.com/quarkusio/quarkus/issues/22654

### DIFF
--- a/http/jaxrs/src/main/java/io/quarkus/ts/security/core/GreetingResource.java
+++ b/http/jaxrs/src/main/java/io/quarkus/ts/security/core/GreetingResource.java
@@ -1,17 +1,127 @@
 package io.quarkus.ts.security.core;
 
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.MediaType;
 
-@Deprecated
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.json.JsonMapper;
+
 @Path("/hello")
 public class GreetingResource {
+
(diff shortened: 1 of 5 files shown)


http/jaxrs/src/test/java/io/quarkus/ts/security/core/SerializationIT.java
```diff
@@ -0,0 +1,51 @@
...
+    public void serializeList() {
+        String result = makeQuery("/hello/serialize-list").asString();
+        Assertions.assertEquals("[Hello RESTEasy]", result);
+    }
+
+    @Test
+    @Disabled("https://github.com/quarkusio/quarkus/issues/22654")
+    public void serializeBigList() {
+        String basePath = "/hello/big-serialize-list?expSize=";
+        mustBeGreaterThan(makeQuery(basePath + (LONGEST_EAGER_ALLOC - 39)).jsonPath().getInt("size"), 0);
```
fedinskiy, line 33: Too many magic numbers in the method. Probably, these should be separate constants

http/jaxrs/src/main/java/io/quarkus/ts/security/core/GreetingResource.java
```diff
@@ -1,17 +1,127 @@
...
+
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    @Path("/serialize-list")
+    public String serlist() throws IOException, ClassNotFoundException {
+
+        byte[] bytes;
+
+        try (
+                ByteArrayOutputStream baos = new ByteArrayOutputStream();
```
fedinskiy, line 64: Can we use jdkSerialize and jdkDeserialize in all methods of this class?

fedinskiy (changes requested):

http/jaxrs/src/main/java/io/quarkus/ts/security/core/GreetingResource.java
pjgg: We have an issue with the native mode that is under investigation. Have a look at the description of this ticket. I think that maybe in the future `jdkSerialize` and `jdkDeserialize` are going to be removed.

http/jaxrs/src/test/java/io/quarkus/ts/security/core/SerializationIT.java
pjgg: These "magic numbers" are random variations just to tease out possible edge cases. 
The original scenario is the following one: https://github.com/FasterXML/jackson-databind/commit/3ccde7d938fea547e598fdefe9a82cff37fed5cb

fedinskiy (approved):

PR #453, opened 2022-01-06
Update sql-db/sql-app and GraalVM / Mandrel

rsvoboda:
Update sql-db/sql-app and GraalVM / Mandrel


rsvoboda:
Should fix daily run fails for `sql-db/sql-app`, at least this worked for me locally.

PR #464, opened 2022-01-10
reactive-hibernate propagate context scenario

pjgg:
Run reactive rest client on Vertx same context scenario


sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/http/GroundedEndpoint.java
```diff
@@ -93,6 +101,27 @@ public GroundedEndpoint() {
...
     }
 
+    // https://github.com/quarkusio/quarkus/issues/18977
+    @GET
+    @Path("bookThroughSession/{authorName}/{name}")
+    public Uni<Author> createBookThroughSession(String authorName, String name) {
+        Author author = new Author();
+        author.setName(authorName);
+        return factory.withTransaction(
+                (session, transaction) -> factory.withSession(session1 -> session1.persist(author).chain(session::flush))
```
fedinskiy, line 111: Is opening a second session required to reproduce the bug?

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/http/GroundedEndpoint.java
```diff
@@ -93,6 +101,27 @@ public GroundedEndpoint() {
...
+                        .onItem().call(() -> someApi.doSomething())
+                        .onItem().transformToUni((ignore -> factory.withSession(session2 -> {
+                            Book book = new Book();
+                            book.setAuthor(author.getId());
+                            book.setTitle(name);
+                            return session2.persist(book);
+
```
fedinskiy, line 118: nitpick: empty line

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/http/GroundedEndpoint.java
fedinskiy: And even if we are going this way, can we, at least, have more uniformed session names? If there are three sessions, let's name them "session1, session2 and session3"(or, even better, "firstSession, secondSession and thirdSession").

fedinskiy (changes requested):

sql-db/hibernate-reactive/src/main/java/io/quarkus/ts/reactive/http/GroundedEndpoint.java
pjgg: > Is opening a second session required to reproduce the bug?

yes is required. 

> f there are three sessions, let's name them "session1, session2 and session3"(or, even better, "firstSession, secondSession and thirdSession")

I prefer as it's, the suggested way is too verbose

fedinskiy: What about "session1" instead of "session", "session2" instead of "session1" and so on?

fedinskiy (approved):

PR #480, opened 2022-01-12
Restored Docker file for native runs

fedinskiy:


rsvoboda (changes requested):
image needs to be ubi8 based - see https://github.com/quarkusio/quarkus/pull/22617

pjgg (approved):

rsvoboda (changes requested):
I see this when I build sample app using Quarkus main:
```
cat getting-started/src/main/docker/Dockerfile.native
####
# This Dockerfile is used in order to build a container that runs the Quarkus application in native (no JVM) mode.
#
# Before building the container image run:
#
# ./mvnw package -Pnative
#
# Then, build the image with:
#
# docker build -f src/main/docker/Dockerfile.native -t quarkus/getting-started .
#
# Then run the container using:
#
# docker run -i --rm -p 8080:8080 quarkus/getting-started
#
###
FROM registry.access.redhat.com/ubi8/ubi-minimal:8.5
WORKDIR /work/
RUN chown 1001 /work \
    && chmod "g+rwX" /work \
    && chown 1001:root /work
COPY --chown=1001:root target/*-runner /work/application

EXPOSE 8080
USER 1001

CMD ["./application", "-Dquarkus.http.host=0.0.0.0"]
```

Template https://github.com/quarkusio/quarkus/blob/main/independent-projects/tools/base-codestarts/src/main/resources/codestarts/quarkus/tooling/dockerfiles/base/src/main/docker/Dockerfile.tpl.qute.native also confirms that the Dockerfile.native shoul look differently than what's in PR

pjgg:
depends on: #481

PR #492, opened 2022-01-20
Rename some customer cases test names in order to improve readability

pjgg:


spring/spring-data/src/test/java/io/quarkus/ts/spring/data/primitivetypes/BookResourceIT.java
```diff
@@ -82,7 +82,7 @@ void testJpaFieldsMapping() {
```
rsvoboda, line 82: `BookResourceIT.java` name is not great. Can we have more descriptive class name about what it is testing and not what endpoint it is using for testing?

rsvoboda (changes requested):

spring/spring-data/src/test/java/io/quarkus/ts/spring/data/primitivetypes/BookResourceIT.java
pjgg: In this case, This scenario "BookResourceIT" is checking several cases:

- https://github.com/quarkusio/quarkus/pull/13015
- https://github.com/quarkusio/quarkus/issues/13234
- QUARKUS-1521

I think that package name "io.quarkus.ts.spring.data.primitivetypes" is good enough in order to understand, what this scenario is talking about. WDYT?

rsvoboda: The coverage is around CrudRepository, PanacheRepository and primitive types, `BookResourceIT` naming doesn't indicate that.

My point is mainly for Polarion, there is no notion of maven modules, the details tracked there are basically package name and `.java` file name. So file name matters there too. 

`BookResourceIT` => `PersistenceRepositoryIT` / `SpringDataPersistenceIT` / `SpringDataBookResourceIT`

I probably like the last one (`SpringDataBookResourceIT`) the most. It indicates the resource name (easier investigation) and the first part should be enough for context for Polarion

rsvoboda: And the same naming concern applies to `CatResourceIT`

rsvoboda (changes requested):
https://github.com/quarkus-qe/quarkus-test-suite/pull/492#discussion_r788845023

rsvoboda (approved):

PR #533, opened 2022-02-14
Replace Vertx HttpClient by restAssurance in PostgresPoolTest scenario

pjgg:
Vertx-sql module is failing with FIPS enabled. The issue is related to the test itself, looks that Mutiny HTTPClient fails with a default configuration. 

Replace Vertx HttpClient by RestAssurance

### DIFF
--- a/sql-db/vertx-sql/src/main/resources/application.properties
+++ b/sql-db/vertx-sql/src/main/resources/application.properties
@@ -4,7 +4,6 @@ app.selected.db=postgresql
 
 # Quarkus
 quarkus.http.port=8082
-quarkus.http.test.port=8081
 
 ## Postgresql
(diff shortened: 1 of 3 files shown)


sql-db/vertx-sql/src/main/resources/application.properties
```diff
@@ -4,7 +4,7 @@ app.selected.db=postgresql
 
 # Quarkus
 quarkus.http.port=8082
-quarkus.http.test.port=8081
```
fedinskiy, line 7: Why do we need this change?

pjgg: You are right is not required anymore (in this scenario)

PR #536, opened 2022-02-15
Move main to Quarkus 2.7

rsvoboda:
Move main to Quarkus 2.7

Revert excludes applicable just for Quarkus 2.8+/main
Adjust CI scripts
Adjust modules used in Ecosystem CI


mjurc (approved):
Looks legit, thanks for update.

There seems to be a failure in CLI tests - I have scheduled another run as I haven't been able to reproduce.

If the failure persists, I think we should create an issue for it and update it as a part of that so this PR is not blocked.

PR #538, opened 2022-02-17
Rename qute test and add FIPS patch

pjgg:
Qute test `ExampleResourceTest` was renamed to `LocalizedMessagesTest` and also tests names should have a meaning, so also was renamed. This is important to us because then later we use these names in order to generate coverage reports.


qute/multimodule/qute-test/src/test/resources/application.properties
```diff
@@ -0,0 +1 @@
+quarkus.native.additional-build-args=-Dcom.redhat.fips=false
```
rsvoboda, line 1: Drop this, not necessary in main

rsvoboda (changes requested):

fedinskiy (approved):
But, please, look into failed checks.

rsvoboda:
@pjgg please drop " quarkus.native.additional-build-args=-Dcom.redhat.fips=false " part

pjgg:
> But, please, look into failed checks.

must be failing because I have updated `qute-test` module, which requires `qute-api` artifact. CI `detect-test-suite-modules` changes only see changes in `qute-test` so... we are missing `qute-api` dependency on this PR.

PR #539, opened 2022-02-17
[2.2] Increase infinispan container timeout to 60 sec

pjgg:


mjurc (approved):
LGTM, thanks for the fix.

PR #544, opened 2022-02-22
Disable native external app scenarios because S2I is not supported

pjgg:


mjurc (changes requested):
Wouldn't it be better to exclude the module from native profile? Any time we add a test, the developer will have to exclude test manually in the module.

pjgg:
> the developer will have to exclude test manually in the module.

But `native` profile doesn't have "modules" is just a set of properties

mjurc (approved):

PR #549, opened 2022-02-28
Fix fails due to DockerHub limit

fedinskiy:


pom.xml
```diff
@@ -206,7 +206,7 @@
                                 <mariadb.10.image>quay.io/bitnami/mariadb:10.6</mariadb.10.image>
                                 <mariadb.102.image>registry.access.redhat.com/rhscl/mariadb-102-rhel7</mariadb.102.image>
                                 <mssql.image>mcr.microsoft.com/mssql/rhel/server:2019-latest</mssql.image>
-                                <oracle.image>gvenzl/oracle-xe:21-slim</oracle.image>
+                                <oracle.image>quay.io/quarkusqeteam/oracle-xe:21.3.0-slim</oracle.image>
```
mjurc, line 209: Let's not specify micro-release here and go with just `oracle-xe:21-slim`, even if it is going to be in Quay

sql-db/sql-app/src/test/java/io/quarkus/ts/sqldb/sqlapp/DevModeMariadbDevServicesUserExperienceIT.java
```diff
@@ -17,7 +17,7 @@
 public class DevModeMariadbDevServicesUserExperienceIT {
 
     private static final String MARIA_DB_VERSION = "5.5.49";
```
mjurc, line 19: This is an ancient tag, last pushed into 6 years ago - we should definitely bump it to 10.8 (I guess), and drop the micro release version

sql-db/sql-app/src/test/java/io/quarkus/ts/sqldb/sqlapp/DevModeMysqlIT.java
```diff
@@ -6,16 +6,22 @@
...
 import io.quarkus.test.scenarios.QuarkusScenario;
 import io.quarkus.test.services.DevModeQuarkusApplication;
+import io.quarkus.test.utils.SocketUtils;
 
 @Tag("QUARKUS-959")
 @QuarkusScenario
 public class DevModeMysqlIT extends AbstractSqlDatabaseIT {
 
+    private static final String DB_IMAGE = "quay.io/quarkusqeteam/mysql";
+    private static final String DB_VERSION = "8.0.24";
```
mjurc, line 16: Let's drop the micro here aswell and use `8.0`

mjurc (changes requested):

rsvoboda:
Closing as we are moving back to the dockerhub as bitnami is gone from quay

Issue #521, opened 2022-02-08
Re-enable qute/multimodule


fedinskiy:
Was disabled in https://github.com/quarkus-qe/quarkus-test-suite/pull/520 but https://github.com/quarkusio/quarkus/pull/23473 fixes this problem.Should be enabled back in several days

rsvoboda:
Module was disabled because the #23473 PR went in and now the build throws error. Anyway, this is planned for next week in trello, now we need to process rhbq.

rsvoboda:
Qute fix was backported to Quarkus 2.7.3.Final

Issue #226, opened 2021-09-06
Review and remove usage of @QuarkusTest


mjurc:
There are still some tests that use `@QuarkusTest` rather than `@QuarkusScenario`. Review these tests and reimplement them with new framework.

rsvoboda:
```
for i in `git grep -l io.quarkus.test.junit.QuarkusTest`; do
  echo "$i"
  git blame $i | grep 'io.quarkus.test.junit.QuarkusTest';
  echo ""
done
http/http-minimum-reactive/src/test/resources/HelloResourceTest.java.template
9d347de6 (Kyrylo Shpak 2021-11-18 14:37:44 +0100  6) import io.quarkus.test.junit.QuarkusTest;

http/http-minimum-reactive/src/test/resources/NoFunctionalTest.java.template
9d347de6 (Kyrylo Shpak 2021-11-18 14:37:44 +0100  7) import io.quarkus.test.junit.QuarkusTest;

http/http-minimum/src/test/resources/HelloResourceTest.java.template
1646e552 (Jose 2021-05-21 10:46:26 +0200  6) import io.quarkus.test.junit.QuarkusTest;

http/http-minimum/src/test/resources/NoFunctionalTest.java.template
3575ad8d (Jose 2021-08-06 08:05:48 +0200  7) import io.quarkus.test.junit.QuarkusTest;

http/vertx-web-client/src/test/java/io/quarkus/qe/vertx/webclient/PureVertxHttpClientTest.java
3d0d3dcc (pablo gonzalez granados 2022-02-14 17:29:41 +0100 14) import io.quarkus.test.junit.QuarkusTest;

qute/multimodule/qute-test/src/test/java/io/quarkus/qe/qute/LocalizedMessagesTest.java
fc26839f qute/multimodule/qute-test/src/test/java/io/quarkus/qe/qute/ExampleResourceTest.java   (Fedor Dudinskiy         2022-01-11 10:49:43 +0100 11) import io.quarkus.test.junit.QuarkusTest;

security/basic/src/test/java/io/quarkus/ts/openshift/security/basic/SecurityTest.java
a4fb584d (Josef Smrcka 2022-01-04 09:59:54 +0100 11) import io.quarkus.test.junit.QuarkusTest;

security/basic/src/test/java/io/quarkus/ts/openshift/security/basic/SecurityTestProfile.java
[...]

PR #557, opened 2022-03-03
Qute module update

rsvoboda:
Qute module update

2 different message interfaces, using MessageBundles.get and injection, localized template render

Fixes https://github.com/quarkus-qe/quarkus-test-suite/issues/521


rsvoboda:
Concern: There was an issue filed in upstream that the original scenario reproduces.

Response: End user use case was invalid according to the details and use-cases described in https://quarkus.io/guides/qute-reference#localization
I believe that the reproducer attached in the GH issue is a corner case and not following the stuff described in qute guide (e.g. use of AppMessagesProviderm, empty messages in AlertMessages). Our tests should be using the intended approaches. GH issue was closed and the owner didn't come back. If you believe the issue is still valid you can communicate that on the GH issue, but we should have sane test coverage.

fedinskiy (approved):
Shouldn't these commits be merged?

rsvoboda:
Second commit is related to https://github.com/quarkus-qe/quarkus-test-suite/issues/226 that's why it was done as separate one.

PR #573, opened 2022-03-16
Use quarkus-resteasy-reactive in hibernate-fulltext-search scenario

pjgg:
Moves all resteasy method to a blocking thread model, in order to use `quarkus-resteasy-reactive` with blocking `quarkus-hibernate-orm-panache`

### DIFF
--- a/sql-db/hibernate-fulltext-search/src/main/java/io/quarkus/qe/hibernate/search/FruitResource.java
+++ b/sql-db/hibernate-fulltext-search/src/main/java/io/quarkus/qe/hibernate/search/FruitResource.java
@@ -6,13 +6,13 @@ import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 import javax.persistence.EntityManager;
 import javax.transaction.Transactional;
-import javax.validation.constraints.NotNull;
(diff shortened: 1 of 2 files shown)


sql-db/hibernate-fulltext-search/src/main/java/io/quarkus/qe/hibernate/search/FruitResource.java
```diff
@@ -6,21 +6,22 @@
 import javax.inject.Inject;
 import javax.persistence.EntityManager;
 import javax.transaction.Transactional;
-import javax.validation.constraints.NotNull;
```
rsvoboda, line 9: Why is `@NotNull` used? Some limitation of RESTEasy Reactive?

pjgg: Was removed from the previous implementation, in this scenario is not adding any value.

rsvoboda: ok, thanks

rsvoboda (approved):

PR #584, opened 2022-03-29
quarkus-cli should allow you to select which java version to target

pjgg:
### Summary

Quarkus-cli should be able to select which java version to target. This PR verifies this behavior

### Checklist:

### DIFF
--- a/quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/QuarkusCliCreateJvmApplicationIT.java
+++ b/quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/QuarkusCliCreateJvmApplicationIT.java
@@ -7,11 +7,14 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.IOException;
+import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
+import java.util.Scanner;
 import java.util.stream.Stream;
 
 import javax.inject.Inject;
@@ -43,13 +46,17 @@ public class QuarkusCliCreateJvmApplicationIT {
     static final String SMALLRYE_HEALTH_EXTENSION = "quarkus-smallrye-health";
     static final String SPRING_WEB_EXTENSION = "quarkus-spring-web";
     static final String RESTEASY_JACKSON_EXTENSION = "quarkus-resteasy-jackson";
+    static final String ROOT_FOLDER = "";
+    static final String DOCKER_FOLDER = "/src/main/docker";
+    static final String JDK_11 = "11";
+    static final String JDK_17 = "17";
+    static final String DOCKERFILE_JVM = "Dockerfile.jvm";
 
     @Inject
     static QuarkusCliClient cliClient;
 
     @Tag("QUARKUS-1071")
     @Tag("QUARKUS-1072")
-    @Tag("QUARKUS-1472")
     @Test
     public void shouldCreateApplicationOnJvm() {
         // Create application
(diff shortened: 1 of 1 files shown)


quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/QuarkusCliCreateJvmApplicationIT.java
```diff
@@ -210,4 +229,8 @@ private File getPomFileFromMavenApplication(QuarkusCliRestService app) {
                 .findFirst()
                 .orElseThrow(() -> new RuntimeException("Malformed Maven Quarkus application. Missing pom.xml"));
     }
+
+    private String truncateCurrentJavaVersion() {
```
jsmrcka, line 233: The name of this method does not correspond with what it does.

And what is the point here? To use JDK which is not the default in the system?
I don't see why. CLI's `--java` supports `11` and `17` (`11` is default), so why not just try both options, regardless to the system default?

quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/QuarkusCliCreateJvmApplicationIT.java
```diff
@@ -62,7 +64,25 @@ public void shouldCreateApplicationOnJvm() {
...
+    @Tag("QUARKUS-1472")
+    @Test
+    public void shouldCreateAnApplicationForcingJavaVersion() {
+
+        // Create application
+        QuarkusCliClient.CreateApplicationRequest args = defaults().withExtraArgs("--java=" + truncateCurrentJavaVersion());
+        QuarkusCliRestService app = cliClient.createApplication("app", args);
+
+        // Should build on Jvm
+        QuarkusCliClient.Result result = app.buildOnJvm();
```
jsmrcka, line 79: I think this won't work unless we make sure that `maven-compiler-plugin` uses the same JDK as the one provided by the `--java` param.
Is that even possible? We would need to make sure the JDK is installed, and also specify `JAVA_HOME`.

jsmrcka (changes requested):

jsmrcka (approved):

quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/QuarkusCliCreateJvmApplicationIT.java
```diff
@@ -188,13 +212,13 @@ private void assertInstalledExtensions(QuarkusCliRestService app, String... expe
...
         MavenXpp3Reader reader = new MavenXpp3Reader();
-        String javaVersion = getSystemJavaVersion();
         try {
             Model model = reader.read(new FileReader(pomFile));
-            Assertions.assertEquals(model.getProperties().get("maven.compiler.release"), javaVersion,
-                    "Unexpected Java version. Java support tool should detect host Java version");
+            Assertions.assertEquals(model.getProperties().get("maven.compiler.release"), expectedJavaVersion,
+                    "Unexpected Java version. Java support tool should detect host Java version or use " +
+                            "the provided one by --java argument");
         } catch (IOException | XmlPullParserException e) {
```
rsvoboda, line 222: Also check generated Dockerfile.jvm. This is needed for good OpenShift experience.

quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/QuarkusCliCreateJvmApplicationIT.java
```diff
@@ -188,13 +212,13 @@ private void assertInstalledExtensions(QuarkusCliRestService app, String... expe
...
-    private void assertExpectedJavaVersion(File pomFile) {
+    private void assertExpectedJavaVersion(File pomFile, String expectedJavaVersion) {
         MavenXpp3Reader reader = new MavenXpp3Reader();
-        String javaVersion = getSystemJavaVersion();
         try {
             Model model = reader.read(new FileReader(pomFile));
-            Assertions.assertEquals(model.getProperties().get("maven.compiler.release"), javaVersion,
-                    "Unexpected Java version. Java support tool should detect host Java version");
+            Assertions.assertEquals(model.getProperties().get("maven.compiler.release"), expectedJavaVersion,
+                    "Unexpected Java version. Java support tool should detect host Java version or use " +
```
rsvoboda, line 220:
```suggestion
                    "Unexpected Java version defined in maven.compiler.release property of pom.xml. Java support tool should detect host Java version or use " +
```

rsvoboda (changes requested):

rsvoboda (approved):

PR #586, opened 2022-03-31
Fix/275 fails

fedinskiy:
### Summary

Fix for two bugs both found when running the TS in OpenShift. One is related to usage of old consul config and to this bug: https://github.com/quarkusio/quarkus/issues/24277, another is due to the property, which was removed from pom.xml

### Checklist:


fedinskiy:
Since this branch fixes two bugs, it contains two commits

properties/pom.xml
```diff
@@ -32,8 +32,9 @@
             <artifactId>quarkus-config-yaml</artifactId>
         </dependency>
         <dependency>
-            <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-consul-config</artifactId>
+            <groupId>io.quarkiverse.config</groupId>
+            <artifactId>quarkus-config-consul</artifactId>
+            <version>1.0.2</version>
```
pjgg, line 37: I think that this version value should be on the main pom.xml properties

pjgg (changes requested):

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/openshift/OpenShiftPostgresqlIT.java
```diff
@@ -0,0 +1,20 @@
...
+
+import org.junit.jupiter.api.Disabled;
+
+import io.quarkus.test.scenarios.OpenShiftScenario;
+import io.quarkus.ts.reactive.PostgresqlDatabaseIT;
+
+@OpenShiftScenario
+//TODO: enable disabled methods then the issue will be fixed
+public class OpenShiftPostgresqlIT extends PostgresqlDatabaseIT {
+    @Override
```
rsvoboda, line 11: https://github.com/quarkusio/quarkus/issues/21700 seems to be fixed in 2.6.2.Final by https://github.com/quarkusio/quarkus/pull/22644

Please check if the tests are passing now and create new PR for drop of  `@Disabled("https://github.com/quarkusio/quarkus/issues/21700")`

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/PostgresqlDatabaseIT.java
```diff
@@ -7,12 +7,12 @@
 import io.quarkus.test.services.QuarkusApplication;
 
 @QuarkusScenario
-public class Postgresql13DatabaseIT extends AbstractReactiveDatabaseIT {
+public class PostgresqlDatabaseIT extends AbstractReactiveDatabaseIT {
```
rsvoboda, line 10: Why this rename? There is also `OpenShiftPostgresql10IT.java`. I would really keep the naming as is in this PR.

If you have proposal for DB related TestCases renaming, please bring it up on Monday meeting.

rsvoboda (changes requested):

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/PostgresqlDatabaseIT.java
rsvoboda: OK, looking more into the code and it probably makes sense to do it this way, it keeps consistency with other tests naming in that module.

Created trello task to discuss the generic naming topic for DB related test cases.

properties/pom.xml
rsvoboda: +1 to parametrize this in the root pom.xml

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/reactive/openshift/OpenShiftPostgresqlIT.java
fedinskiy: Will do
https://trello.com/c/9mpNtShc/1320-enable-disabled-tests-in-hibernate-reactive

rsvoboda (approved):

properties/pom.xml
fedinskiy: Since it is the only usage of this artefact, I assume that this change will bring us more harm, than good, but it's done.

rsvoboda: best practice is to have versions defined in one place - root pom.xml

PR #596, opened 2022-04-11
Disable Vertx PostgresPoolTest QuarkusTest scenario over Openshift

pjgg:
### Summary

PostgresPoolTest is a baremetal test and should be excluded from Openshift runs.

Also QuarkusTest SpringDiIT was renamed to SpringDiTest 

### Checklist:


jsmrcka (approved):
The result of the GChat discussion is to proceed with this change.
On top of that, prepare a follow-up PR, which will exclude `*Test` classes from `surefire-plugin` for `openshift` profile in POM.
Let's also reference this PR in the new PR (or vice-versa).

PR #601, opened 2022-04-13
Stork-custom loadBalancer/serviceDiscovery scenario

pjgg:
### Summary

This PR covers a basic implementation of a custom stork LB.

- [X] This change requires a documentation update

### Checklist:

### DIFF
--- a/service-discovery/stork-custom/src/main/java/io/quarkus/ts/stork/custom/SimpleLoadBalancerProvider.java
+++ b/service-discovery/stork-custom/src/main/java/io/quarkus/ts/stork/custom/SimpleLoadBalancerProvider.java
@@ -1,0 +1,15 @@
+package io.quarkus.ts.stork.custom;
+
+import io.smallrye.stork.api.LoadBalancer;
+import io.smallrye.stork.api.ServiceDiscovery;
+import io.smallrye.stork.api.config.LoadBalancerType;
+import io.smallrye.stork.spi.LoadBalancerProvider;
+
+@LoadBalancerType("simple")
+public class SimpleLoadBalancerProvider implements LoadBalancerProvider<SimpleLoadBalancerProviderConfiguration> {
+    @Override
+    public LoadBalancer createLoadBalancer(SimpleLoadBalancerProviderConfiguration config,
+            ServiceDiscovery serviceDiscovery) {
+        return new SimpleLoadBalancer(config);
+    }
+}
--- a/README.md
+++ b/README.md
@@ -662,6 +662,14 @@ Verifies Stork integration in order to provide service discovering and round-rob
 * Ping: is the main client microservice that will use `pung` and `pong` (Pong and PongReplica) services. The service 
 discovery will be done by Stork, and the request dispatching between "pong" services is going to be done by Stork load balancer. 
 
+### Service-discovery/stork-custom
+
+Verify Stork custom service discovery and load balancer implementation.
+`StorkCustomLoadBalancerIT` scenario deploys three services:
+* Pong: is a simple endpoint that returns "pong" as a string
+* PongReplica: is a "Pong service" replica, that is deployed in another physical service
+* Ping: is the main client microservice that will use `pong` and `pong-replica` services.
+
 ### `monitoring/microprofile`
 
 Verifies combined usage of MicroProfile RestClient, Fault Tolerance and OpenTracing.
--- a/pom.xml
+++ b/pom.xml
@@ -351,6 +351,7 @@
                 <module>javaee-like-getting-started</module>
                 <module>scaling</module>
                 <module>service-discovery/stork</module>
+                <module>service-discovery/stork-custom</module>
                 <module>lifecycle-application</module>
                 <module>external-applications</module>
                 <module>scheduling/quartz</module>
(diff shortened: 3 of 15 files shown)


service-discovery/stork-custom/src/test/java/io/quarkus/ts/stork/custom/StorkCustomLoadBalancerIT.java
```diff
@@ -0,0 +1,86 @@
...
+            .withProperty("stork.pong.load-balancer", "simple")
+            .withProperty("stork.pong.service-discovery", "simple")
+            .withProperty("stork.pong-replica.load-balancer", "simple")
+            .withProperty("stork.pong-replica.service-discovery", "simple")
+            .withProperty("stork.pong.service-discovery.pongServicePort", () -> "" + pongService.getPort())
+            .withProperty("stork.pong-replica.service-discovery.pongReplicaServicePort",
+                    () -> "" + pongReplicaService.getPort());
+
+    @Test
+    public void storkCustomLoadBalancer() {
```
jsmrcka, line 50: It also tests the custom service discovery, so maybe we should reflect that in the names of the method and the class.

service-discovery/stork-custom/src/test/java/io/quarkus/ts/stork/custom/StorkCustomLoadBalancerIT.java
```diff
@@ -0,0 +1,86 @@
...
+        Map<String, Integer> uniqueResp = new HashMap<>();
+        final int requestAmount = 100;
+        final int randomLbError = (requestAmount / 2) - 20;
```
jsmrcka, line 53: The `Error` in the name is a little bit confusing. It is, in fact, something like minimum hits, right? Maybe `randomLbMinHits` would be better. WDYT?

Also, just out of curiosity, is the value an estimate, or is it based on something solid?

service-discovery/stork-custom/src/main/java/io/quarkus/ts/stork/custom/SimpleLoadBalancerProvider.java
```diff
@@ -0,0 +1,17 @@
+package io.quarkus.ts.stork.custom;
+
+import io.smallrye.stork.api.LoadBalancer;
+import io.smallrye.stork.api.ServiceDiscovery;
+import io.smallrye.stork.api.config.LoadBalancerAttribute;
+import io.smallrye.stork.api.config.LoadBalancerType;
+import io.smallrye.stork.spi.LoadBalancerProvider;
+
+@LoadBalancerType("simple")
+@LoadBalancerAttribute(name = "my-attribute", description = "Attribute that alters the behavior of the LoadBalancer", required = false)
```
jsmrcka, line 10: It seems to me that this attribute is never used. It it still usefull here? It influences the contents of the respective `Configuration`, but we do not verify that anywhere. So I suggest to remove it.

jsmrcka (changes requested):

service-discovery/stork-custom/src/test/java/io/quarkus/ts/stork/custom/StorkCustomLoadBalancerIT.java
```diff
@@ -0,0 +1,86 @@
...
+
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.QuarkusApplication;
+import io.restassured.response.ValidatableResponse;
+
+import junit.framework.AssertionFailedError;
+
+@QuarkusScenario
+public class StorkCustomLoadBalancerIT {
```
jsmrcka, line 25: This scenario is part of `QUARKUS-1413` TD, right? I think we should make that clear somehow, perhaps a `@Tag`.

jsmrcka (changes requested):

service-discovery/stork-custom/src/test/java/io/quarkus/ts/stork/custom/StorkCustomLoadBalancerIT.java
pjgg: Name changes to the suggested one. 
`> just out of curiosity, is the value an estimate, or is it based on something solid?`

No is an estimation, I mean... a random generator should follow a uniform distribution

```
The probabilities for uniform distribution function are simple to calculate due to the simplicity of the function form. Therefore, there are various applications that this distribution can be used for as shown below: hypothesis testing situations, random sampling cases ...
```
So in theory a `java.util.Random` should distribute the load in a "uniform way", so if we have two instances, as is the case, we should have 50 req per node. However, after some test I saw an error of...10%, to us is not relevant... this is just a fake LB that the only thing that is probing is that the custom LB implementation is working. 

The standard round-robin implementation works [perfectly](https://github.com/quarkus-qe/quarkus-test-suite/blob/main/service-discovery/stork/src/test/java/io/quarkus/ts/stork/StorkLoadBalancerIT.java#L72)!

jsmrcka: Cool, thanks for the clarification.

jsmrcka (approved):

PR #612, opened 2022-04-22
NotFound RestEasy resources should not return any internal impl details

pjgg:
### Summary

Fails: 
```
cd http/rest-client
mvn clean verify -Dquarkus.platform.version=2.2.5.Final
```

Success: `mvn clean verify -Dall-modules -pl http/rest-client`

### Checklist:

### DIFF
--- a/http/rest-client/src/test/java/io/quarkus/ts/http/restclient/ClientBookResourceIT.java
+++ b/http/rest-client/src/test/java/io/quarkus/ts/http/restclient/ClientBookResourceIT.java
@@ -4,6 +4,8 @@ import static io.restassured.RestAssured.given;
 import static org.hamcrest.CoreMatchers.is;
 
 import org.apache.http.HttpStatus;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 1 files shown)


http/rest-client/src/main/java/io/quarkus/ts/http/restclient/RestInterface.java
```diff
@@ -24,4 +24,9 @@ public interface RestInterface {
     @Path("/json")
     @Produces(MediaType.APPLICATION_JSON)
     Book getAsJson();
+
+    @GET
+    @Path("/notFound")
```
michalvavrik, line 29: I don't see a point of exposing this endpoint when you never actually call it, please mention you are calling `given().get("/notFound")` not `given().get("/client/book/notFound")`.

michalvavrik (changes requested):
Please answer my question before I merge request. Thanks

http/rest-client/src/main/java/io/quarkus/ts/http/restclient/RestInterface.java
pjgg: You are right!. Even I don't need an endpoint over `ClientBookResource`. The issue could be reproduced with a simple notFound.

michalvavrik (approved):
Lgtm, thank you.

PR #618, opened 2022-04-26
Explicitly specify BOM GAV for created application

rsvoboda:
Explicitly specify BOM GAV for created application

As Quarkus 2.8 is available the default app is generated with the latest version, causing troubles with restarts because of extension removal when the latest bits get picked

Please select the relevant options.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Dependency update
- [ ] New scenario (non-breaking change which adds functionality)
- [ ] This change requires a documentation update

### Checklist:
- [x] Methods and classes used in PR scenarios are meaningful
- [x] Commits are well encapsulated and follow [the best practices](https://cbea.ms/git-commit/)

PR #619, opened 2022-04-26
Daily build fixs

pjgg:
### Summary

This pull request fixes the following [daily](https://github.com/quarkus-qe/quarkus-test-suite/runs/6167108941?check_suite_focus=true) build issues:
- Http/restClientReactive: fails because the event-loop was blocked. This issue requires a refactor of the current scenario. We have moved from an imperative code style to a reactive implementation. 

### Checklist:

### DIFF
--- a/http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/OsUtils.java
+++ b/http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/OsUtils.java
@@ -1,19 +1,23 @@
 package io.quarkus.ts.http.restclient.reactive;
 
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


rsvoboda:
My proposal for CLI fix is in https://github.com/quarkus-qe/quarkus-test-suite/pull/618
I have filed upstream issue https://github.com/quarkusio/quarkus/issues/25184

In `shouldAddAndRemoveExtensions` testcase the app gets restarted but the expected version of Quarkus doesn't get picked, thus the removal of `target` is sort of dirty help, but imho the right workaround is to generate the app with the expected BOM version to avoid the problematic situation.

restClientReactive is failing :/

jsmrcka (approved):

PR #624, opened 2022-04-29
Use images, which can be run on OpenShift without volumes

fedinskiy:
### Summary
Some of the images we used could not be used on OpenShift without mounted volumes. Since volumes can not be mounted via framework, we need workarounds for that.

### Checklist:


mjurc (approved):
LGTM, thanks for the fix!

PR #634, opened 2022-05-05
Fix failing tests

fedinskiy:
Put postgres data dir to folder, where root access is not required
Remove duplicated consul image

### Summary

### Checklist:


jsmrcka (commented):
So, this is a follow-up to https://github.com/quarkus-qe/quarkus-test-suite/pull/624, right?

What is the reason for defining the `PGDATA` property? Does it only involve OpenShift scenarios, or baremetal as well?

I noticed it's only defined for containers using `postgresql.13.image`, does it involve both `docker.io/library/postgres:13.6` and `registry.redhat.io/rhscl/postgresql-13-rhel7`? And `postgresql.10.image` does not need it?

What about other scenarios, that use PostgreSQL, e.g. `PostgresqlMultitenantHibernateIT`, or `PostgresqlHandlerIT`?

fedinskiy:
@jsmrcka 
Yes, this is basically the same as #624 , but for tests which were missed

Pgdata property points to folder, where postgres stores data. Default location in `/var` requires advanced rights, which cannot be provided in Openshift, but it works fine in baremetal. Up to this moment, these problems only affected Openshift and `library` images, so I didn't apply this fix for other combinations.

jsmrcka (approved):

PR #651, opened 2022-05-17
Add scenario for resteasy extension collisions

pjgg:
### Summary

When "resteasy" and "resteasy-reactive" are added as extensions an error should be thrown.
This PR covers the above scenario.

### Checklist:

### DIFF
--- a/quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/QuarkusCliCreateJvmApplicationIT.java
+++ b/quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/QuarkusCliCreateJvmApplicationIT.java
@@ -1,9 +1,12 @@
 package io.quarkus.ts.quarkus.cli;
 
+import static io.quarkus.test.bootstrap.QuarkusCliClient.CreateApplicationRequest;
+import static io.quarkus.test.bootstrap.QuarkusCliClient.Result;
 import static io.quarkus.test.bootstrap.QuarkusCliClient.CreateApplicationRequest.defaults;
 import static io.quarkus.test.utils.AwaitilityUtils.untilAsserted;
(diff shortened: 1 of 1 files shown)


quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/QuarkusCliCreateJvmApplicationIT.java
```diff
@@ -219,6 +220,18 @@ public void shouldCreateJacocoReportsFromApplicationOnJvm() {
                 "JaCoCo exec file doesn't exist");
     }
 
+    @Tag("QUARKUS-1296")
+    @Test
+    public void verifyRestEasyReactiveAndClassicResteasyCollisionUserMsg() {
+        QuarkusCliRestService app = cliClient.createApplication("dependencyCollision",
+                defaults().withExtensions("resteasy", "resteasy-reactive"));
+
+        CompletableFuture.runAsync(() -> app.start());
```
jsmrcka, line 229:
```suggestion
        app.buildOnJvm();
```
Just `build` is enough.

Also, what is the point of running it asynchronously?

jsmrcka (changes requested):

quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/QuarkusCliCreateJvmApplicationIT.java
pjgg: Basically, this app is going to fail at start-time and we want to double-check that is failing because there are a collision between `resteasy` and `resteasy-reactive`

`app.start()` should be async because we need to run the app and check the logs even if the app fails at start-time.

jsmrcka (commented):
See the discussion thread.

jsmrcka (approved):

PR #656, opened 2022-05-23
Move to Quarkus 999-SNAPSHOT including daily and PR CI

pjgg:
### Summary

Move to Quarkus 999-SNAPSHOT including daily and PR CI

Changes in testSuite:
* Reactive resteasy multipart should return a [RestResponse](https://quarkus.io/guides/resteasy-reactive#setting-other-response-properties)
* Use [confluent AVRO registry extension](https://quarkus.io/guides/kafka-schema-registry-avro#creating-the-maven-project)
* Qute multimodule scenario has duplicate properties defined in both projects
* Missing Stork `stork-configuration-generator` path on `maven-compiler-plugin`
        -  https://github.com/quarkusio/quarkus/issues/25746 
* Add support to 999-SNAPSHOT - Quarkus-cli + the following issue resolution
        - https://github.com/quarkusio/quarkus/issues/25761  

- [X] Dependency update

### Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,9 +18,9 @@ jobs:
           distribution: 'temurin'
           java-version: ${{ matrix.java }}
           check-latest: true
-      - name: Build Quarkus 2.7
+      - name: Build Quarkus main
         run: |
-          git clone https://github.com/quarkusio/quarkus.git && cd quarkus && git checkout 2.7 && ./mvnw -B -s .github/mvn-settings.xml clean install -Dquickly -Prelocations -pl devtools/cli -am
+          git clone https://github.com/quarkusio/quarkus.git && cd quarkus && ./mvnw -B -s .github/mvn-settings.xml clean install -Dquickly -Prelocations
       - name: Tar Maven Repo
         shell: bash
         run: tar -I 'pigz -9' -cf maven-repo.tgz -C ~ .m2/repository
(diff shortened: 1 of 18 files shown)


.github/workflows/ci.yml
```diff
@@ -18,9 +18,9 @@ jobs:
           distribution: 'temurin'
           java-version: ${{ matrix.java }}
           check-latest: true
-      - name: Build Quarkus 2.7
+      - name: Build Quarkus main
         run: |
-          git clone https://github.com/quarkusio/quarkus.git && cd quarkus && git checkout 2.7 && ./mvnw -B -s .github/mvn-settings.xml clean install -Dquickly -Prelocations -pl devtools/cli -am
```
jsmrcka, line 23: Not a bug, just a question:
Why did we build just a single single module `devtools/cli` (+ its dependencies)?
I would understand skipping it (`-pl !devtools/cli`), but this does not make sense to me.

jsmrcka (approved):

.github/workflows/ci.yml
pjgg: I think that was because on the previous version we were using Quarkus 2.7, so the only module that we want to compile at that time was Quarkus-cli. All the other extension was taken from MavenCentral (Quarkus 2.7)

jsmrcka: That sounds plausible.
But we build CLI in a later step (`PR - Linux - JVM build - Latest Version` -> `Build Quarkus CLI`).
So it seems to me like we could have skipped either this build entirely, or the `Build Quarkus CLI` step.

pjgg:
Windows -JVM Build fails because of an error in CI/Scripts on `qute/multimodule`. Basically, the CI script detects some changes on `qute/test` and `qute/multimodule`  but runs `qute/test` before `qute/multimodule`.

jsmrcka (changes requested):

jsmrcka (approved):

rsvoboda:
just for the record - drop of properties in qute-test is triggered by https://github.com/quarkusio/quarkus/issues/24090 which was fixed in 2.8. Had reminder to get back to it once we move to Quarkus main, but Pablo was faster :)

PR #661, opened 2022-05-27
Check, that header *contains* application/json

fedinskiy:
### Summary
We expected Content-Type header to contain only "application/json".
But it can(and now does) contain other things,
namely "charset" and "boundary" fields[1].
Therefore, we need to fix our tests.

[1] https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type

- [ ] Dependency update

### Checklist:


pjgg (approved):
LGTM

PR #655, opened 2022-05-22
Test RESTEasy Reactive issues

michalvavrik:
### Summary

Test coverage for [recently reported issues](https://github.com/quarkusio/quarkus/issues?q=is%3Aissue+author%3Asithmein+) related to a RESTEasy Reactive client and RESTEasy Reactive and the closest context.

#### RESTEasy Reactive client

- [Quarkus Issue #25326](https://github.com/quarkusio/quarkus/issues/25326) pointed out that special characters in a request path are not correctly encoded. `ReactiveRestClientIT#decodedRequestPath` checks that characters in foreign languages and reserved characters are encoded using a special subset of ASCII characters (Alphanumeric, Unereserved, Reserved). Must be tested with `mvn clean install -Dit.test=ReactiveRestClientIT#decodedRequestPath,ReactiveRestClientIT#encodedRequestPath -Dquarkus.platform.version=2.8.3.Final` as in 2.7.5.Final both encoded and decoded characters in request path are decoded by the client.

#### RESTEasy Reactive

Following tests must be run with `-Dquarkus-platform-version-2.8.3-or-higher=true`. This build time property enable `io.quarkus.ts.http.advanced.reactive.NinetyNineBottlesOfBeerResource` to be registered as a bean. `@Path#value()` validation was fixed in 2.8.3 and build fails in earlier Quarkus versions. The property must be added manually to all the relevant jobs/runs that should run this tests.

- According to [JAX-RS specification section 3.4.](https://download.oracle.com/otn-pub/jcp/jaxrs-2_0-fr-eval-spec/jsr339-jaxrs-2.0-final-spec.pdf?AuthParam=1653215184_88a21fd21b3459fc962038199675c61e) template parameters can contain dashes, underscores, ... which wasn't a case till 3.8.5. Issue was raised by [Quarkus #25258](https://github.com/quarkusio/quarkus/issues/25258). Added test `HttpAdvancedReactiveIT#uriPathTemplate` cover allowed characters as described in `javax.ws.rs.Path` Javadoc. Resource name inspired by [Donald Knuth 99 Bottles of Beer](https://en.wikipedia.org/wiki/99_Bottles_of_Beer#References_in_computer_science).

- [Media type formal parameter of the MessageBodyWriter#writeTo](https://github.com/quarkusio/quarkus/issues/25263) was always null with response type `Response`. Fixed in 2.7.6.Final. Tested by `HttpAdvancedReactiveIT#messageBodyWriter`.

[...]

PR #663, opened 2022-05-30
Enable `HttpAdvancedReactiveIT#uriPathTemplate` in Github workflows

michalvavrik:
Follow up to [PR 655](https://github.com/quarkus-qe/quarkus-test-suite/pull/655), one test must be disabled by build property in order to prevent build failure if run with Quarkus 2.7.5 and lesser.

### Summary

Class `io.quarkus.ts.http.advanced.reactive.NinetyNineBottlesOfBeerResource` breaks the build as `Path#value()` contain special characters as dash etc. [Issue 25258](https://github.com/quarkusio/quarkus/issues/25258) fixed the bug, the fix has been added to Quarkus since 2.7.6 and 2.8.3. The test resource is registered unless explicitly disabled with `-Dquarkus-platform-version-less-than-2.8.3=true`.

- [ ] Dependency update
- [x] Housekeeping

### Checklist:

### DIFF
--- a/http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/NinetyNineBottlesOfBeerResource.java
+++ b/http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/NinetyNineBottlesOfBeerResource.java
@@ -1,7 +1,7 @@
 package io.quarkus.ts.http.advanced.reactive;
 
-import static io.quarkus.ts.http.advanced.reactive.NinetyNineBottlesOfBeerResource.ENABLED_ON_QUARKUS_2_8_3_OR_HIGHER_NAME;
-import static io.quarkus.ts.http.advanced.reactive.NinetyNineBottlesOfBeerResource.ENABLED_ON_QUARKUS_2_8_3_OR_HIGHER_VAL;
+import static io.quarkus.ts.http.advanced.reactive.NinetyNineBottlesOfBeerResource.QUARKUS_PLATFORM_VERSION_LESS_THAN_2_8_3;
+import static io.quarkus.ts.http.advanced.reactive.NinetyNineBottlesOfBeerResource.QUARKUS_PLATFORM_VERSION_LESS_THAN_2_8_3_VAL;
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 
 import javax.ws.rs.GET;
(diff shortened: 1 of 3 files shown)


jsmrcka:
There were some errors/failures in JVM build in CI, which I don't think are related to the PR.
The other thing is that the build timed out (4 hours). It's building all modules, so it is expected to take long, but maybe we will need to increase the timeout.

michalvavrik:
I checked logs, the failure has nothing to do with the changes as only an empty string was added (thus no effect) as no module was specified. I added condition that if no module is specified, the property is going to be always added (as -Dall-modules is added by default.). Judging by logs so far, the change is working as expected.

I agree with the timeout thingy, but this PR simply enables one unit test, it has no impact to the runtime or other test results except for `HttpAdvancedReactiveIT#uriPathTemplate`.

michalvavrik:
I simplified the solution as I couldn't see 

> HttpAdvancedReactiveIT.JAX-RS URI path template test

before and couldn't figure out why, now the test runs and passes.

.github/workflows/ci.yml
```diff
@@ -126,7 +126,9 @@ jobs:
             MODULES_MAVEN_PARAM="-pl ${MODULES_ARG}"
           fi
 
-          mvn -fae -V -B -s .github/mvn-settings.xml clean verify -Dall-modules -Dinclude.quarkus-cli-tests -Dts.quarkus.cli.cmd="${PWD}/quarkus-dev-cli" $MODULES_MAVEN_PARAM -am
+          mvn -fae -V -B -s .github/mvn-settings.xml clean verify -Dquarkus-platform-version-2.8.3-or-higher="true" \
```
jsmrcka, line 129: I've been thinking about this and it seems to me that it might be easier to invert the condition: enable the bean and related tests unless Quarkus version is <2.8.3.

It would require changes in `NinetyNineBottlesOfBeerResource`:
```java
@RegisterForReflection
@UnlessBuildProperty(name = QUARKUS_PLATFORM_VERSION_LESS_THAN_2_8_3, stringValue = QUARKUS_PLATFORM_VERSION_LESS_THAN_2_8_3_VAL, enableIfMissing = true)
@Path(NinetyNineBottlesOfBeerResource.PATH)
public class NinetyNineBottlesOfBeerResource {

    public static final String QUARKUS_PLATFORM_VERSION_LESS_THAN_2_8_3 = "quarkus-platform-version-less-than-2.8.3";
    public static final String QUARKUS_PLATFORM_VERSION_LESS_THAN_2_8_3_VAL = "true";
    ...
}
```
and at `HttpAdvancedReactiveIT#uriPathTemplate`:
```java
    @DisabledIfSystemProperty(matches = QUARKUS_PLATFORM_VERSION_LESS_THAN_2_8_3_VAL, named = QUARKUS_PLATFORM_VERSION_LESS_THAN_2_8_3, disabledReason = TEST_REQUIRES_AT_LEAST_2_8_3)
    @DisplayName("JAX-RS URI path template test")
    @Test
    public void uriPathTemplate() {
        ...
    }
```

That way, we don't need to add any properties anywhere:
- `2.7` branch (and anything older) does not containt the bean/tests.
- `main` (and any future branches) will have the bean/tests enabled by default.

And in case we forget to add the property somewhere (if we backport the tests, or e.g. when running the TS with `-Dquarkus.platform.version=2.8.1.Final`), we will at least get a visible buildtime error.

WDYT?

michalvavrik: Absolutely, that's much better. Done, thanks for the suggestion.

jsmrcka (changes requested):

http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/NinetyNineBottlesOfBeerResource.java
```diff
@@ -1,29 +1,30 @@
...
 
-import io.quarkus.arc.properties.IfBuildProperty;
+import io.quarkus.arc.properties.UnlessBuildProperty;
 import io.quarkus.runtime.annotations.RegisterForReflection;
 import io.smallrye.mutiny.Uni;
 
 /**
  * {@link Path#value()} set in this resource contains characters that were causing the build time
- * validation failure. The issue was resolved in 2.8.3. with https://github.com/quarkusio/quarkus/issues/25258.
+ * validation failure. The issue was resolved in 2.8.3 and backported to 2.7.6 with
```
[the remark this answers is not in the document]
michalvavrik: True, fixed, thank you.

jsmrcka (approved):

PR #681, opened 2022-06-02
Test MessageBodyWriter receives right accepted Media Type with Response

michalvavrik:
### Summary

Verifies [Quarkus Issue 22119](https://github.com/quarkusio/quarkus/issues/22119). According to JAX-RS standard, if request specifies accepted Media type, the type should be passed to `javax.ws.rs.ext.MessageBodyWriter`, however for use case when `javax.ws.rs.core.Response` is returned, that was only fixed in Quarkus 2.8.1 and backported to 2.7.6. Previously, if multiple media types were defined in `@Produces`, the first one was taken. This test verifies that a value of `Accept` HTTP header is same as the media type passed to `MessageBodyWriter` when endpoint response type is `Response`.

#### Reproducer

`mvn clean verify -Dit.test=HttpAdvancedReactiveIT#testMediaTypePassedToMessageBodyWriter -Dquarkus.platform.version=2.8.0.Final -Dquarkus-platform-version-less-than-2.8.3=true` with commented out `@DisabledOnQuarkusVersion`.

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/MultipleResponseSerializersResource.java
+++ b/http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/MultipleResponseSerializersResource.java
@@ -1,0 +1,106 @@
+package io.quarkus.ts.http.advanced.reactive;
+
+import static io.quarkus.ts.http.advanced.reactive.MultipleResponseSerializersResource.MULTIPLE_RESPONSE_SERIALIZERS_PATH;
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON_TYPE;
+import static javax.ws.rs.core.MediaType.APPLICATION_OCTET_STREAM;
+import static javax.ws.rs.core.MediaType.APPLICATION_OCTET_STREAM_TYPE;
+import static javax.ws.rs.core.MediaType.TEXT_HTML;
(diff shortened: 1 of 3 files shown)


http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/MultipleResponseSerializersResource.java
```diff
@@ -0,0 +1,108 @@
...
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.MessageBodyWriter;
+import javax.ws.rs.ext.Provider;
+
+import io.quarkus.runtime.annotations.RegisterForReflection;
+import io.vertx.core.http.HttpServerRequest;
+
+@RegisterForReflection
```
jsmrcka, line 32:
```suggestion
```
We don't need this. All classes here are CDI beans.

jsmrcka (changes requested):

http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/MultipleResponseSerializersResource.java
michalvavrik: You're right, fixed, thank you

PR #702, opened 2022-06-09
Fix OpenShiftGraphQLTelemetryIT - filter Jaeger traces by operation name

michalvavrik:
### Summary

Currently traces are filtered by `operationName` which is not valid filter parameter and operations are not filtered by this. That causes problem in OpenShift where there are other calls (other operations with different names, e.g. `HTTP GET` - probably probe check). Problem can be easily reproduced locally on `GraphQLTelemetryIT` when you call GET to `/` couple of times, which leads to test failure.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


michalvavrik:
@pjgg I just realized this PR is only solution for 999-SNAPSHOT (you can reproduce the bug running `mvn clean verify -f http/graphql-telemetry -Dopenshift` on `main`), but in 2.7.6.Final traces are still not filtered according to operation name. I'll look into it tomorrow and let you know.

pjgg (approved):

pjgg:
LGTM

michalvavrik:
I re-checked, fix works in 2.7.6.Final too.

Issue #707, opened 2022-06-13
LargeFileHandlingIT#uploadMultipart fails wtih OutOfMemory error


fedinskiy:

fedinskiy:
https://github.com/quarkus-qe/quarkus-test-suite/runs/6832504714?check_suite_focus=true

fedinskiy:
Upstream bug: 
https://github.com/quarkusio/quarkus/issues/26187

michalvavrik:
upstream issue is closed and test is now disabled for other reasons https://github.com/quarkus-qe/quarkus-test-suite/pull/854 and we have got https://github.com/quarkus-qe/quarkus-test-framework/issues/584 to track OS Windows issue

PR #717, opened 2022-06-15
CI testing: Disable unused tests

fedinskiy:


rsvoboda:
What's the intention of this PR? Just experiments? There is no description.

fedinskiy:
@rsvoboda yes, since https://github.com/quarkus-qe/quarkus-test-suite/issues/707 reproduces reliably only on GH, I decided to run some of my experiments here.

PR #747, opened 2022-07-15
Revert helm CI setup

pjgg:
We don't have enough details from the product side about which version of Helm is going to be supported. So, let's revet these changes and wait until are clarified.


rsvoboda (commented):
This is jsoup version bump

rsvoboda:
And TBH we can imho keep helm client stuff in GH actions because upstream should target the latest helm version

pjgg:
No, we don't know which Helm version we are gonna support. Also, Helm makes sense when you are running your CI test over K8s/Ocp. This GH setup cloud makes sense if you are doing some kind of verification, but we don't know yet what needs to be covered so, let's remove it until having more news about helm requirements.

pjgg:
> This is jsoup version bump

Ok, let's me do a `git revert` instead of a `git reset` in order to avoid confusion.

rsvoboda (approved):

PR #776, opened 2022-07-28
SmallRye GraphQL - test route context in DEV mode is terminated

michalvavrik:
### Summary

Verifies: https://github.com/quarkusio/quarkus/issues/26748

In the Quarkus Smallrye GraphQL 2.10.2.Final and 2.10.1.Final, a route context had not been properly terminated. If BE received 3+ requests then 3rd (and every other) request were handled with the first request's route context (thus identity, headers etc. were incorrect). I could only reproduce this behavior with `mvn quarkus:dev` and `@QuarkusTest`, my best guess is that's due to [this hack](https://github.com/quarkusio/quarkus/blob/main/extensions/smallrye-graphql/deployment/src/main/java/io/quarkus/smallrye/graphql/deployment/SmallRyeGraphQLProcessor.java#L303), but I can be easily wrong!

Issue can be reproduced with `mvn clean verify -Dit.test=DevModeJwtSecurityGraphQLIT -Dquarkus.platform.version=2.10.2.Final`.

The issue verification is done by the `DevModeJwtSecurityLIT`. In addition, this PR also runs `testRouteContextTermination` for `cookie`, `Oauth2` and `OpenShift` scenarios as we don't have a basic JWT scenario with GraphQL yet. `BaseJwtSecurityIT` is newly run in DEV mode which  increase the coverage (+29 tests).

I also refactored `smallrye.jwt.verify.publickey.location` to the `mp.jwt.verify.publickey.location` and `smallrye.jwt.verify.issuer` to the `mp.jwt.verify.issuer` as former is going to be removed in the next version.

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/security/jwt/src/test/java/io/quarkus/ts/security/jwt/BaseJwtSecurityIT.java
+++ b/security/jwt/src/test/java/io/quarkus/ts/security/jwt/BaseJwtSecurityIT.java
@@ -8,8 +8,11 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import java.security.KeyPairGenerator;
 import java.security.NoSuchAlgorithmException;
 import java.util.Objects;
+import java.util.stream.Stream;
 
 import org.apache.http.HttpStatus;
+import org.hamcrest.CoreMatchers;
+import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 
 import io.restassured.http.ContentType;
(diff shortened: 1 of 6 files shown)


rsvoboda:
Pablo had quite a few reviews this week, please set reviewer to @jsmrcka or @fedinskiy to spread the load a bit in the team

security/jwt/src/test/java/io/quarkus/ts/security/jwt/BaseJwtSecurityIT.java
```diff
@@ -13,9 +13,8 @@
 import org.junit.jupiter.api.Test;
 
 import io.restassured.http.ContentType;
-import io.restassured.specification.RequestSpecification;
 
-public abstract class BaseJwtSecurityIT {
+public abstract class BaseJwtSecurityIT extends BaseJwtSecurityGraphQLIT {
```
jsmrcka, line 17: I understand what we want to do here (thanks to the excellent PR description :+1:), but I have 2 objections:
- The hierarchy of the test classes is a bit confusing. There is a GraphQL-related class as a parent to a more generic test class, weird placement of an abstract method, a parent calling a static method of a child.
- We perform only one dev mode test, tailor made to the GraphQL scenario. Why not run all the existing tests in dev mode?

My suggestion is:
- Get rid of `BaseJwtSecurityGraphQLIT`.
  - Move `protected abstract RequestSpecification givenWithToken` back to `BaseJwtSecurityIT` together with both new methods. I.e. make `testRouteContextTermination` part of the `BaseJwtSecurityIT` suite.
- Make `DevModeJwtSecurityGraphQLIT` extend `BaseJwtSecurityIT`.
  - And also maybe replace `RestAssured.given().auth().oauth2(token)` call with `app.given().auth().oauth2(token)` to make it more obvious. :slightly_smiling_face:

That way, the test structure will be more readable and we increase the coverage (+29 tests).
`DevModeJwtSecurityGraphQLIT` is not `fips-incompatible`, so we don't have to worry about that (I tested it on FIPS machine).

What do you think?

michalvavrik: I agree with suggestions, thank you. Done.

jsmrcka (approved):

PR #777, opened 2022-07-29
Keycloak 18 requires uma_protection scope on http/http_advanced

pjgg:
### Summary

From now I didn't see this problem in other scenarios so I am not going to propagate this scope to all the realms/scenarios. Maybe is need it on `http/http-advanced` ( and reactive) because `quarkus-keycloak-authorization` is making some internal query that requires `uma_protection` (this is only an issue with Keycloak 18 +)
 
### Checklist:


michalvavrik (approved):
To be honest I can't reproduce it locally anymore, but I think we should at this just to be sure. Thanks

PR #825, opened 2022-09-08
Test Kubernetes Config sources prioritization

michalvavrik:
### Summary

In addition to existing test coverage of the Quarkus Kubernetes Config, we newly test that:

1. Properties retrieved from ConfigMaps and Secrets created from an `application.yml` file.
2. Priority of obtained properties:
   1. Properties defined in Secrets have higher priority than same properties defined in Config Maps.
   2. When multiple ConfigMaps (or Secrets) are used, ConfigMaps (or Secrets) defined later in the list have a higher priority that ConfigMaps defined earlier in the list.
   3. The properties obtained from the ConfigMaps and Secrets have a higher priority than (i.e. they override) any properties of the same name that are found in application.properties (or the YAML equivalents), but they have lower priority than properties set via Environment Variables or Java System Properties.
3. If property `quarkus.kubernetes-config.fail-on-missing-config` set to true, the application will not start if any of the configured config sources cannot be located.

I left out testing of the ConfigMaps/Secrets placed in another namespace than the app is running in as our OC user can't grant `ClusterRole`s.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/config/src/main/resources/application.properties
+++ b/config/src/main/resources/application.properties
@@ -1,0 +1,4 @@
+hello.preamble=preamble-never-used-value
+hello.epilogue=epilogue-never-used-value
+side-note=side-note-never-used-value
+hello.side-note=${side-note}
--- a/config/src/test/resources/configmap-yml.yaml
+++ b/config/src/test/resources/configmap-yml.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: app-config-yml
+data:
+  application.yml: |
+    hello:
+      message: message configmap-yml
+      epilogue: epilogue configmap-yml
+      preamble: preamble configmap-yml
--- a/config/src/main/java/io/quarkus/ts/configmap/api/server/HelloConfigProperties.java
+++ b/config/src/main/java/io/quarkus/ts/configmap/api/server/HelloConfigProperties.java
@@ -1,0 +1,38 @@
+package io.quarkus.ts.configmap.api.server;
+
+import java.util.Optional;
+
+import org.eclipse.microprofile.config.inject.ConfigProperties;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+@ConfigProperties(prefix = "hello")
+public class HelloConfigProperties {
+
+    @ConfigProperty(name = "message")
+    Optional<String> message;
+
+    @ConfigProperty(name = "preamble")
+    String preamble;
+
+    @ConfigProperty(name = "epilogue")
+    String epilogue;
+
+    @ConfigProperty(name = "side-note")
+    String sideNote;
+
+    public Optional<String> getMessage() {
+        return message;
+    }
+
+    public String getPreamble() {
+        return preamble;
+    }
+
+    public String getEpilogue() {
+        return epilogue;
+    }
+
+    public String getSideNote() {
+        return sideNote;
+    }
+}
--- a/config/src/main/java/io/quarkus/ts/configmap/api/server/HelloResource.java
+++ b/config/src/main/java/io/quarkus/ts/configmap/api/server/HelloResource.java
@@ -1,7 +1,5 @@
 package io.quarkus.ts.configmap.api.server;
 
-import java.util.Optional;
-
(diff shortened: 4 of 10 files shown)


config/src/test/resources/configmap-yml.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: app-config-yml
+data:
+  application.yml: |
+    hello:
+      message: configmap-yml
```
fedinskiy, line 8: You are using the same value for `message`, `epilogue` and `preamble` fields. Since you are checking the values of these fields later, it is better if they are different.

config/src/main/resources/application.properties
```diff
@@ -0,0 +1,4 @@
+hello.preamble=application properties
+hello.epilogue=application properties
```
fedinskiy, line 2: Are these values used? Also, all fields hold the same value.

fedinskiy (changes requested):

config/src/test/resources/configmap-yml.yaml
michalvavrik: Good point, they are prefixed with a field name now, thanks.

config/src/main/resources/application.properties
michalvavrik: The point is to be sure they are not used -> they are overridden by other config sources (OC ones). Also setting defaults means existing tests are not affected by these changes.

michalvavrik:
OC tests blocked by https://github.com/quarkus-qe/quarkus-test-framework/pull/556

michalvavrik:
I run all tests against OC 4.6 and 4.11 and it's fine, but we can wait for the fix in TFW

config/src/main/resources/application.properties
fedinskiy: Ok, then replace the content to something like "preamble-never-used-value" so it could be easily understood just by looking at the file

michalvavrik: sure, I'll do it.

michalvavrik: @fedinskiy any particular reason for dashes? I'll need to unify the values of all the properties (in secrets/configmaps) so I wonder if you want dashes everywhere or just here?

fedinskiy (approved):
But I suggest to make sure OCP jobs are successful. Also some files doesn't have newlines at the end, could you add them?

config/src/main/resources/application.properties
michalvavrik: Done.

michalvavrik:
All the files ends with a new line now and I'll wait for the TFW release -> rebase -> run OCP tests and check results. Thank you for the rview

michalvavrik:
blocked by https://github.com/quarkus-qe/quarkus-test-framework/pull/558

pjgg:
@michalvavrik could you do a rebase/resolve conflicts?

michalvavrik:
@pjgg I'm waiting for https://github.com/quarkus-qe/quarkus-test-suite/pull/838 as I already know that's required. I am hoping to do the rebase today

rsvoboda:
rebase pls

michalvavrik:
@rsvoboda I was fighting the issues you run into when re-base this PR (and not just there, you can see it in dailys probably), it's caused by https://github.com/quarkusio/quarkus/pull/27733. I try to create reproducer that does not include our TS, but so far I failed. Once I have reproducer, I'll create upstream issue.

I'll will rebase this today anyway, as there is a simple workaround.

michalvavrik:
blocked by https://github.com/quarkus-qe/quarkus-test-framework/pull/573

mjurc:
Run tests

michalvavrik:
run tests

michalvavrik:
There is one failing test in `quarkus-main-rhel8-jdk17-openshift-ts-native`, `OpenShiftFileSystemConfigMapConfigIT`. The test isn't affected in any way by this PR. I just debugged it and it's flaky test - once it fails, once it succeeds and so on. I'll keep debugging it tonight and address the issue in a separate PR.

PR #826, opened 2022-09-09
Provide info about Maven

rsvoboda:
Provide info about Maven

Example:
```
Mac OS X 12.5.1

    Architecture
        x86_64
    JVM
        OpenJDK 64-Bit Server VM GraalVM CE 21.3.0 (build 11.0.13+7-jvmci-21.3-b05, mixed mode, sharing)
        11
    MVN
        Apache Maven 3.8.1 (05c21c65bdfed0f71a2f2ada8b84da59348c4c5d)
        Maven home: /Users/rsvoboda/.sdkman/candidates/maven/current
        Java version: 11.0.13, vendor: GraalVM Community, runtime: /Users/rsvoboda/.sdkman/candidates/java/21.3.0.r11-grl
        Default locale: en_GB, platform encoding: UTF-8
        OS name: "mac os x", version: "12.5.1", arch: "x86_64", family: "mac"

```

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/env-info/src/main/java/io/quarkus/ts/envinfo/OsJvmEnvInfoLogger.java
+++ b/env-info/src/main/java/io/quarkus/ts/envinfo/OsJvmEnvInfoLogger.java
@@ -6,8 +6,11 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.Collections;
+import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 public class OsJvmEnvInfoLogger {
     private static final Path REDHAT_RELEASE_PATH = Path.of("/etc/redhat-release");
@@ -24,6 +27,11 @@ public class OsJvmEnvInfoLogger {
             + "        <li>%s</li>\n"
             + "      </ul>\n"
             + "    </li>\n"
+            + "    <li>MVN\n"
+            + "      <ul>\n"
+            + "        <li>%s</li>\n"
+            + "      </ul>\n"
+            + "    </li>\n"
             + "  </ul>\n"
             + "</li>\n";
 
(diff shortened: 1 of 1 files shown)


env-info/src/main/java/io/quarkus/ts/envinfo/OsJvmEnvInfoLogger.java
```diff
@@ -53,4 +63,15 @@ private static String getJavaVersionCmdOutput() throws InterruptedException, IOE
         return new BufferedReader(new InputStreamReader(pr.getInputStream())).lines().reduce((first, second) -> second)
                 .orElse(null);
     }
+
+    private static List<String> getMvnVersionCmdOutput() {
+        final String cmd = System.getProperty("maven.home") + File.separator + "bin" + File.separator + "mvn --version";
+        try {
```
jsmrcka, line 69: I would prefer letting the exceptions show themselves.

jsmrcka (changes requested):

env-info/src/main/java/io/quarkus/ts/envinfo/OsJvmEnvInfoLogger.java
rsvoboda: I did this on purpose, the exception breaks the run of mvn command / whole TS.

jsmrcka: It is IMO correct to break the whole run when the mvn command fails.

rsvoboda: > It is IMO correct to break the whole run when the mvn command fails.
 
Definitely not, the main goal is to run the TS. 
Getting env details is not the main goal of TS execution.
TS shouldn't stop because the secondary task didn't finish properly.

The command fails only if `System.getProperty("maven.home")` gives undefined / non-relevant value.

rsvoboda: "java.home" property is defined thanks to JDK so it's fine to fail if `getJavaVersionCmdOutput` doesn't succeed.

"maven.home" property is not defined thanks to JDK and I didn't find any guarantee for the availability. So it's better to return empty result than stop the world.

jsmrcka: Ok. I think it would be better to have the environment set up in such a way that would guarantee the consistent availability of Maven (after all, we do always run a maven build), but let's not block this PR because of it.

jsmrcka (approved):

PR #868, opened 2022-10-13
Fix flaky Kafka tests

michalvavrik:
### Summary

Fixes

- `testAlertMonitorEventStream` in `kafka-confluent-avro-reactive-messaging`. This test scheduled generation of events to the midnight (when also our CI runs) and therefore test results sometimes (depending on when CI started) had 100 more events than expected. I think the intention of scheduling this to midnight was to disable generation of events in certain tests.
- ```ScheduledInvoker_generate_c9e4876862861cd03eb297791b102d51bdc3ba72, cron=* * * * * ?]: java.util.concurrent.CompletionException: java.lang.RuntimeException: Error injecting org.eclipse.microprofile.reactive.messaging.Emitter<io.quarkus.ts.messaging.kafka.StockPrice> io.quarkus.ts.messaging.strimzi.kafka.reactive.KStockPriceProducer.emitter``` We tried to emit events before `source-stock-price` channel was ready, which lead to the injection exception (basically, when emitter is invoked, exception was thrown due to channel state). Now we wait till app is ready, to my observations, this seems to be enough.
- Stability - we don't run scheduled job concurrently from now on (see `concurrentExecution = SKIP` on `@Scheduled`), as although `org.eclipse.microprofile.reactive.messaging.Emitter#send(T)` is thread safe, our tests are expecting batches that have at least _xyz_ size, they expect to consume one batch at time, not one and half. Let's imagine you consume more events one time and on second call, you have less than expected number (because in first run you consumed more, just because execution was a little bit faster). I don't think it was an issue (AFAIK we try to fetch all in one batch, couple months back we didn't), but it's bad practice, tests should be idempotent.

Works around

- https://github.com/quarkusio/quarkus/issues/28567 by stopping generation after 4 iterations (we don't need that much anyway...). The intention is to avoid generation of events when the endpoint is returning response. I added FIXME so we can get rid of this as soon as the issue is fixed.

I'd like to keep in place logging that I added previously (for debugging) as I can't reproduce none of these things, so all above are just my best guesses.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/messaging/kafka-confluent-avro-reactive-messaging/src/main/java/io/quarkus/ts/messaging/kafka/KStockPriceProducer.java
+++ b/messaging/kafka-confluent-avro-reactive-messaging/src/main/java/io/quarkus/ts/messaging/kafka/KStockPriceProducer.java
@@ -1,6 +1,9 @@
 package io.quarkus.ts.messaging.kafka;
 
+import static io.quarkus.scheduler.Scheduled.ConcurrentExecution.SKIP;
+
 import java.util.Random;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.BiConsumer;
 import java.util.stream.IntStream;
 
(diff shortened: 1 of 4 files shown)


rsvoboda:
run tests

pjgg (approved):
LGTM

PR #883, opened 2022-10-26
Remove Quarkus-helm dependency from main pom

pjgg:
### Summary

Currently, we are not supporting Quarkus/helm due to upstream extension instability. 

- [X] Dependency update

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -31,7 +31,8 @@
         <jjwt.version>0.11.5</jjwt.version>
         <jsoup.version>1.15.3</jsoup.version>
         <camel-quarkus-bom.version>2.13.0</camel-quarkus-bom.version>
-        <quarkus-helm.version>0.1.1</quarkus-helm.version>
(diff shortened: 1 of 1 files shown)


fedinskiy:
Given that camelet is also not compatible with 999-SNAPSHOT[1], I suggest to create a separate maven profile for these two modules and enable it only for released versions. WDYT?

[1] https://github.com/quarkusio/quarkus/issues/28007

pjgg:
Quarkus-helm is not going to be released on 2.13 so doesn't make sense in this case.

fedinskiy (approved):

fedinskiy:
Failures are peculiar (especially this one: `Caused by: com.github.dockerjava.api.exception.DockerClientException: Could not pull image: failed to register layer: ApplyLayer exit status 1 stdout:  stderr: write /opt/ibm/db2/V11.5/lib64/libdb2osse.so.1: no space left on device`), but unrelated.

PR #886, opened 2022-10-27
Add support for marking configs as secret

fedinskiy:
### Summary

Issue: https://issues.redhat.com/browse/QUARKUS-2479
Test plans: https://github.com/quarkus-qe/quarkus-test-plans/pull/105

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/config/src/main/java/io/quarkus/ts/configmap/api/server/secrets/BuiltConfigResource.java
+++ b/config/src/main/java/io/quarkus/ts/configmap/api/server/secrets/BuiltConfigResource.java
@@ -1,0 +1,25 @@
+package io.quarkus.ts.configmap.api.server.secrets;
+
+import javax.ws.rs.Path;
+
+import io.smallrye.config.SmallRyeConfig;
+import io.smallrye.config.SmallRyeConfigBuilder;
+
+@Path("/built")
+public class BuiltConfigResource extends SecretResource {
+
+    private final SmallRyeConfig config;
+
+    public BuiltConfigResource() {
+        config = new SmallRyeConfigBuilder()
+                .addDefaultSources()
+                .addDefaultInterceptors()
+                .withSecretKeys("secret.ip")
+                .build();
+    }
+
+    @Override
+    public String getProperty(String key) {
+        return config.getRawValue(key);
+    }
+}
--- a/config/src/main/java/io/quarkus/ts/configmap/api/server/secrets/ConfigInterceptorFactory.java
+++ b/config/src/main/java/io/quarkus/ts/configmap/api/server/secrets/ConfigInterceptorFactory.java
@@ -1,0 +1,15 @@
+package io.quarkus.ts.configmap.api.server.secrets;
+
+import java.util.Set;
+
+import io.smallrye.config.ConfigSourceInterceptor;
+import io.smallrye.config.ConfigSourceInterceptorContext;
+import io.smallrye.config.ConfigSourceInterceptorFactory;
+import io.smallrye.config.SecretKeysConfigSourceInterceptor;
+
+public class ConfigInterceptorFactory implements ConfigSourceInterceptorFactory {
+    @Override
+    public ConfigSourceInterceptor getInterceptor(ConfigSourceInterceptorContext context) {
+        return new SecretKeysConfigSourceInterceptor(Set.of("secret.password"));
+    }
+}
--- a/config/src/main/java/io/quarkus/ts/configmap/api/server/secrets/InjectedConfigResource.java
+++ b/config/src/main/java/io/quarkus/ts/configmap/api/server/secrets/InjectedConfigResource.java
@@ -1,0 +1,18 @@
+package io.quarkus.ts.configmap.api.server.secrets;
+
+import javax.inject.Inject;
+import javax.ws.rs.Path;
+
+import io.smallrye.config.SmallRyeConfig;
+
+@Path("/injected")
+public class InjectedConfigResource extends SecretResource {
+
+    @Inject
+    SmallRyeConfig config;
+
+    @Override
+    public String getProperty(String key) {
+        return config.getRawValue(key);
+    }
+}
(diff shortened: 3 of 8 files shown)


config/src/main/java/io/quarkus/ts/configmap/api/server/HelloResource.java
```diff
@@ -9,13 +9,23 @@
 import javax.ws.rs.core.Response;
 
 import org.eclipse.microprofile.config.inject.ConfigProperties;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+import io.smallrye.config.SmallRyeConfig;
+import io.smallrye.config.SmallRyeConfigBuilder;
```
michalvavrik, line 15: Where are 2 imports used, please?

fedinskiy: Good catch, they are used nowhere :)

removed

config/src/main/java/io/quarkus/ts/configmap/api/server/SmallRyeResource.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Path("/secret")
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    public Response getSecret() {
+        Response.ResponseBuilder result;
+        try {
+            result = Response.ok(config.getRawValue("secret.password"));
+        } catch (SecurityException ex) {
+            result = Response.serverError().entity(ex.getMessage());
```
michalvavrik, line 33: You could return DTO that contains both ex. message and return of
```Java
String secretValue = SecretKeys.doUnlocked(() -> {
    config.getValue("secret", String.class);
});

```
so we know it actually works when config is build manually in constructor.

config/src/main/java/io/quarkus/ts/configmap/api/server/HelloResource.java
```diff
@@ -34,4 +41,17 @@ public HelloConfigProperties getProperties() {
         return configProperties;
     }
 
+    @Path("/secret")
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    public String getSecret() {
+        return secret;
+    }
+
```
michalvavrik, line 50: Please also add there 
```
 @Inject
  SmallRyeConfig config;
```
and endpoint that access the value e.g. `config.getConfigValue("secrets.my.secret")`, I know ATM it's tested upstream, but in my eyes, it's only actual value of this feature. (just my opinion)

michalvavrik (changes requested):
So, here is my issue:

- The tests you added basically allows everyone to access the secret (inject secrets via `@ConfigProperty`). How is it secured if I can use `@ConfigProperty` or `ConfigProvider.getConfig().getValue("secret.password")`? It doesn't provide any extra security for common use case! Way I understand docs https://smallrye.io/smallrye-config/2.11.1/config/secret-keys/ is that it should be harder, like I should be required to use 
- ```Java
String secretValue = SecretKeys.doUnlocked(() -> {
    config.getValue("secret", String.class);
});
```

All secrets are unlocked for injection. So literally only value of https://github.com/quarkusio/quarkus/pull/26509 is that if you access secrets via `config.getConfigValue("secrets.my.secret")` :-(
 - Do you understand meaning of ConfigSourceInterceptorFactory https://github.com/quarkusio/quarkus/pull/26509/files? I don't understand why it's there. Are you sure we don't need it in order to actually test secret failure?

File: config/src/main/java/io/quarkus/ts/configmap/api/server/SmallRyeResource.java (hunk 2 above)
Comment (fedinskiy) [reply config/src/main/java/io/quarkus/ts/configmap/api/server/SmallRyeResource.java:34]: Added a separate test for this case

File: config/src/main/java/io/quarkus/ts/configmap/api/server/HelloResource.java (hunk 3 above)
Comment (fedinskiy) [reply config/src/main/java/io/quarkus/ts/configmap/api/server/HelloResource.java:50]: This config will not have `.withSecretKeys("secret.password")` in its builder and therefore will not have any keys marked as secret. I guess, we can create a separate bean for that, but  for now the idea is to verify, that:
1) We can inject properties and use them as is
2) We can build our own config and mark some properties as secret

Comment (michalvavrik) [reply config/src/main/java/io/quarkus/ts/configmap/api/server/HelloResource.java:50]: Ah, good, I wasn't sure if that's what is happening. But that basically means you are not testing primary way user is going to do, right? How many users are going to build config manually? They are going to use injected config, so we need an interceptor and add some secret (f.e. under another name).

Comment (michalvavrik) [reply config/src/main/java/io/quarkus/ts/configmap/api/server/HelloResource.java:50]: Btw. that's not very user friendly. what you can't provide secrets without interceptor, e.g. config property that says "secure xyz"

**Comment by fedinskiy** [comment]
@michalvavrik  regarding your concerns:

- This feature is WIP in upstream so now we just check that this API is integrated into Quarkus. And regarding protection in general — even SmallRye docs do not recommend it as the only line of defence.
- To my understanding, since Interceptor factory is in module `integration-tests/` and inside `src/test` folder, it is a part of internal API and is used for testing only.

**Comment by michalvavrik** [comment]
Look @fedinskiy , you are an expert here :-), I'm just not sure I believe you :-D. The interceptor looks like the only way for users to actually use the feature in production and it's not part of testing, see imports
```Java
import io.smallrye.config.ConfigSourceInterceptorFactory;
import io.smallrye.config.SecretKeysConfigSourceInterceptor
```
Just please prove me wrong.

config/src/main/java/io/quarkus/ts/configmap/api/server/HelloResource.java
fedinskiy: Hmm, I wanted to argue, but it looks like[1], that config interceptors are supported in Quarkus. Good idea, will add this to tests.

[1] https://quarkus.io/guides/config-extending-support#config-interceptors

michalvavrik: Thank you

michalvavrik (approved):
lgtm

PR #948, opened 2022-12-14
Set OIDC trust-store and key-store providers

pjgg:
### Summary

RE-enable a property that was commented due to an issue that is already fixed.

### Checklist:

### DIFF
--- a/security/oidc-client-mutual-tls/src/main/resources/application.properties
+++ b/security/oidc-client-mutual-tls/src/main/resources/application.properties
@@ -10,8 +10,8 @@ quarkus.oidc.tls.verification=certificate-validation
 quarkus.oidc.tls.key-store-password=${ks-pwd}
 quarkus.oidc.tls.trust-store-password=${ks-pwd}
 
-# TODO https://github.com/quarkusio/quarkus/issues/25972
-#quarkus.oidc.tls.trust-store-provider=SunRsaSign,SunJCE
-#quarkus.oidc.tls.key-store-provider=SunRsaSign,SunJCE
+# Motivation: https://github.com/quarkusio/quarkus/issues/25972
+quarkus.oidc.tls.trust-store-provider=SunRsaSign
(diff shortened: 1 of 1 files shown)


security/oidc-client-mutual-tls/src/main/resources/application.properties
```diff
@@ -10,8 +10,8 @@ quarkus.oidc.tls.verification=certificate-validation
 quarkus.oidc.tls.key-store-password=${ks-pwd}
 quarkus.oidc.tls.trust-store-password=${ks-pwd}
 
-# TODO https://github.com/quarkusio/quarkus/issues/25972
-#quarkus.oidc.tls.trust-store-provider=SunRsaSign,SunJCE
-#quarkus.oidc.tls.key-store-provider=SunRsaSign,SunJCE
+# Motivation: https://github.com/quarkusio/quarkus/issues/25972
+quarkus.oidc.tls.trust-store-provider=SunRsaSign
```
jsmrcka, line 14: I'm a little bit lost here, so let me ask:
- Does this actually test the given value? Because it can still accept any value.
- Is there a way to test the property value, e.g. use a keystore type which not well-known to the implementation => cannot be defaulted to?
  - If not, does it make sense to keep the properties here?

pjgg: That is actually true is like is not checking if the `trust-store-provider` is valid.

pjgg: Ok take a look to the original [PR](https://github.com/quarkusio/quarkus/issues/25972#issuecomment-1148746984) comment:
```
I think these random values are ignored by the Java Keystore implementation for well-known key store types, it probably defaults to a default provider for dealing with the PKCS12 type if a given provider can't be found - if you debug then you'll be able to confirm
```

So yes, maybe at the end we are not checking anything here.

pjgg:
Let's close for now this PR, because if the values are not valid then are ignored.

PR #985, opened 2023-01-10
Reactive Rest Client: Process paths before sub resources

pjgg:
### Summary

Verify that Reactive RestClient `@PathParam` parameters are applied to the web target before passing it to the sub-resource instance.

### Checklist:

### DIFF
--- a/http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/PlainRootResource.java
+++ b/http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/PlainRootResource.java
@@ -1,0 +1,51 @@
+package io.quarkus.ts.http.restclient.reactive;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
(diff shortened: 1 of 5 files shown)


michalvavrik (commented):
Thanks, I'm fine with this in principle, but I wonder if we can do better:

- all your `@PathParam` are String, please mix things up little bit (Integer, ....) as here we are testing byte code generated method, so let's check it's generated for all data types
- I mentioned this comment https://github.com/quarkusio/quarkus/pull/29821/files#diff-4816ed1f36211e410643f8c4ffe21e5dca287b7dbbc09cecff69332931bf1e3dR, AFAICT behavior for non-`@PathParam` did not change. What about if your client method had also formal parameters with `@QueryParam` or `@HeaderParam` etc. so that we test combination works?

I don't expect any failures though...

michalvavrik (approved):
As my suggestion were not required in order to test QUARKUS-2741 (I just though we it would look nicer), we can go with your version too :-)

PR #1046, opened 2023-02-08
Verify that Transfer-Encoding and Content-Length doesn´t goes together in the same http response

pjgg:
### Summary

The combination of these two HTTP headers is illegal, so let's make sure RESTEasy Reactive does not send the Content-Length header when Transfer-Encoding is set

JiraLink: https://issues.redhat.com/browse/QUARKUS-2754

### Checklist:


fedinskiy:
Since CI on `main` is broken at the moment, I run a manual verification: `mvn clean verify -pl http/http-minimum-reactive/,http/http-minimum/ -Dquarkus.platform.version=2.16.1.Final`. It worked, so I consider this solution to be verified

PR #1053, opened 2023-02-09
Disable CLI tests on native build

fedinskiy:
### Summary
Before that change, a single test was enabled in CLI for Native Since we do not install Quarkus CLI during Native runs in our Jenkins, this lead to failures.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


rsvoboda (changes requested):
I don't think this is proper change as it makes the test run for jvm mode runs.

and the test is about native build from CLI - "QuarkusCliClient.Result result = app.buildOnNative();"

Proper way imho to install CLI in the job.

PR #1101, opened 2023-03-09
Use RESTEasy CDI to inject beans with `@Inject` instead of `@Context`

michalvavrik:
### Summary

[It is now possible](https://github.com/gsmet/quarkus/commit/2d4332267a81ec49b9d950c44790a44327c8c6f0) to use `@Inject` on field injection points in RESTEasy Classic providers instead of `@Context`. RESTEasy Reactive has already supported this approach. [There is a plan to drop support](https://github.com/quarkusio/quarkus/issues/29240) of `@Context` on method parameters too, however this is not supported by RESTEasy (classic) yet. RESTEasy Reactive already supports this method parameter injection.

Changes done by this PR:

- module HTTP advanced:
  - `UriInfo` is newly injected with `@Inject`, however legacy `@Context` is stll used for `HttpRequest` injection
- module HTTP advanced reactive:
  - in `PremierLeagueContainerRequestFilter`, `UriInfo` is newly injected with `@Inject`, however legacy `@Context` is stll used for `HttpRequest` injection 
  - in `StringResponseSerializer`, `HttpServerRequest` is newly using `@Inject`
- module HTTP Rest Client Reactive:
  - I dropped the only `@Conext` that was there as the provider member wasn't used at all
- module OpenTelemetry Reactive (uses RESTEasy Mutiny):
  - dropped `@Context` from resource method param as we have several tests where resources are using `@Context` and I consider it interesting that `@Context` is in fact not needed there at all (we can see similar behavior on multiple RESTEasy Reactive upstream tests, e.g. in `io.quarkus.resteasy.reactive.jsonb.deployment.test.sse.SseResource`, `io.quarkus.resteasy.reactive.jackson.deployment.test.streams.StreamResource` and plenty others (same goes for `@Context` usage, it's frequently used in upstream)
- module Security - Basic
  - dropped `@Context SecurityContext security` from resource method parameters and instead, I used `@Inject` on setter. IMO we should test this pattern as it's legit and in this module, `@Context SecurityContext security` is used in 5 other cases (`UserResource`, `DenyAllResource`, `EveryoneResource`, `PermitAllResource` and `UnannotatedResource`). Note: `@Inject` can't be used on fields.

This change is done according to the [QUARKUS-2736 test plan](https://github.com/quarkus-qe/quarkus-test-plans/blob/main/QUARKUS-2736.md).

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/MultipleResponseSerializersResource.java
+++ b/http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/MultipleResponseSerializersResource.java
@@ -15,11 +15,11 @@ import java.io.OutputStream;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
 
+import jakarta.inject.Inject;
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.Path;
 import jakarta.ws.rs.Produces;
 import jakarta.ws.rs.WebApplicationException;
-import jakarta.ws.rs.core.Context;
 import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.MultivaluedMap;
 import jakarta.ws.rs.core.Response;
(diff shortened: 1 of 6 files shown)


mjurc (approved):
LGTM

PR #1105, opened 2023-03-10
Fix OpenAPI test failure and extend test coverage for security schemes

michalvavrik:
### Summary

With https://github.com/quarkusio/quarkus/pull/31671 expected behavior of Smallrye OpenAPI has changed (torwards the specification) in regards to expected security scheme for basic authentication. We can't expect any value for 'basic' auth in Spring Data module now, therefore I added additional tests to Keycloak OIDC client extended modules so that we have tested scenarios when security scheme is not empty.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/security/keycloak-oidc-client-extended/src/test/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/OpenApiStoreSchemaIT.java
+++ b/security/keycloak-oidc-client-extended/src/test/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/OpenApiStoreSchemaIT.java
@@ -2,6 +2,8 @@ package io.quarkus.ts.security.keycloak.oidcclient.extended.restclient;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.FileReader;
@@ -77,5 +79,38 @@ public class OpenApiStoreSchemaIT extends BaseOidcIT {
                 "Expected component.schema.Score object.");
         assertTrue(content.getJsonObject("paths").containsKey("/rest-ping"), "Missing expected path: /rest-ping");
         assertTrue(content.getJsonObject("paths").containsKey("/rest-pong"), "Missing expected path: /rest-pong");
+
+        // verify that path /secured/admin is only accessible by user with role 'admin'
+        var expectedRole = getRequiredRoleForPath(content, "/secured/admin");
+        assertEquals("admin", expectedRole);
+
+        // verify that path /secured/getClaimsFromBeans is accessible by any authenticated user
+        expectedRole = getRequiredRoleForPath(content, "/secured/getClaimsFromBeans");
+        // note: '**' is equivalent of @Authenticated and @RolesAllowed("**")
+        assertEquals("**", expectedRole);
+
+        // verify 'oidc' security schema
+        var securitySchema = content
+                .getJsonObject("components")
+                .getJsonObject("securitySchemes")
+                .getJsonObject("SecurityScheme");
+        var actual = securitySchema.getString("type");
+        assertEquals("openIdConnect", actual);
+        actual = securitySchema.getString("description");
+        assertEquals("Authentication", actual);
+        actual = securitySchema.getString("openIdConnectUrl");
+        assertNotNull(actual);
+        assertTrue(actual.endsWith("/auth/realms/test-realm/.well-known/openid-configuration"));
+    }
+
+    private static String getRequiredRoleForPath(JsonObject content, String path) {
+        return content
+                .getJsonObject("paths")
+                .getJsonObject(path)
+                .getJsonObject("get")
+                .getJsonArray("security")
+                .getJsonObject(0)
+                .getJsonArray("SecurityScheme")
+                .getString(0);
     }
 }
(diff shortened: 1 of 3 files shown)


michalvavrik:
Adding @fedinskiy as I'm afraid @jsmrcka is busy with other tasks.

security/keycloak-oidc-client-extended/src/test/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/OpenApiStoreSchemaIT.java
```diff
@@ -77,5 +79,40 @@ private void assertContent(JsonObject content) {
...
+        var expectedRole = getRequiredRoleForPath(content, "/secured/admin");
+        assertEquals("admin", expectedRole);
+
+        // verify that path /secured/getClaimsFromBeans is accessible by any authenticated user
+        expectedRole = getRequiredRoleForPath(content, "/secured/getClaimsFromBeans");
+        // note: '**' is equivalent of @Authenticated and @RolesAllowed("**")
+        assertEquals("**", expectedRole);
+
+        // verify 'oidc' security schema
+        var securitySchema = content.getJsonObject("components").getJsonObject("securitySchemes")
```
fedinskiy, line 93: Nitpick: could we have either all `getJsonObject`  calls on the same line, or each one on a separate one?

security/keycloak-oidc-client-extended/src/test/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/OpenApiStoreSchemaIT.java
```diff
@@ -77,5 +79,40 @@ private void assertContent(JsonObject content) {
...
+                .getJsonObject("SecurityScheme");
+        // - expected type is 'openIdConnect'
```
fedinskiy, line 95: This comment is not needed, we have line 97 which **shows** that.

Same for all other comments in this method, except the note.

Same goes for its reactive counterpart

spring/spring-data/src/test/java/io/quarkus/ts/spring/data/rest/additional/OpenAPIIT.java
```diff
@@ -61,12 +61,11 @@ void rolesAllowedResourceAuthPermitted() {
         assertNotNull(json.getString("paths.\"/article-jpa/{id}\".get"));
 
         json.setRootPath("");
-        assertEquals("admin", json.getString("paths.\"/secured/deny-all/{id}\".get.security[0].SecurityScheme[0]"));
-        assertEquals("admin", json.getString("paths./secured/roles-allowed.get.security[0].SecurityScheme[0]"));
-        assertEquals("admin", json.getString("paths.\"/secured/roles-allowed/{id}\".get.security[0].SecurityScheme[0]"));
-
-        // TODO: https://github.com/quarkusio/quarkus/issues/30997
-        //        assertEquals("user", json.getString("paths.\"/secured/roles-allowed/{id}\".delete.security[0].SecurityScheme[0]"));
+        // non-Oauth2 security requirement object should be empty
```
fedinskiy, line 64: Can we enable Oauth2 for Spring-data? Now it looks like there shouldn't be any security for spring and it doesn't seem right.

fedinskiy (changes requested):

spring/spring-data/src/test/java/io/quarkus/ts/spring/data/rest/additional/OpenAPIIT.java
michalvavrik: - In that case we won't have testing non-oauth2 
- It doesn't look like there is any security for Spring as OpenAPI definition says there is basic auth (situation didn't change in that respect), please see here https://github.com/quarkus-qe/quarkus-test-suite/blob/main/spring/spring-data/src/test/java/io/quarkus/ts/spring/data/rest/additional/OpenAPIIT.java#L84 only thing that changed is that path security scheme doesn't contain security requirement object as the specification doesn't allow that
- If I enable it for Spring Data, it will cost addition resources\
- OAuth2 OpenAPI definition is tested by OIDC tests I added (sits on top of Oauth2); OpenAPI generation is not really related to Spring/RESTEasy Reactive etc. The information is based on the security information provider (so here it is provided by vertx-http, for OIDC it is oidc extension) that provides the information to the OpenAPI extension

@fedinskiy I'd rather not do it unless I understand pros.

fedinskiy: Ok, so it is expected result of fixing https://github.com/quarkusio/quarkus/issues/27373.
I suggest to change the wording of the comment to make it a little bit more clear.
Something like "OpenAPI should generate per-scope security requirements only for OAuth2". Or post a link to https://github.com/OAI/OpenAPI-Specification/blob/3.0.1/versions/3.0.1.md#patterned-fields-3. Or both :)

fedinskiy: Ok, so it is expected result of fixing https://github.com/quarkusio/quarkus/issues/27373.
I suggest to change the wording of the comment to make it a little bit more clear.
Something like "OpenAPI should generate per-scope security requirements only for OAuth2". Or post a link to https://github.com/OAI/OpenAPI-Specification/blob/3.0.1/versions/3.0.1.md#patterned-fields-3. Or both :)

fedinskiy (changes requested):

PR #1130, opened 2023-03-26
Update Maven wrapper and fix Quarkus CLI test failures in daily builds

michalvavrik:
### Summary

Apache Maven and wrapper has been recently updated in upstream - https://github.com/quarkusio/quarkus/pull/31836, our CI is creating Quarkus CLI from JARs created by this wrapper and when we run `./quarkus-dev-cli build`, it is using our wrapper from Quarkus Test Suite root directory which is old one and incompatible with the one that was used for bulding CLI jar. I opened discussion here https://github.com/quarkusio/quarkus/issues/32081#issuecomment-1484208999 to confirm it is expected behavior (and documented). All files I added are copy & paste from Quarkus main project.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


.mvn/jvm.config
```diff
@@ -0,0 +1,4 @@
+-Dmaven.wagon.httpconnectionManager.ttlSeconds=120
+-Dmaven.wagon.http.retryHandler.requestSentEnabled=true
+-Dmaven.wagon.http.retryHandler.count=10
+-Xmx5g
```
rsvoboda, line 4: why is this set?

michalvavrik: I planned to use same settings as Quarkus does, but it seems like they only needs this to fix transient issues on Azure, so we should be fine. I dropped it.

michalvavrik:
@rsvoboda it looks like it is either getting fixed / reverted in upstream. it seems like a good thing we had this old wrapper as long as it's supported. let's wait little with this, if there is no fix tomorrow, I can disable Quarkus CLI test that is failing.

michalvavrik:
@rsvoboda looks like we have final decision now - 3.6.2 won't be supported now - https://github.com/quarkusio/quarkus/issues/32081#issuecomment-1485314049, so we need to do some upgrade in order to fix CI. This PR would do, let me know if you want something different.

rsvoboda (approved):

PR #1213, opened 2023-05-05
Let Dev Services run database init for reactive vanilla clients

michalvavrik:
### Summary

We discussed with @gtroitsk this TODO https://github.com/quarkus-qe/quarkus-test-suite/blob/main/sql-db/reactive-vanilla/src/main/java/io/quarkus/ts/reactive/db/clients/CommonResource.java#L15 and IMO https://github.com/quarkusio/quarkus/issues/19766 is not going to be fixed anytime soon. So when I looked at this, I can't see a point why we should use Flyway at all as these tests are only concerned with Dev Services and if you look to tickets https://issues.redhat.com/browse/QUARKUS-1080 https://issues.redhat.com/browse/QUARKUS-1408 I can't see relation to Flyway either. IMO running this manual init is not a good pattern as your tests don't wait for queries to finish, you just hope they finish before the first test query was triggered. I checked Flyway extension docs and code and while there is nothing that makes it JDBC datasource specific (it could work with reactive?) it is not tested anywhere and upstream issue also only discuss Dev Services. My proposal is to use init sql property that we already have.

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/sql-db/reactive-vanilla/src/main/java/io/quarkus/ts/reactive/db/clients/CommonResource.java
+++ b/sql-db/reactive-vanilla/src/main/java/io/quarkus/ts/reactive/db/clients/CommonResource.java
@@ -4,33 +4,7 @@ import java.net.URI;
 
 import jakarta.ws.rs.core.UriInfo;
 
-import org.jboss.logging.Logger;
-
-import io.vertx.mutiny.sqlclient.Pool;
-
 public class CommonResource {
-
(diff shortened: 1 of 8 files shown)


jsmrcka (approved):
Looks good, I like it.

The Flyway discussion was probably caused by the fact that the `init-script-path` property wasn't available until Quarkus 3.

PR #1222, opened 2023-05-11
Quarkus 3 OpenTelemetry hardening

michalvavrik:
### Summary

- OpenTelemetry JDBC instrumentation added to Narayana Transactions module as we already test there REST traces and start Jaeger:
  - tests https://github.com/quarkusio/quarkus/pull/31567, covers scenarios missing in upstream test coverage:
    - works in native (also) for MySQL and MSSQL datasource, important for we need to verify their drivers in native
    - use actual Jaeger container, upstream uses InMemorySpanExporter or mocks everywhere
    - test instrumentation in OpenShift environment with Red Hat registry images
  - tests https://github.com/quarkusio/quarkus/pull/28792 as upstream doesn't test this at all
    - in order to verify this issue, module `sql-db/narayana-transactions` is now using RESTEasy Reactive, which is preferred REST stack anyway and not relevant for Narayana tests
- [Smallrye Context Propagation with OpenTelemetry](https://github.com/quarkusio/quarkus/issues/30362) is also placed in Narayana Transactions module as there we already have (transitive) Smallrye Propagation Context dependency and running tests so we don't have to start extra application for that (this way, impact on resources is minimal)
- [Quarkus Scheduler tracing](https://github.com/quarkusio/quarkus/pull/29802) because upstream only test this with unit test and doesn't even check traces go to exporter
- [OTEL beans available according to the MicroProfile Telemetry Tracing specification](https://github.com/quarkusio/quarkus/pull/28735), it's important as upstream:
  - only use `@QuarkusTest` so we want to check normal lifecycle and native mode
  - only tests static OTEL methods returns same stuff as injected (so basically tests used class loader is correct), but we are actually going to use them little bit and see that traces are propagated to Jaeger
  - I absolutely failed to find reasonable usage of injected Baggage. It is immutable and I can't see to be able to propagate anything in.I can see that OTEL java instrumentation git project basically use it as carrier within span context, but then you can't get it into injected immutable baggage anyway. Upstream docs doesn't mention baggage anyway and doesn't support it for MDC logging. Perhaps I needed to use Jaeger OTEL propagator...
- [RESTEasy Classic and Reactive now have filters](https://github.com/quarkusio/quarkus/pull/30073) that adds 2 of [recommended code attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#source-code-attributes)
  - it's probably not super important feature, but I thought it would be nice to have at least smoke test for there filters (in native) and verification was added to other test so it won't cost us more than couple of seconds
- [OpenTelemetry SDK autoconfiguration](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md) has [now](https://github.com/quarkusio/quarkus/pull/30033) limited support for those OTEL configuration [properties we already support](https://github.com/quarkusio/quarkus/blob/main/extensions/opentelemetry/runtime/src/main/java/io/quarkus/opentelemetry/runtime/config/OTelFallbackConfigSourceInterceptor.java#L85)
  - upstream PR was massive, however it's importance is mostly for future changes, right now from user point of view, it is basically cutting down `quarkus.otel.some-property-name` to `otel.some-property-name` https://github.com/quarkusio/quarkus/blob/43a53588daa126477e24e81c75abfbf017547df5/extensions/opentelemetry/runtime/src/main/java/io/quarkus/opentelemetry/runtime/OpenTelemetryProducer.java#L143 and passing them to SDK builder, so what we want is to test that selected properties are passed to OTEL (otherwise we would be expensively testing whether someone didn't do typo)
    - we test that `quarkus.otel.service.name` has preference over `quarkus.application.name` and it doesn't https://github.com/quarkusio/quarkus/issues/33317 (so I opened issue)
    - we test that `quarkus.otel.attribute.value.length.limit` actually limits attribute length. that is OTEL business, but we want to ensure Quarkus OpenTelemetry extensions passed the property
  - important change is that `quarkus.opentelemetry` has changed to `quarkus.otel`, I didn't migrate properties as @linartova  is working on this and backwards compatibility should work for the moment being (and changes are not one to one, e.g. `quarkus.opentelemetry.tracer.exporter.otlp.endpoint` has changed to `quarkus.otel.exporter.otlp.traces.endpoint` as documented in [migration guide](https://github.com/quarkusio/quarkus/wiki/Migration-Guide-3.0#extension-re-write))

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/sql-db/narayana-transactions/src/test/java/io/quarkus/ts/transactions/TransactionCommons.java
+++ b/sql-db/narayana-transactions/src/test/java/io/quarkus/ts/transactions/TransactionCommons.java
@@ -2,22 +2,26 @@ package io.quarkus.ts.transactions;
 
 import static io.restassured.RestAssured.given;
 import static org.awaitility.Awaitility.await;
-import static org.hamcrest.Matchers.containsInAnyOrder;
 
 import java.time.Duration;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
 
 import org.apache.http.HttpStatus;
 import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.MethodOrderer;
+import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestMethodOrder;
 
 import io.quarkus.test.bootstrap.JaegerService;
 import io.quarkus.test.services.JaegerContainer;
 import io.restassured.http.ContentType;
 import io.restassured.response.Response;
 
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class) // we keep order to ensure JDBC traces are ready
 public abstract class TransactionCommons {
 
     static final String ACCOUNT_NUMBER_MIGUEL = "SK0389852379529966291984";
@@ -28,11 +32,11 @@ public abstract class TransactionCommons {
 
     static final String ACCOUNT_NUMBER_EDUARDO = "ES8521006742088984966899";
     static final int ASSERT_SERVICE_TIMEOUT_MINUTES = 1;
-    private Response jaegerResponse;
 
-    @JaegerContainer(useOtlpCollector = true, expectedLog = "\"Health Check state change\",\"status\":\"ready\"")
+    @JaegerContainer(expectedLog = "\"Health Check state change\",\"status\":\"ready\"")
     static final JaegerService jaeger = new JaegerService();
 
+    @Order(1)
     @Tag("QUARKUS-2492")
     @Test
     public void verifyNarayanaProgrammaticApproachTransaction() {
@@ -56,6 +60,7 @@ public abstract class TransactionCommons {
         Assertions.assertEquals(100, miguelJournal.getAmount(), "Unexpected journal amount.");
     }
 
+    @Order(2)
     @Tag("QUARKUS-2492")
     @Test
     public void verifyLegacyNarayanaLambdaApproachTransaction() {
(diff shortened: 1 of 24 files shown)


sql-db/narayana-transactions/src/test/java/io/quarkus/ts/transactions/TransactionCommons.java
```diff
@@ -119,14 +130,28 @@ public void verifyRollbackForNarayanaProgrammaticApproach() {
...
+    @Order(6)
+    @Test
+    public void verifyJdbcTraces() {
+        for (String operationName : getExpectedJdbcOperationNames()) {
+            verifyRequestTraces(operationName);
+        }
     }
 
+    protected String[] getExpectedJdbcOperationNames() {
+        return new String[] { "SELECT mydb.account", "INSERT mydb.journal", "UPDATE mydb.account" };
```
rsvoboda, line 151: Do you know what makes the operation names different for Oracle and MS SQL databases?

`AgroalOpenTelemetryWrapper` is used to get the details, so there are some specifics coming from the drivers or directly from the DB engine?

rsvoboda (changes requested):
One question, otherwise this looks really nice!

sql-db/narayana-transactions/src/test/java/io/quarkus/ts/transactions/TransactionCommons.java
michalvavrik: > Do you know what makes the operation names different for Oracle and MS SQL databases?

MS SQL is different due to DB name used by `SqlServerService`
https://github.com/quarkus-qe/quarkus-test-framework/blob/04f65af3d42547cb2e756ef9a509581af2b56474/quarkus-test-service-database/src/main/java/io/quarkus/test/bootstrap/SqlServerService.java#L7

Oracle database - TBH I don't know why it doesn't have table name in that trace as I didn't think it should concern us. Due to how this is implemented, there is no way this is Quarkus bug, either it is specific of Oracle driver, Agroal impl. or OpenTelemetry bug. I can dig into it if you want, but only on demand :-)

> AgroalOpenTelemetryWrapper is used to get the details, so there are some specifics coming from the drivers or directly from the DB engine?

`AgroalOpenTelemetryWrapper` is a trick (out of my head, please don't take this word for word): `OpenTelemetry` provides `java.sql.Datasource` wrapper that basically spy for every traced SQL/opened connected and whatever else can be tracked, but OTEL is optional dep in Agroal so we can't refer it from files that can be used without OTEL dep, so there is wrapper. Wrapper provides Agroal DS wraper that partially delegates to OTEL wrapper. But Agroal specifics still requires using Agroal DS too (=we can't wrap Agroal DS with OTEL wrapper)

rsvoboda: > I can dig into it if you want, but only on demand :-)

+1, please check that. It looks suspicious.

rsvoboda (approved):

PR #1196, opened 2023-04-26
[WIP] Fix DevModePostgresqlDevServiceUserExperienceIT failure when run in CI

michalvavrik:
### Summary

debugging here as I can't reproduce it locally, found here https://github.com/quarkus-qe/quarkus-test-suite/pull/1193#issuecomment-1523806447, here https://github.com/quarkus-qe/quarkus-test-suite/pull/1192 and here https://github.com/quarkus-qe/quarkus-test-suite/pull/1194

Please select the relevant options.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport
- [ ] New scenario (non-breaking change which adds functionality)
- [ ] This change requires a documentation update
- [ ] This change requires execution against OCP (use `run tests` phrase in comment)

### Checklist:
- [ ] Methods and classes used in PR scenarios are meaningful
- [ ] Commits are well encapsulated and follow [the best practices](https://cbea.ms/git-commit/)

PR #1223, opened 2023-05-14
Use Alpine image for flaky DevModePostgresqlDevServiceUserExperienceIT to mitigate possible causes of failure

michalvavrik:
### Summary

Literally every dependabot bump I've reviewed in past weeks had failed `DevModePostgresqlDevServiceUserExperienceIT`. Here https://github.com/quarkus-qe/quarkus-test-suite/pull/1196 I experimentally verified that removing of PG images from Docker works in Github CI and also that test only fails when full CI is running, when just selected modules (those combinations I tested) are run in CI, test passes. Locally it's impossible to reproduce (at least for me).

Only idea I have is to try to use image that no other test is using and see if the test is still flaky. It only "covers" flakyness, but failing test is doing us no good as other runs are cancelled. I do't have better solution, but if there is one, please go ahead.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


rsvoboda (approved):
+1 to try this

PR #1228, opened 2023-05-16
Workaround DevModeReactivePostgresqlDevServiceUserExperienceIT flakyness and don't start other ignored services

michalvavrik:
### Summary

Daily build 827 failed over `DevModeReactivePostgresqlDevServiceUserExperienceIT`, so I follow strategy described in https://github.com/quarkus-qe/quarkus-test-suite/pull/1223. These 3 UX DEV mode tests were always starting other dev services they didn't need, therefore I disabled them so that we safe time on test execution. Bullseye is current stable Debian, so that will change over time (we will be informed when test fails).

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/sql-db/reactive-vanilla/src/test/java/io/quarkus/ts/reactive/db/clients/DevModeReactiveMssqlDevServiceUserExperienceIT.java
+++ b/sql-db/reactive-vanilla/src/test/java/io/quarkus/ts/reactive/db/clients/DevModeReactiveMssqlDevServiceUserExperienceIT.java
@@ -23,6 +23,8 @@ public class DevModeReactiveMssqlDevServiceUserExperienceIT {
     @DevModeQuarkusApplication
     static RestService app = new RestService()
             .withProperty("quarkus.datasource.mssql.devservices.image-name", "${mssql.image}")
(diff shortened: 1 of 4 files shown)


rsvoboda:
These failures happen only on GitHub or are you able to reproduce them on physical machines too - local machine / Jenkins.

michalvavrik:
> These failures happen only on GitHub or are you able to reproduce them on physical machines too - local machine / Jenkins?

@rsvoboda 
I once (just yesterday) experienced this locally when running with `-Dit.test=DevModeReactivePostgresqlDevServiceUserExperienceIT` so I am absolutely sure this is not CI related or interference of other tests. IMHO 2 possibilities:

- utilities doesn't work in certain circumstances
- it's something related to image labeling and it is possible that image that can be classified as 15 is not labeled with that because it was downloaded under different label... really in order to even describe this correctly, I'd have to understand it and that takes time. I invested time to it yesterday without success and I don't have more time this week as long as you expect me to work on other tasks :-)

rsvoboda (approved):
Let's proceed with this change and thanks for the time you spent with this.

PR #1265, opened 2023-06-07
Replace OpenTracing with OTel in javaee-like-getting-started

jsmrcka:
### Summary

The goal is to have a module which combines OTel with other extensions, namely `smallrye-jwt` and `smallrye-openapi`.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


michalvavrik (commented):
Changes are okay, but I have feeling we should do more. It's really pity there are no tracing tests. I don't understand why there are currently opentraciing and jwt  (and others) deps while we don't use them, it is probably just to check compilation works.

Can you add one OpenTelemetry test class that:
- test OpenAPI traces (probably of OpenAPI document, but it could be of interest to also test access to Swagger UI endpoint)? I don' know if that's possible, but @rsvoboda mentioned it and I agree it makes sense in some cases; if OTel extension doesn't support it, never mind
- I think there are no Smallrye JWT traces we could trace, correct? If so, nothing there

michalvavrik (approved):
@jsmrcka I think my comments were valid, but changes LGTM, so if you fancy doing more, you can do it in next PR. Merging, thanks

Issue #1284, opened 2023-06-19
Investigate FIPS incompatibility for Strimzi container


michalvavrik:
`io.quarkus.ts.messaging.kafka.reactive.streams.SslAlertMonitorIT` is disabled on FIPS compatible machines as strimzi container fails to start with exception

```
Caused by: org.apache.kafka.common.KafkaException: Failed to load SSL keystore src/test/resources/kafka-truststore.p12 of type PKCS12
	at org.apache.kafka.common.security.ssl.DefaultSslEngineFactory$FileBasedStore.load(DefaultSslEngineFactory.java:375)
	at org.apache.kafka.common.security.ssl.DefaultSslEngineFactory$FileBasedStore.<init>(DefaultSslEngineFactory.java:347)
	at org.apache.kafka.common.security.ssl.DefaultSslEngineFactory.createTruststore(DefaultSslEngineFactory.java:320)
	at org.apache.kafka.common.security.ssl.DefaultSslEngineFactory.configure(DefaultSslEngineFactory.java:168)
	at org.apache.kafka.common.security.ssl.SslFactory.instantiateSslEngineFactory(SslFactory.java:140)
	at org.apache.kafka.common.security.ssl.SslFactory.configure(SslFactory.java:97)
	at org.apache.kafka.common.network.SslChannelBuilder.configure(SslChannelBuilder.java:73)
	at org.apache.kafka.common.network.ChannelBuilders.create(ChannelBuilders.java:192)
	at org.apache.kafka.common.network.ChannelBuilders.clientChannelBuilder(ChannelBuilders.java:81)
	at org.apache.kafka.clients.ClientUtils.createChannelBuilder(ClientUtils.java:105)
	at org.apache.kafka.clients.producer.KafkaProducer.newSender(KafkaProducer.java:511)
	at org.apache.kafka.clients.producer.KafkaProducer.<init>(KafkaProducer.java:454)
	... 78 more
Caused by: java.io.IOException: parseAlgParameters failed: PBE AlgorithmParameters not available
	at java.base/sun.security.pkcs12.PKCS12KeyStore.parseAlgParameters(PKCS12KeyStore.java:839)
	at java.base/sun.security.pkcs12.PKCS12KeyStore.engineLoad(PKCS12KeyStore.java:2074)
	at java.base/sun.security.util.KeyStoreDelegator.engineLoad(KeyStoreDelegator.java:222)
[...]

PR #869, opened 2022-10-14
Enable tests, which were disabled due to fixed issues

fedinskiy:
### Summary
Enable tests, which were disabled due to fixed issues and purge unused code

Please select the relevant options.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Dependency update
- [x] Refactoring
- [ ] Backport
- [ ] New scenario (non-breaking change which adds functionality)
- [ ] This change requires a documentation update
- [ ] This change requires execution against OCP (use `run tests` phrase in comment)

### Checklist:
- [x] Methods and classes used in PR scenarios are meaningful
- [x] Commits are well encapsulated and follow [the best practices](https://cbea.ms/git-commit/)

PR #1285, opened 2023-06-19
[2.13] Disable SslAlertMonitorIT on FIPS compatible machines

michalvavrik:
### Summary

This test was enabled https://github.com/quarkus-qe/quarkus-test-suite/pull/869 due to upstream issue being fixed https://github.com/quarkusio/quarkus/issues/23964, however I believe that is only the client side as I tested Quarkus Integration test both with and without `kafka.ssl.engine.factory.class` (and so I did with our test) and fact is that Strimzi container won't start as `PBE AlgorithmParameters not available` (when SSL engine factory is set on our TS, ex. is little different but signals provider doesn't know the algorithm).

It should be possible with 0.33+ (by default in 2.13 we are using 0.31) https://strimzi.io/blog/2023/01/25/running-apache-kafka-on-fips-enabled-kubernetes-cluster/ but only thing that I did experienced was different exception. If we want it to work, we need to update Strimzi version and algorithm.

Opened: https://github.com/quarkus-qe/quarkus-test-suite/issues/1284

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/messaging/kafka-streams-reactive-messaging/src/test/java/io/quarkus/ts/messaging/kafka/reactive/streams/SslAlertMonitorIT.java
+++ b/messaging/kafka-streams-reactive-messaging/src/test/java/io/quarkus/ts/messaging/kafka/reactive/streams/SslAlertMonitorIT.java
@@ -1,5 +1,7 @@
 package io.quarkus.ts.messaging.kafka.reactive.streams;
 
+import org.junit.jupiter.api.Tag;
+
 import io.quarkus.test.bootstrap.KafkaService;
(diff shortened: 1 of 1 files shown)


mjurc (approved):
NB: this would be nice to have tested, but we currently cannot maintain it considering `kafka-streams` are tech preview for now:

image

michalvavrik:
FYI - I just fixed imports order.

michalvavrik:
Failure not related.

PR #1300, opened 2023-07-02
Modest Hibernate hardening

michalvavrik:
### Summary

Verifies:

- https://github.com/quarkusio/quarkus/issues/28593 - allows setting quotation strategy
- https://github.com/quarkusio/quarkus/issues/28644 - multitennancy with discriminator column (as so far, we only tested db and scheme)
- https://github.com/quarkusio/quarkus/issues/24075 - Spring Data JPA repositories should support named datasources

Refactor:

- move `SpringDiIT` from `io.quarkus.ts.spring.data.rest` to `io.quarkus.ts.spring.data.di` package

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/sql-db/hibernate/src/main/java/io/quarkus/qe/hibernate/items/Item.java
+++ b/sql-db/hibernate/src/main/java/io/quarkus/qe/hibernate/items/Item.java
@@ -5,8 +5,10 @@ import jakarta.persistence.FetchType;
 import jakarta.persistence.Id;
 import jakarta.persistence.JoinColumn;
 import jakarta.persistence.ManyToOne;
+import jakarta.persistence.Table;
 
 @Entity
+@Table(name = "item") // import script expects lower case table name, identifiers are quoted, hence case-sensitive
 public class Item {
 
     @Id
--- a/spring/spring-data/src/main/java/io/quarkus/ts/spring/data/nameddatasource/AlternativeArticle.java
+++ b/spring/spring-data/src/main/java/io/quarkus/ts/spring/data/nameddatasource/AlternativeArticle.java
@@ -1,0 +1,58 @@
+package io.quarkus.ts.spring.data.nameddatasource;
+
+import jakarta.persistence.Entity;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+import jakarta.validation.constraints.NotBlank;
+
+import org.hibernate.validator.constraints.Length;
+
+@Entity
+public class AlternativeArticle {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
+    @Length(min = 2, max = 50, message = "length must be between {min} and {max}")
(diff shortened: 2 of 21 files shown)


michalvavrik:
I'll check failures.

michalvavrik:
Seems like a Quarkus bug.

michalvavrik:
Ready for review.

sql-db/hibernate/src/main/java/io/quarkus/qe/hibernate/items/Item.java
```diff
@@ -15,7 +17,7 @@ public class Item {
     public String note;
 
     @ManyToOne(fetch = FetchType.LAZY)
-    @JoinColumn(name = "customerId")
+    @JoinColumn(name = "customerid")
```
jsmrcka, line 20: I don't really like this. To me, it is needlessly complicated to decipher what is going on here.

Here is what I've managed to understand, let's see if I'm right:
- The way it was done before could be considered a showcase of the fact that the column name is case-insensitive in both column definition and in the import script.
- Now, because the quoting strategy is `all-except-column-definitions` and the column is explicitly quoted here, the case in the column definition matters. However, the import script is still interpreted in case-insensitive mode, because the column name is not quoted over there.

So, it is imo doing 3 things at once:
- testing the quoting strategy
- using explicit quoting
- relying on case-insensitive interpretation in the import script

All of which without a proper test, that would give it a bit of context.

I would appreciate a form that is more comprehensible. I'm ok with keeping the code, but let's add some comments both here and in the `import.sql`, which would explain the situation.

jsmrcka (changes requested):

sql-db/hibernate/src/main/java/io/quarkus/qe/hibernate/items/Item.java
michalvavrik: no, mmnt, wrong branch, I'll re-check

michalvavrik: > I don't really like this. To me, it is needlessly complicated to decipher what is going on here.

The point here is that we can't create table with "customerId" and refer to it with "customerId" when case sensitivity is important. It is legitimate case for you can see from linked issue there are users who need to use it.

> The way it was done before could be considered a showcase of the fact that the column name is case-insensitive in both column definition and in the import script.

If supported Quarkus extension provides option that users (and intentional customers) can use, we need to know it works (in theory, we can't cover everything). Every other module that tests Hibernate use case insensitivity, so I need to add it somewhere. It strikes me as interesting issue because when our customers are going to migrate their tables, it can possibly happen that their legacy tables will now be called with reserved / key words.

> All of which without a proper test, that would give it a bit of context.

It fails without changes which verifies the issue.


My conclusion: What I can do is to escape column in insert statement myself and keep customerId/customerid as it was. Also I can add comments as you suggested.

Thanks.

jsmrcka (approved):
LGTM

PR #1301, opened 2023-07-03
Fix daily runs

fedinskiy:
### Summary

First change enables back temporarily disabled test.
The second one disables CLI on native runs (it should be disabled in java code, but that doesn't currenctly work, see commit description for details).
The last one enables back test, which was temporarily disabled on native.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/nosql-db/infinispan/pom.xml
+++ b/nosql-db/infinispan/pom.xml
@@ -51,19 +51,5 @@
                 </plugins>
             </build>
         </profile>
-        <profile>
-            <!-- Disable native build on this module -->
-            <!-- https://github.com/quarkusio/quarkus/issues/30304 -->
-            <id>native</id>
-            <activation>
-                <property>
-                    <name>native</name>
-                </property>
-            </activation>
-            <properties>
-                <!-- To not build the module on Native -->
-                <quarkus.package.type>fast-jar</quarkus.package.type>
-            </properties>
-        </profile>
     </profiles>
 </project>
--- a/nosql-db/infinispan/src/test/java/io/quarkus/ts/infinispan/InfinispanIT.java
+++ b/nosql-db/infinispan/src/test/java/io/quarkus/ts/infinispan/InfinispanIT.java
@@ -8,12 +8,10 @@ import org.junit.jupiter.api.Test;
 import io.quarkus.test.bootstrap.InfinispanService;
 import io.quarkus.test.bootstrap.RestService;
 import io.quarkus.test.scenarios.QuarkusScenario;
-import io.quarkus.test.scenarios.annotations.DisabledOnNative;
 import io.quarkus.test.services.Container;
 import io.quarkus.test.services.QuarkusApplication;
 
 @QuarkusScenario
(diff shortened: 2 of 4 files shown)


michalvavrik:
I'd like to wait with this till we agree on _more permanent solution here_ https://github.com/quarkus-qe/quarkus-test-framework/pull/824. Please @fedinskiy @jsmrcka .

fedinskiy:
@michalvavrik @jsmrcka since `the more permanent solution` turned out to be dead end, the real solution was added to this module. Descriptions of PR and the commit were updated.

michalvavrik:
I can see that @fedinskiy, but changes to Github CI files are still there. Are you sure it is still necessary? Our CLI tests should be disabled by annotation on native now.

michalvavrik:
I just read `3fcdf4f665861944c6b39c8378bd726a269a0e18` description and I would expect that skipping build should be just enough. Please update commit or provide info (or shout back as I didn't actually test it).

fedinskiy:
@michalvavrik out cli tests are disabled in the pom, so there is no reason to try to enable them in the ci scripts, right?

.github/workflows/ci.yml
```diff
@@ -180,7 +180,7 @@ jobs:
           if [[ -n ${MODULES_ARG} ]]; then
             echo "Running modules: ${MODULES_ARG}"
             mvn -fae -V -B -s .github/mvn-settings.xml -fae -Dall-modules \
-                        -Dinclude.quarkus-cli-tests -Dts.quarkus.cli.cmd="${PWD}/quarkus-dev-cli" \
+                        -Dts.quarkus.cli.cmd="${PWD}/quarkus-dev-cli" \
```
jsmrcka, line 183: So, if I understand it right, this is not strictly necessary, because the `quarkus.build.skip` in the `quarkus-cli` module should be enough to not do native build on that module.

But we do it anyway, because:
1.  Why enable CLI tests in native pipeline, when the CLI module is not enabled for native build anyway.
2. CLI tests in general do not make much sense in native mode.

Is that correct, or did I miss anything?

If so, than this change (and the second one) is ok.

fedinskiy: You understand it right, and point one is correct. 
I presume, that there may be cli tests in our ts, which a) make sense in native mode b) are not called `QuarkusCli*IT.java`(so they are not affected by `include.quarkus-cli-tests`). This is why I left `ts.quarkus.cli.cmd` as is.
There are no such tests in quarkus-cli module, though.

A counter argument against this change, is that in (unlikely?) case, if we decide to add some native-related tests into quakus-cli, we would have to remember to fix ci scripts as well.

michalvavrik (approved):

.github/workflows/ci.yml
jsmrcka: In that case I prefer to discard this change.
It is IMO more systematic to not exclude all CLI test in the pipeline, but rather exclude individual modules or tests.

fedinskiy: The change discarded

jsmrcka (approved):

PR #1310, opened 2023-07-10
Fix AgroalPoolTest by removing interfering datasource connections

michalvavrik:
### Summary

With https://github.com/quarkusio/quarkus/pull/34262 our `AgroalPoolTest` is failing, without https://github.com/quarkusio/quarkus/pull/34262, it is not (no flakyness, always fails). Flyway is using Agroal datasources to open connection, so `jdbc.max-size`, `idle-removal-interval` and so on should work even for Flyway. I'm not sure what underlying Flyway changes are (it already took me long to find this one), but it is clear that we **actually** test both opened connections for both datasources, but expect only opened connections for one datasource. So when I ensured that idle connections are removed for `with-xa` and that `min-size` is zero, problem got fixed.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/sql-db/panache-flyway/src/main/resources/application.properties
+++ b/sql-db/panache-flyway/src/main/resources/application.properties
@@ -30,6 +30,12 @@ quarkus.flyway.schemas=test
 %agroal_pool_test.quarkus.datasource.jdbc.acquisition-timeout=60S
 %agroal_pool_test.quarkus.datasource.jdbc.validation-query-sql=SELECT CURRENT_TIMESTAMP
 %agroal_pool_test.quarkus.datasource.jdbc.new-connection-sql=SELECT CURRENT_TIMESTAMP
+# make sure all 'with-xa' connections are removed, and we only test default datasource ones
+%agroal_pool_test.quarkus.datasource.with-xa.jdbc.max-size=1
+%agroal_pool_test.quarkus.datasource.with-xa.jdbc.min-size=0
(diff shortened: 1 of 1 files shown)


fedinskiy:
@michalvavrik it looks like the problem is not in minsize, but in limit on max size of a pool

fedinskiy:
@michalvavrik could you answer in a separate comment, please? Or, at least, not in **my** comment :)

michalvavrik:
:-D, sorry, that was not my intention :-/ I don't think it is alright that I have right to adjust your comments btw.

michalvavrik:
> @michalvavrik it looks like the problem is not in minsize, but in limit on max size of a pool

I realize min size 0 is default, but our point is to allow 0 connections and make sure they are idle so that they are removed before we start testing. Max size is not important because we don't want any connections (ideally).

Update: I also set max size etc. Basically I wanted to be sure, we can cut it down if necessary.

fedinskiy:
It looks like we do not need xa connection for agroal test at all, so why don't we set `agroal_pool_test.quarkus.datasource.with-xa.jdbc.max-size` to zero and call it a day?

michalvavrik:
> It looks like we do not need xa connection for agroal test at all, so why don't we set `agroal_pool_test.quarkus.datasource.with-xa.jdbc.max-size` to zero and call it a day?

I tried, there is some minimum that doesn't allows it.

fedinskiy (approved):
Weird, that we didn't find this earlier. Thank you!

Issue #1181, opened 2023-04-18
Transform `SpringDiTest` by exposing endpoints that provides information injected to test


michalvavrik:
We use `@QuarkusTest` here
https://github.com/quarkus-qe/quarkus-test-suite/blob/c2b235a7473875eeffe07961802ca87ba5585c13/spring/spring-data/src/test/java/io/quarkus/ts/spring/data/di/SpringDiTest.java#L20

as we can't access right instance of `CDI.current()` in `QuarkusScenario`. However we can expose endpoints that provides same information. It should be copy & paste to endpoints and current test methods will just call these endpoint. When this is done, replace `@QuarkusTest` with `@QuarkusScenario`.

PR #1315, opened 2023-07-12
Transform SpringDiTest by exposing endpoints that provides information injected to test

linartova:
### Summary

Closes [issue](https://github.com/quarkus-qe/quarkus-test-suite/issues/1181).

We replaced SpringDiTest with SpringDiIT and BeanResource, because we want to  transform @QuarkusTest to @QuarkusScenario. 

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:


spring/spring-data/src/test/java/io/quarkus/ts/spring/data/rest/SpringDiIT.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Test
+    public void testBeanDeclaringClassMatch() {
+        app.given().get("/bean/beanDeclaringClassMatch")
+                .then()
+                .statusCode(200)
+                .body(is("true"));
+    }
+
+    @QuarkusApplication
```
michalvavrik, line 69: it is convention to have this class above all methods

spring/spring-data/src/main/java/io/quarkus/ts/spring/data/rest/BeanResource.java
```diff
@@ -0,0 +1,142 @@
...
+    @GET
+    @Path("/givenAccountServiceAutowiredToUserService")
+    @Produces("text/plain")
+    public boolean givenAccountServiceAutowiredToUserService_WhenGetAccountServiceInvoked_ThenReturnValueIsNotNull() {
+        UserService userService = CDI.current().select(UserService.class).get();
+        return userService.getAccountService() != null;
+    }
+
+    @GET
+    @Path("/givenPersonDaoAutowiredToSpringPersonServiceBySetterInjection")
```
michalvavrik, line 69: Few general notes to path:

- Strictly speaking camel case is sometimes used for paths, but IMHO it is extraordinary. For example Google recommends hyphnes https://developers.google.com/search/docs/crawling-indexing/url-structure please prefer `green-apple` to `greenApple`
- your method name is spot on and I'm quite sure that some reviewers of our team would prefer it much more than to what I'm going to suggest, but it is very long and it took me while to figure what it means. Can we simplify it so that when I browse through this resource, it is clear that one method tests SpringPersonService, second test UserService
- I'd like to follow conventions regarding REST naming, this is random page I found https://restfulapi.net/resource-naming/ please don't follow it word by word, but found more with search term `REST url`. Please note this is very opinionated subject and I don't have pattern on what is right and what is wrong, still:
  - this is Dependency injection test, so resource path `/dependency-injection` instead of `/bean` or something similar would alow you to drop `given...autowired` because it is clear what you are testing
  - you have to thing quite hard to understand what these methods/paths mean, can you simplify it heavily? My examples (find yours!):
    - subpath will be `/spring-person-service` and so full path (with resource path) is `/dependency-injection/spring-person-service` so it is clear that you test dependency injection of SpringPersonService?
    - if you want to test `SpringPersonService#getPersonDao`, you could expose new endpoint and put it to `/dependency-injection/spring-person-service/person-dao`. isn't it bit more readable?

spring/spring-data/src/test/java/io/quarkus/ts/spring/data/rest/SpringDiIT.java
linartova: Addressed.

spring/spring-data/src/main/java/io/quarkus/ts/spring/data/rest/BeanResource.java
linartova: Hello,
there is my solution:

- Addressed.
- I tried to make the names shorter and simpler.
- I learned something about REST naming. I applied your suggestions, except the last one, cause it would seem too segmented to me.

Appreciate the helpful suggestions!

michalvavrik (approved):
Nice work!

PR #1317, opened 2023-07-13
Fix for Openshift builds

fedinskiy:
### Summary

Our openshift builds now fail with error`java.lang.StringIndexOutOfBoundsException: begin 34, end -1, length 41`.
Bisection shows, that this is caused by parameter `quarkus.kubernetes-config.secrets.enabled=true`.
It seems, that the Framework changes this parameter into some kind of environment variable (or even two) and thus breaks Smallrye.
Since this option is fixed[1] at build time anyway, I suggest to simply move it into application.properties.

[1] https://quarkus.io/guides/kubernetes-config#quarkus-kubernetes-config_quarkus.kubernetes-config.secrets.enabled

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/config/src/main/resources/application.properties
+++ b/config/src/main/resources/application.properties
@@ -21,3 +21,5 @@ secrets.app-prop-sha256-handler=${sha256::0e8c2a8b8ecfbd52c3ef17acd44498ee2b892c
 
 # disable k8 dev services as we don't use them and currently Windows instances don't support them
 quarkus.kubernetes-client.devservices.enabled=false
+
+quarkus.kubernetes-config.secrets.enabled=true
--- a/config/src/test/resources/config-source-priority.properties
+++ b/config/src/test/resources/config-source-priority.properties
@@ -1,5 +1,4 @@
 quarkus.kubernetes-config.enabled=true
 quarkus.kubernetes-config.config-maps=app-config-yml,app-config-yaml,app-config
 quarkus.kubernetes-config.secrets=app-config,app-config-yml
-quarkus.kubernetes-config.secrets.enabled=true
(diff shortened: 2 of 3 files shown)


michalvavrik:
@fedinskiy this is very good catch as a hotfix (thanks!), but it also signals bug in our test framework because ``quarkus.kubernetes-config.secrets.enabled` is build time property (as you mentioned) and it is present in out `quarkus-test-core/src/main/resources/build-time-list` therefore putting it into dedicated properties file specific for this test class should be enough. Could you create issue in FW and add TODO to this PR please?

Disadvantage of this approach is that we are enabling secrets for all the tests and we can't test other tests without this property (it is good to have multiple combinations as we can't presume where the bug is, right?). In the end, if this is always going to be transformed into _sort of illegal env variable for SmallRye Config_, we need to find a way to avoid this or report config bug.

fedinskiy:
run tests

fedinskiy:
@michalvavrik I would say the way we handle properties in the framework is defective by design. We need to add new build-time properties into build-time-list` file, but they will be transformed into runtime parameters anyway (see `ExtensionOpenShiftQuarkusApplicationManagedResource.withEnvVars`)

The better option would be to explicitly set properties as build-time or runtime for every service. Some day, I will add this functionality :)

The more important question is why does Quarkus fails with `StringIndexOutOfBoundsException` at all when it sees weird environment params, but pinpointing the exact condition and creating a minimal reproducer will take too much time.

michalvavrik:
yeah, I agree @fedinskiy

PR #1346, opened 2023-08-04
Limit Java heap used during native image generation to fix build time analytics native failures

michalvavrik:
### Summary

Daily build in native is failing for several days in row as 

```
2023-08-04T04:57:17.3514775Z 04:40:29,705 INFO  [org.inf.HOTROD] ISPN004074: Native Epoll transport not available, using NIO instead: io.netty.channel.epoll.Epoll
2023-08-04T04:57:17.3515138Z 04:40:29,709 INFO  [org.inf.HOTROD] ISPN004108: Native IOUring transport not available, using NIO instead: io.netty.incubator.channel.uring.IOUring
2023-08-04T04:57:17.3515380Z Terminating due to java.lang.OutOfMemoryError: GC overhead limit exceeded
```

GraalVM GC use max 80% of physical memory https://www.graalvm.org/latest/reference-manual/native-image/optimizations-and-performance/MemoryManagement/ by default. Not sure that setting `MaximumHeapSizePercent` is great idea though, for we can't really tell how much memory is available and if we set it too high, GC won't be able to reclaim enough memory in some cases and fail again.

I can't reproduce it, but I think that setting max Java heap used during native image generation to 5 g similarly as Quarkus main project does https://github.com/quarkusio/quarkus/blob/main/.github/workflows/ci-actions-incremental.yml#L46 should help.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,6 +180,7 @@ jobs:
           if [[ -n ${MODULES_ARG} ]]; then
             echo "Running modules: ${MODULES_ARG}"
             mvn -fae -V -B -s .github/mvn-settings.xml -fae -Dall-modules \
+                        -Dquarkus.native.native-image-xmx=5g \
                         -Dinclude.quarkus-cli-tests -Dts.quarkus.cli.cmd="${PWD}/quarkus-dev-cli" \
                         -pl $MODULES_ARG clean verify -Dnative -am
           fi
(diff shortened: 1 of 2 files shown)


rsvoboda:
Setting xmx is a good idea. In BeefyTS we had `4g` as the limit, but here we probably have slightly bigger apps.
Let's wait for CI to see if `5g` can fly. A bit concerned about memory, but I think the runners have 7GB of RAM, so we should fit in.

michalvavrik:
> Setting xmx is a good idea. In BeefyTS we had `4g` as the limit, but here we probably have slightly bigger apps. Let's wait for CI to see if `5g` can fly. A bit concerned about memory, but I think the runners have 7GB of RAM, so we should fit in.

I didn't realize it at first, but we don't run native tests when changes were not only done in a specific module. I can setup CI to test this, but perhaps it would be best to merge this and trigger daily build (or wait for the next one). Daily build is already failing in native and CI of this PR won't tell us anything useful (especially JVM builds).

rsvoboda (approved):

PR #1351, opened 2023-08-07
Enable OIDC mTLS tests on OpenJDK 17 and FIPS using PKCS12 keystore

gtroitsk:
### Summary

Since OpenJDK 17 has FIPS support [1] for PKCS12 keystores, we can test it on FIPS and with Keycloak 22.

To test OIDC quarkus.oidc.tls.trust-store-provider we must set special provider that won't be well-known/default to Java.
KeyCloak officially supports BCFIPS provider, for these purpose  we need to setup custom Docker image [2], but  creating ad hoc docker image every time would cost time and prolong the test execution which is already so long. 

[1] https://issues.redhat.com/browse/RHELPLAN-110420
[2] https://www.keycloak.org/server/fips#_keycloak_server_in_fips_mode_in_the_container

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


security/oidc-client-mutual-tls/src/test/java/io/quarkus/ts/security/oidcclient/mtls/Pkcs12OidcMtlsIT.java
```diff
@@ -15,14 +15,12 @@
  * Keystore file type is automatically detected in following tests by its extension in quarkus-oidc.
  * Extension declared here is used by tests only.
  */
-@Tag("fips-incompatible")
+@Tag("fips-incompatible") // OpenJDK 11
```
michalvavrik, line 18: Can you only enable it for OpenJDK 17? One example is here https://www.baeldung.com/junit-5-conditional-test-execution (you will need to play with it little bit). So adding tag is complication, either you can introduce custom disabling check that ensures FIPS is disabled or you can refactor how we are working with "fips-incompatible" tags, either way, I'd like to disable it on FIPS if possible

security/oidc-client-mutual-tls/src/test/java/io/quarkus/ts/security/oidcclient/mtls/Pkcs12OidcMtlsIT.java
```diff
@@ -15,14 +15,12 @@
...
 @QuarkusScenario
 public class Pkcs12OidcMtlsIT extends KeycloakMtlsAuthN {
 
-    //TODO Remove workaround after Keycloak is fixed https://github.com/keycloak/keycloak/issues/9916
-    @KeycloakContainer(command = KC_DEV_MODE_P12_CMD, port = KEYCLOAK_PORT)
+    @KeycloakContainer(image = "quay.io/keycloak/keycloak:22.0", command = KC_DEV_MODE_P12_CMD, port = KEYCLOAK_PORT)
```
michalvavrik, line 22: There is no good reason to use Keycloak 19 as main is on 22, Ghost is on 21 (subject to change) and the last productized version is 18. Please change default to 22 and release framework. If we hardcode images like that, we need to remember changing it on several places, therefore we should have a good reason to set one specific version to one test. Changing default has advantage of using one image everywhere, therefore less images going to be pulled (less docker memory, less execution time).

security/oidc-client-mutual-tls/src/main/resources/application.properties
```diff
@@ -11,6 +11,7 @@ quarkus.oidc.tls.key-store-password=${ks-pwd}
 quarkus.oidc.tls.trust-store-password=${ks-pwd}
 
 # TODO https://github.com/quarkusio/quarkus/issues/25972
+# We cannot test these properties before we configure all environment to run KeyCloak with BCFIPS provider in container
```
michalvavrik, line 14: You probably mean we cannot test these properties in FIPS? Because SunJCE can work without BCFIPS provider. It's provider itself after all. SunJCE is part of OpenJDK now, isn't it? I can see it f.e. in 21 https://github.com/openjdk/jdk21u/blob/master/src/java.base/share/classes/com/sun/crypto/provider/SunJCE.java

Question: isn't Keycloak container based on OpenJDK 17 based on UBI image https://issues.redhat.com/browse/RHELPLAN-110420, therefore PKCS12 with SunJCE provider should be supported without BCFIPS?

security/oidc-client-mutual-tls/README.md
```diff
@@ -0,0 +1,42 @@
+# PKCS12 keystore and truststore generation for server and client
+
+
+1. Create a keystore for the client
+
+`keytool -genkey -alias Client -keyalg RSA -keystore client-keystore.p12 -keysize 2048 -storetype PKCS12 -storepass password -dname "cn=localhost, ou=QuarkusQE, o=Redhat, L=Brno, st=BR, c=CZ"`
```
michalvavrik, line 6: I created certificate this way a minute ago and it has validity for 90 days (`notAfter=Nov  6 14:38:57 2023 GMT`), can you specify validity and regenerate certificates? It's not ideal to redo this often.

michalvavrik (changes requested):

security/oidc-client-mutual-tls/src/main/resources/application.properties
gtroitsk: I mean that we cannot test it because default Java providers work well with no error message and the same behaviour is for random provider name, `quarkus.oidc.tls.key-store-provider=Random1` , for example. We don't get an error message, because "these random values are ignored by the Java Keystore implementation for well-known key store types, it probably defaults to a default provider for dealing with the PKCS12 type if a given provider can't be found", as Sergey wrote [here](https://github.com/quarkusio/quarkus/issues/25972#issuecomment-1148746984). To test these properties properly we need to set unknown provider for Java, BCFIPS is an option because it's supported by KeyCloak.

So, for PKCS12 keystores we can set SunJCE provider. Keycloak 22 container is based on UBI 9 and OpenJDK 17.

michalvavrik: thanks for explanation

security/oidc-client-mutual-tls/src/test/java/io/quarkus/ts/security/oidcclient/mtls/Pkcs12OidcMtlsIT.java
```diff
@@ -44,4 +48,24 @@ protected KeycloakService getKeycloakService() {
...
+            String javaVMVendor = new PropertyLookup("java.vm.vendor").get();
+
+            List<String> commandOutputLines = new ArrayList<>();
+            new Command("sysctl", "crypto.fips_enabled").outputToLines(commandOutputLines).runAndWait();
+            boolean isFipsEnabled = commandOutputLines.get(0).matches(".*1");
+
+            return javaVersion.matches("11.*") && javaVMVendor.matches(".*Red.*Hat.*") && isFipsEnabled;
+        }
+
+        return true;
```
michalvavrik, line 69: To sum this method up:

- Ubuntu => don't run
- (OpenJDK Red Hat 11) && FIPS enabled => don't run

Questions:

- Why shouldn't we run `Pkcs12OidcMtlsIT` on Ubuntu?
- We check OS for `sysctl crypto.fips_enabled` command returns error. I know exceptions are expensive, but dividing world on Ubuntu / the rest feels strange. You could avoid OS system check by catching error. BUT OS check is acceptable when done properly. What with happen when `sysctl cryptofips_enabled` is run on Windows? Because right now `Pkcs12OidcMtlsIT` is disabled on Github CI Windows, but in a week or 2, we will run it on our  own Windows with Docker, what then? Then even `grep` feels problematic. Can you please adjust this check to work on all OS types? I think it can be as simple as `(RH 11 && RHEL && FIPS enabled)` (in that exact order, OS only evaluated when 11 RH)

michalvavrik (approved):
Let's get it in, thanks.

PR #1394, opened 2023-09-02
Run QuarkusTests on Windows when Docker is available

michalvavrik:
### Summary

Explicitly enables `QuarkusTest` scenarios only when Docker is available for otherwise it fails on Windows runners with Windows containers. Prefix build time analytics tests with`QuarkusCli` for we need to detect that Quarkus CLI is used there. Also I have reservations to these names in general, but that's up to future discussion, not part of this PR.

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:


michalvavrik:
run tests

michalvavrik:
Changed the reviewer as Michal is on PTO.

build-time-analytics/src/test/java/io/quarkus/ts/buildtimeanalytics/QuarkusCliDevModeIT.java
```diff
@@ -24,20 +24,17 @@
...
 import io.quarkus.test.scenarios.QuarkusScenario;
 import io.quarkus.test.scenarios.annotations.DisabledOnNative;
 
-@DisabledOnOs(OS.WINDOWS) // TODO mvavrik: must only run when Quarkus CLI and Docker are available
 @Tag("QUARKUS-2812")
 @Tag("quarkus-cli")
 @QuarkusScenario
 @DisabledOnNative(reason = "Only for JVM verification")
-public class DevModeIT extends AbstractAnalyticsIT {
+public class QuarkusCliDevModeIT extends AbstractAnalyticsIT {
```
fedinskiy, line 37: Regarding renaming in general: why `@Tag("quarkus-cli")` is not enough?

fedinskiy, line 37: And why don't we add `@EnabledWhenLinuxContainersAvailable` here?

michalvavrik: Sure, if we don't go with the renaming, we can add the annotation and it will do the same job.

michalvavrik: It's not necessary, but shouldn't QuarkusCLi tests start with QuarkusCli? If you think about it, then `DevModeIT` really tells you nothing. You can look at packages for context, but it's not how we usually name CLI tests, no?

Anyway, from POV of my concern - detecting CLI test - the tag is just enough, I just didn't realize it during implementation (should have you as a reviewer :-)), so if you don't want to rename it, I'll annotate the test for now. @fedinskiy please state your final thoughts

michalvavrik: btw from packages or name it is not clear we use Quarkus CLI and therefore it won't be clear in tests results in Jenkins unless you check logs.

fedinskiy: I am strongly against using names for detection purposes (we already have this for Openshift, and it creates to much of visual clutter like `OpenShiftBinaryCloudEventsOverProdAmqpIT` and `OpenShiftPostgresqlTransactionGeneralUsageIT`), but I am fine with changing names for readability purposes. I suggest to change 'QuarkusCliSomethingIT' into 'CliSomethingIt', but that is up to you.

michalvavrik: I'm creating PR for FW to handle this, but basically my new approach is: use the annotation on all the CLI tests that requires Docker (as this way, we have chance to run all CLI tests that doesn't need it on Windows Github CI). Then I need to go step by step and determine when Docker is in action on CLI tests.

michalvavrik:
@fedinskiy I'd suggest to not make this PR about Quarkus CLI tests. Please review it as it is and I'll go one by one on Quarkus CLI tests, test whether they need Docker and use the annotation when necessary. Related to the https://github.com/quarkus-qe/quarkus-test-framework/pull/872

fedinskiy (approved):

PR #1437, opened 2023-09-30
Fix flaky DevModeOpenTelemetryIT especially on Windows with waiting till live reload has finished

michalvavrik:
### Summary

`DevModeOpenTelemetryIT` always fails on Windows (see respective Jenkins job), strange thing I can reproduce it 6 times of 10 on Linux, while in our CI it doesn't fail. Anyway, the root cause is that we don't wait till live reload is finished. That's what this PR does.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


mjurc (approved):
This is a good one, thanks

Issue #1408, opened 2023-09-12
Leaking Vertx and HttpClient


rsvoboda:
As part of upstream reviews I noticed comment from Clement saying:
```
        Vertx vertx = Vertx.vertx();
if you create your Vert.x you MUST close it at the end of the tests. Otherwise, it will leak.
```
Closing http client in such case is not enough.

I did a quick check of our codebase and we do poor resources cleanup.

TS
```
git grep "Vertx.vertx()"
http/vertx-web-client/src/test/java/io/quarkus/ts/http/vertx/webclient/PureVertxHttpClientTest.java:        httpClient = WebClient.create(Vertx.vertx(), new WebClientOptions());
http/vertx/src/test/java/io/quarkus/ts/vertx/AbstractVertxIT.java:        HttpClient httpClient = Vertx.vertx().createHttpClient();
http/vertx/src/test/java/io/quarkus/ts/vertx/AbstractVertxIT.java:        HttpClient httpClient = Vertx.vertx().createHttpClient(httpClientOptions);
http/vertx/src/test/java/io/quarkus/ts/vertx/AbstractVertxIT.java:        HttpClient httpClient = Vertx.vertx().createHttpClient();
security/vertx-jwt/src/test/java/io/quarkus/ts/security/vertx/AbstractCommonIT.java:        this.vertx = Vertx.vertx();
```

FW + TS (RestService#mutiny method)
```
git grep "Vertx.vertx()"
quarkus-test-core/src/main/java/io/quarkus/test/bootstrap/RestService.java:            webClient = WebClient.create(Vertx.vertx(), options
git grep 'mutiny('
examples/greetings/src/test/java/io/quarkus/qe/ReactiveGreetingResourceIT.java:        String response = app.mutiny().get("/reactive-greeting")

git grep 'mutiny(
http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/BaseHttpAdvancedReactiveIT.java:        Uni<JsonObject> content = getApp().mutiny(defaultVertxHttpClientOptions())
http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/BaseHttpAdvancedReactiveIT.java:        Uni<Integer> statusCode = getApp().mutiny(defaultVertxHttpClientOptions())
[...]

PR #1486, opened 2023-10-23
Create security/webauthn module with reactive mysql client

jcarranzan:
Based on the guide:
 https://es.quarkus.io/guides/security-webauthn 
and the quickstart:
 https://github.com/quarkusio/quarkus-quickstarts/tree/main/security-webauthn-quickstart I've created the new security/webauthn module with Mysql reactive client.

To execute the tests use the mvn command: 
`mvn verify -Dit.test=MySqlWebAuthnIT`


QuarkusQE:
Can one of the admins verify this patch?

security/webauthn/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-hibernate-reactive-panache</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-arc</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-test-security-webauthn</artifactId>
```
michalvavrik, line 35: We test real-world approach to Quarkus extensions (usually :-)), I very much believe that to test WebAuth you will need to use `WebClient` and do stuff as user would. 

Let me give an example of what I mean little below `AbstractWebAuthnTest.java` where you use `WebAuthnEndpointHelper` and `WebAuthnHardware`. At the very least,this must be test scoped in order to avoid registration of `WebAuthTestUserProvider`, but I don't think it is enough.

security/webauthn/src/main/java/io/quarkus/ts/security/webauthn/model/WebAuthnCertificate.java
```diff
@@ -0,0 +1,17 @@
...
+
+@Entity
+public class WebAuthnCertificate extends PanacheEntity {
+    @ManyToOne
+    public WebAuthnCredential webAuthnCredential;
+
+    /**
+     * The list of X509 certificates encoded as base64url.
+     */
+    public String x5c;
```
michalvavrik, line 16: no idea what `x5c` means, please rename this property

michalvavrik: Okay, I though about it and it probably means x509certificate. Please rename this property.

security/webauthn/src/test/java/AbstractWebAuthnTest.java
```diff
@@ -0,0 +1,156 @@
...
+        verifyLoggedOut(cookieFilter);
+
+        // two-step registration
+        String challenge = WebAuthnEndpointHelper.invokeRegistration(userName, cookieFilter);
+        JsonObject registrationJson = token.makeRegistrationJson(challenge);
+        if (endpoint == Endpoint.DEFAULT)
+            WebAuthnEndpointHelper.invokeCallback(registrationJson, cookieFilter);
+        else {
+            invokeCustomEndpoint("/register", cookieFilter, request -> {
+                WebAuthnEndpointHelper.addWebAuthnRegistrationFormParameters(request, registrationJson);
```
michalvavrik, line 52: What is happening when you are using `WebAuthHardware` and `WebAuthEndpointHelper` is that exactly same configuration of headers, encoding, signature, keys generator is used as is already tested by the extension unit tests and quickstart. 

I think the proper way to do this is to actually configure Microsoft Playwright we use for UI testing to provide hardware token. I think it can be difficult and to stop using this utility classes would increase time required for TD, so I'll let final decision on you.

This just seems very much like mocking.

security/webauthn/src/test/java/AbstractWebAuthnTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+        // logout
+        WebAuthnEndpointHelper.invokeLogout(cookieFilter);
+
+        verifyLoggedOut(cookieFilter);
+    }
+
+    private void invokeCustomEndpoint(String uri, Filter cookieFilter, Consumer<RequestSpecification> requestCustomiser) {
+        RequestSpecification request = given()
+                .when();
```
michalvavrik, line 88: both `when` behind `given` return same thing

image

michalvavrik (changes requested):

security/webauthn/src/main/java/io/quarkus/ts/security/webauthn/model/WebAuthnCertificate.java
jcarranzan: Renamed it!

security/webauthn/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-security-webauthn</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-hibernate-reactive-panache</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-arc</artifactId>
```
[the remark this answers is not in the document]
jcarranzan: I've removed it, thanks

security/webauthn/pom.xml
jcarranzan: I've removed it and I've used vertx HttpClient for testing purposes as you mentioned,  but I need to take another look to figure out what we really want to test...

security/webauthn/src/test/java/AbstractWebAuthnTest.java
jcarranzan: removed `when`

security/webauthn/src/test/java/AbstractWebAuthnTest.java
jcarranzan: Ok, at the beginning of this task, I was thinking of doing some selenium UI interaction with Webdriver but firstly I wanted to have something working as we already have.
But, if you use Microsoft Playwright I for UI testing will use it, do you know some uses case in quarkus test suite? I've checked that it is in the Quarkus test framework (DevModeQuarkusService) but that's all.

michalvavrik: > I've checked that it is in the Quarkus test framework (DevModeQuarkusService) but that's all.

Yeah, that's what I meant, if you can make it work with `org.htmlunit`, fine with me. Uou can just fulltext search for it inside QE Test FW, but I warn you, this is probably not going to be easy :-)

security/webauthn/pom.xml
michalvavrik: Well, I meant HTML Unit WebClient for this is a web app, but please don't take me word by word, I think principle is to test it close to the real world, which client is used is not important.

security/webauthn/pom.xml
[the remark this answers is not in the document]
michalvavrik: I still can see it.

security/webauthn/src/test/java/AbstractWebAuthnTest.java
jcarranzan: You were right is not easy, I've tried with playwright (https://github.com/microsoft/playwright/issues/7276) and with webclient (htmlunit) without success, however, I've found something that could be interesting with selenium webdriver-->
https://www.selenium.dev/documentation/webdriver/interactions/virtual_authenticator/
The register of a user seems to work but not really, I will do a push, could you take a look?

security/webauthn/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkus.ts.qe</groupId>
+        <artifactId>parent</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+        <relativePath>../..</relativePath>
+    </parent>
+    <artifactId>security-webauthn</artifactId>
+    <version>1.0.0-SNAPSHOT</version>
```
michalvavrik, line 11: <name>Quarkus QE TS: Security: WebAuth</name>

security/webauthn/src/test/java/AbstractWebAuthnTest.java
```diff
@@ -0,0 +1,110 @@
...
+import io.restassured.filter.Filter;
+import io.vertx.core.Vertx;
+import io.vertx.core.http.HttpClient;
+import io.vertx.core.http.HttpMethod;
+
+public abstract class AbstractWebAuthnTest {
+
+    protected abstract RestService getApp();
+
+    private static HttpClient httpClient = Vertx.vertx().createHttpClient();
```
michalvavrik, line 19: You are not closing neither Vertx or HttpClient, while this creates new instance of both. I think it should be closed or it can lead to resource leak. WDYT?

security/webauthn/src/test/java/AbstractWebAuthnTest.java
michalvavrik: > I will do a push, could you take a look? 

I can't really see anything or test it, CI fails over _[ERROR] /home/runner/work/quarkus-test-suite/quarkus-test-suite/security/webauthn/src/test/java/AbstractWebAuthnSelenium.java:[42,9] cannot find symbol_

Tests with selenium doesn't look bad, but it looks very similar to HtmlUnit and Playwright. I'd suggest that if you are in hurry as you say, let's start simple without it and do second PR. It will be easier to manage. Decision is completely yours!

michalvavrik: Also it brings in additional dependencies that needs to be managed, so there is cost of running CI.

michalvavrik:
@jcarranzan I think we discussed this in past already, but eventually, you will have to squash all these commits. That can wait to when this is not draft, sure.

security/webauthn/src/test/java/AbstractWebAuthnTest.java
jcarranzan: > > I will do a push, could you take a look?
> 
> I can't really see anything or test it, CI fails over _[ERROR] /home/runner/work/quarkus-test-suite/quarkus-test-suite/security/webauthn/src/test/java/AbstractWebAuthnSelenium.java:[42,9] cannot find symbol_
>   _Fixed it! Can you try now?_
> Tests with selenium doesn't look bad, but it looks very similar to HtmlUnit and Playwright. I'd suggest that if you are in hurry as you say, let's start simple without it and do second PR. It will be easier to manage. Decision is completely yours!
let's do a try to this and if tomorrow morning still not see the light then I will do simplest as possible.

security/webauthn/src/test/java/AbstractWebAuthnTest.java
jcarranzan: Yes it should be closed it as it is commented on and suggested in the other PR opened that we have: #1408

jcarranzan:
> @jcarranzan I think we discussed this in past already, but eventually, you will have to squash all these commits. That can wait to when this is not draft, sure.

I've just squashed and I will squash the next commits pushed, thank you.

security/webauthn/src/test/java/MySqlWebAuthnIT.java
```diff
@@ -0,0 +1,32 @@
...
+
+    private static final String MYSQL_USER = "quarkus_test";
+    private static final String MYSQL_PASSWORD = "quarkus_test";
+    private static final String MYSQL_DATABASE = "quarkus_test";
+    private static final int MYSQL_PORT = 3306;
+
+    @Container(image = "${mysql.80.image}", port = MYSQL_PORT, expectedLog = "Only MySQL server logs after this point")
+    static MySqlService database = new MySqlService()
+            .withUser(MYSQL_USER)
+            .withPassword(MYSQL_PASSWORD)
```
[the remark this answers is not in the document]
jcarranzan: done

security/webauthn/src/test/java/AbstractWebAuthnTest.java
```diff
@@ -0,0 +1,177 @@
...
+  private String adminAPIUrl = "/api/admin";
+
+  private String userName = "Roosvelt";
+  private static Filter cookieFilter;
+
+  @BeforeAll
+  public static void setup() {
+    vertx = Vertx.vertx();
+    httpClient = vertx.createHttpClient();
+    cookieFilter = new CookieFilter();
```
michalvavrik, line 41: I can't see `cookieFilter` used anywhere.

security/webauthn/src/test/java/MySqlWebAuthnIT.java
```diff
@@ -0,0 +1,32 @@
...
+
+@QuarkusScenario
+public class MySqlWebAuthnIT extends AbstractWebAuthnTest {
+
+    private static final String MYSQL_USER = "quarkus_test";
+    private static final String MYSQL_PASSWORD = "quarkus_test";
+    private static final String MYSQL_DATABASE = "quarkus_test";
+    private static final int MYSQL_PORT = 3306;
+
+    @Container(image = "${mysql.80.image}", port = MYSQL_PORT, expectedLog = "Only MySQL server logs after this point")
```
michalvavrik, line 15: Optionally:
we don't have these things very much defined, but when we use something, it's harder to stop using it later, because you don't know why this specific image was used. So I'd like to propose to use `mysql.upstream.80.image` instead, because it's easier to switch from upstream to product than otherwise when there is such requirement.

security/webauthn/src/test/resources/mysql.properties
```diff
@@ -0,0 +1,3 @@
+quarkus.datasource.db-kind=mysql
+quarkus.hibernate-orm.database.charset=utf-8
```
michalvavrik, line 2: default is `UTF-8`, so you can drop it

michalvavrik (changes requested):

michalvavrik (changes requested):
Please add @OpenShiftScenario.

security/webauthn/src/test/java/AbstractWebAuthnTest.java
```diff
@@ -0,0 +1,177 @@
...
+    RestAssured.given().then().statusCode(200);
```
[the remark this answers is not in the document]
jcarranzan: I wanted to check that everything(quarkus app) was ready before the tests, but yes, it is not needed to have in setUp

security/webauthn/src/test/java/AbstractWebAuthnTest.java
```diff
@@ -0,0 +1,177 @@
...
+      .get(publicAPIUrl)
+      .then()
+      .statusCode(200)
+      .body(Matchers.is("public"));
+  }
+
+  @Test
+  public void testRegisterWebAuthnUser() {
+
+    httpClient.request(HttpMethod.POST, registerUrl)
```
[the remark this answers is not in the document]
jcarranzan: Basically, because I am more familiar with this vertx httpclient approach and think as quarkus relies on vertx underneath I thought it was fine, but if you think it is to keep some homogeneity with the rest of the tests I could change it to the external high-level framework RestAssured.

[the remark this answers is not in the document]
michalvavrik: Not a blocker, it would make tests easier to read as I presume it would require less lines and you could use handy assertion methods. Please don't spend time on this suggestion unless you have it, np.

```
given().header("Content-Type", "text/plain").body(USERNAME).post(REGISTER_URL).then().status(200);
verifyLoggedIn(cookieFilter, USERNAME);
```

> quarkus relies on vertx underneath

yeah, but we test it separately, I don't think this adds additional value

security/webauthn/src/test/java/AbstractWebAuthnTest.java
michalvavrik: I can see it now.

michalvavrik: I think it wasn't there before.

security/webauthn/src/test/java/CookieFilter.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.http.impl.cookie.RFC6265StrictSpec;
+import org.apache.http.message.BasicHeader;
+
+import io.restassured.filter.Filter;
+import io.restassured.filter.FilterContext;
+import io.restassured.response.Response;
+import io.restassured.specification.FilterableRequestSpecification;
+import io.restassured.specification.FilterableResponseSpecification;
+
+public class CookieFilter implements Filter {
```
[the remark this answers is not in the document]
michalvavrik: It is now used.

security/webauthn/src/test/java/MySqlWebAuthnIT.java
jcarranzan: done

security/webauthn/src/test/java/AbstractWebAuthnTest.java
```diff
@@ -0,0 +1,177 @@
...
+      .compose(httpClientRequest -> {
+        httpClientRequest.putHeader("Content-Type", "text/plain")
+          .write(userName);
+        return httpClientRequest.send();
+      }).onSuccess(httpClientResponse -> {
```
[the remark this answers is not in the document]
jcarranzan: yes, you were right and this is now a problem because verifyLoggedIn wasn't called and when I've fixed it it seems verifyLoggedIn fails : 
`esponse body doesn't match expectation.
Expected: is "Roosvelt"
  Actual: <not logged in>`

security/webauthn/src/test/java/AbstractWebAuthnTest.java
```diff
@@ -0,0 +1,177 @@
...
+    vertx = Vertx.vertx();
+    httpClient = vertx.createHttpClient();
+    cookieFilter = new CookieFilter();
+    RestAssured.given().then().statusCode(200);
+
+  }
+
+  @AfterAll
+  public static void cleanUp() {
+    if (httpClient != null && vertx != null) {
```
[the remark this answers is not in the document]
jcarranzan: ok, I have just splited it on 2 if statements

security/webauthn/src/test/java/AbstractWebAuthnTest.java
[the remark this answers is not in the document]
jcarranzan: ok I will change it later

security/webauthn/src/test/java/MySqlWebAuthnIT.java
jcarranzan: I've come back to the origin image because works, with MySQL upstream image there some SQL errors that I had not time to investigate.

security/webauthn/src/test/java/AbstractWebAuthnTest.java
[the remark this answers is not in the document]
jcarranzan: I've already solved that with the introduction of the new methods and MyWebAuthnHardware class

security/webauthn/src/test/java/MyWebAuthnHardware.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        this.credID = new byte[32];
+        new Random().nextBytes(this.credID);
+        this.id = Base64.getUrlEncoder().withoutPadding().encodeToString(this.credID);
+    }
+
+    public JsonObject makeRegistrationJson(String challenge) {
+        JsonObject clientData = (new JsonObject()).put("type", "webauthn.create").put("challenge", challenge)
+                .put("origin", "http://localhost").put("crossOrigin", false);
```
michalvavrik, line 46: this won't work in OCP you need to use app.getHost...

security/webauthn/src/test/java/AbstractWebAuthnTest.java
```diff
@@ -0,0 +1,215 @@
...
+                .body(Matchers.is("<not logged in>"));
+
+        // user API not accessible
+        given()
+                .filter(cookieFilter)
+                .redirects().follow(false)
+                .get(USER_API_URL)
+                .then()
+                .statusCode(302)
+                .header("Location", Matchers.is("http://localhost:1101/login.html"));
```
michalvavrik, line 201: same here

security/webauthn/src/test/java/AbstractWebAuthnTest.java
```diff
@@ -0,0 +1,215 @@
...
+    private static Filter cookieFilter;
+
+    enum User {
+        USER,
+        ADMIN;
+    }
+
+    @BeforeAll
+    public static void setup() {
+        cookieFilter = new CookieFilter();
```
michalvavrik, line 42: Through this `cookieFilter` unit tests share state, but test class methods are not ordered. Please enforce method order so that we have explicitly set expected order of execution.

security/webauthn/src/test/java/OpenshiftWebAuthnIT.java
```diff
@@ -0,0 +1,15 @@
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.OpenShiftScenario;
+import io.quarkus.test.services.QuarkusApplication;
+
+@OpenShiftScenario
+public class OpenshiftWebAuthnIT extends AbstractWebAuthnTest {
+
+    @QuarkusApplication
+    static RestService app = new RestService();
```
michalvavrik, line 9: Your Quarkus app also needs to know about database, shouldn't you set datasource url?

michalvavrik (commented):
@jcarranzan added few comment, in general, I'll wait with proper review till I see tests green. I won't pretend I comprehend all the code, this feels very complicated and TBH I wouldn't want users to have to do all of this just for authentication. Wow, but let's go with it, that's not your fault.

README.md
```diff
@@ -755,13 +755,30 @@ this.router.get("/secured")
...
+MyWebAuthnHardware class has been created to simulate that hardware token.
+Also, we invoke some webauthn enpoint to do the user register.
+Restrictions are defined using common annotations (`@RolesAllowed` etc.).
+
+Test cases
+
+* Check the quarkus application, the different endpoint without any registered user.
+* Register a user with webauthn.
+* Check the login with the registered user.
+* Try to simulate a register user without the specific webauthn data needed.
```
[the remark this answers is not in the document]
jcarranzan: Ok, I've put it in a more descriptive way, thanks.

security/webauthn/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus.qe</groupId>
+            <artifactId>quarkus-test-service-database</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.dataformat</groupId>
+            <artifactId>jackson-dataformat-cbor</artifactId>
+            <version>2.15.2</version>
```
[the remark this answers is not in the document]
jcarranzan: Code linked where is using that library: 
https://github.com/quarkus-qe/quarkus-test-suite/blob/0e3ca4bd996a2f2cc62263bb4e2c75e97b6ae68b/security/webauthn/src/test/java/MyWebAuthnHardware.java#L49

https://github.com/quarkus-qe/quarkus-test-suite/blob/0e3ca4bd996a2f2cc62263bb4e2c75e97b6ae68b/security/webauthn/src/test/java/MyWebAuthnHardware.java#L100
Well, I conducted some tests without utilizing this library and encountered encoded JSON errors. However, it might be possible to leverage an inner Jackson dependency within Vert.x.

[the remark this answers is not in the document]
michalvavrik: Makes sense. This code originally comes from `quarkus-test-security-webauthn` that uses same dependency. The `quarkus-test-security-webauthn` is managed by Quarkus BOM, so can we include `quarkus-test-security-webauthn` with test scope instead and bring it as transitive dependency? Please feel free to object.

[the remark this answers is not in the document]
michalvavrik: Which reminds me, `<scope>test</scope>` please.

security/webauthn/src/test/java/AbstractWebAuthnTest.java
michalvavrik: My point is to guarantee that added tests are idempotent and expected state at each method execution is clear.

security/webauthn/src/test/java/MySqlWebAuthnIT.java
```diff
@@ -0,0 +1,32 @@
...
+    private static final String MYSQL_USER = "quarkus_test";
+    private static final String MYSQL_PASSWORD = "quarkus_test";
+    private static final String MYSQL_DATABASE = "quarkus_test";
+    private static final int MYSQL_PORT = 3306;
+
+    @Container(image = "${mysql.80.image}", port = MYSQL_PORT, expectedLog = "ready for connections")
+    static MySqlService database = new MySqlService()
+            .withUser(MYSQL_USER)
+            .withPassword(MYSQL_PASSWORD)
+            .withDatabase(MYSQL_DATABASE);
```
michalvavrik, line 19: Can you name one thing that setting `withUser`, `withPassword` and `withDatabase` will give us? I don't see an issue with it, but it's unnecessary noise, I'd only set there is some positive output of setting it. I tried it without these 3 methods and works.

security/webauthn/src/test/java/MySqlWebAuthnIT.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @QuarkusApplication
+    static RestService app = new RestService().withProperties("mysql.properties")
+            .withProperty("quarkus.datasource.username", database::getUser)
```
michalvavrik, line 23:
```suggestion
            .withProperty("quarkus.datasource.username", database.getUser())
```
It's static.

security/webauthn/src/test/java/MySqlWebAuthnIT.java
```diff
@@ -0,0 +1,32 @@
...
+            .withProperty("quarkus.datasource.password", database::getPassword)
```
michalvavrik, line 24:
```suggestion
            .withProperty("quarkus.datasource.password", database.getPassword())
```

security/webauthn/src/test/java/AbstractWebAuthnTest.java
```diff
@@ -0,0 +1,215 @@
...
+                .body(Matchers.is("<not logged in>"));
+
+        // user API not accessible
+        given()
+                .filter(cookieFilter)
+                .redirects().follow(false)
+                .get(USER_API_URL)
+                .then()
+                .statusCode(302)
+                .header("Location", Matchers.is(getApp().getHost() + "/login.html"));
```
michalvavrik, line 201: `getApp().getHost()` is deprecated, also using `getApp().getURI(Protocol.HTTP))` solves test failures.

security/webauthn/src/test/java/OpenshiftWebAuthnIT.java
```diff
@@ -0,0 +1,15 @@
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.OpenShiftScenario;
+import io.quarkus.test.services.QuarkusApplication;
+
+@OpenShiftScenario
+public class OpenshiftWebAuthnIT extends AbstractWebAuthnTest {
```
michalvavrik, line 6: It needs to be `OpenShiftWebAuthIT` for there is Maven exclude pattern and with lower case, this test is run in a bare metal.

michalvavrik (changes requested):

security/webauthn/pom.xml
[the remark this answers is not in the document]
jcarranzan: Well, I removed quarkus-test-security-webauthn dependency in my "draft" PR to try to test it as close to the client approach as possible, but even so, it was necessary to introduce some classes to make testing possible.
scope testing added.
If I include that library again, I will need to do extra work doing a refactor and delete the classes that have already created but I can do it if it is a better approach. so WDYT?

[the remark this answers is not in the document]
michalvavrik: > Well, I removed quarkus-test-security-webauthn dependency in my "draft" PR to try to test it as close to the client approach as possible

Adding `quarkus-test-security-webauthn` in test scope to use its dependencies doesn't affect prod-first approach because it doesn't change a **single** file.

**You don't need to add quarkust-test-security-webauth** because Jackson BOM is part of Quarkus BOM https://github.com/quarkusio/quarkus/blob/main/bom/application/pom.xml#L281 therefore `<version>2.15.2</version>` is absolutely unnecessary and it well lead to our CI running without any point. Drop `<version>2.15.2</version>` and forget about `quarkus-test-security-webauth`.

> If I include that library again, I will need to do extra work doing a refactor and delete the classes that have already created but I can do it if it is a better approach. so WDYT?

I've never suggested you use it, I suggested you include it for transitive deps, which as I mentioned above is also unnecessary. It wouldn't lead to refactoring anyway.

security/webauthn/src/test/java/OpenshiftWebAuthnIT.java
jcarranzan: OpenshiftWebAuthnIT extending from MySqlWebAuthnIT will resolve this I think

security/webauthn/src/test/java/MySqlWebAuthnIT.java
jcarranzan: You are right, the app is using the default config from the container's image config so here is not needed, I am removing those custom configurations.

security/webauthn/src/test/java/MySqlWebAuthnIT.java
jcarranzan: I am using  method references, and think it is clear,  legible and works, but if you think this could cause some discrepancies because in the abstract DatabaseService those variables are not static so I will change it as you mention

michalvavrik: Value is set when initializing instance assigned to static field. The point here is not whether method reference is clear and it has nothing to do with discrepancies. The point is that value is always going to be same for `database":getUser` and `database.getUser()` and method reference will take additional space in metaspace, nah? I don't understand why to use it, that is all.

To be clear, these are unimportant things, I am pointing them out to provide you with information, I don't insist on changes.

security/webauthn/pom.xml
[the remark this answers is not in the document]
jcarranzan: ok get it, thanks for the explanation, I will keep the dependency: 
`<dependency>
            <groupId>com.fasterxml.jackson.dataformat</groupId>
            <artifactId>jackson-dataformat-cbor</artifactId>
            <scope>test</scope>
        </dependency>`

security/webauthn/src/test/java/MySqlWebAuthnIT.java
jcarranzan: ok Michal, I really appreciate this kind of information, thank you.

security/webauthn/src/test/java/OpenshiftWebAuthnIT.java
jcarranzan: Good catch!, changing class name to `OpenShiftWebAuthIT`

security/webauthn/src/test/java/AbstractWebAuthnTest.java
jcarranzan: I've included @TestMethodOrder(MethodOrderer.OrderAnnotation.class) at level class on AbstractWebAuthnTest.
And also @Order() test methods.

michalvavrik:
native failure seems related

jcarranzan:
> native failure seems related

Same failures on Openshift

security/webauthn/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+    <parent>
+        <groupId>io.quarkus.ts.qe</groupId>
+        <artifactId>parent</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+        <relativePath>../..</relativePath>
+    </parent>
+    <artifactId>security-webauthn</artifactId>
+    <version>1.0.0-SNAPSHOT</version>
+    <name>Quarkus QE TS: Security: WebAuth</name>
+    <dependencies>
```
michalvavrik, line 13: all tests you are adding needs a database and `quarkus.datasource.db-kind` is build-time property, therefore you know our QE Test FW will always re-build native executable and build performed my Quarkus Maven plugin will be ignored. You can spare few minutes by adding

```
<properties>
    <quarkus.build.skip>true</quarkus.build.skip>
</properties>
```

security/webauthn/src/test/resources/mysql.properties
```diff
@@ -0,0 +1,2 @@
+quarkus.datasource.db-kind=mysql
+quarkus.hibernate-orm.database.generation=drop-and-create
```
michalvavrik, line 2: you already set this property i n`application.properties`, no need to repeat it

security/webauthn/src/test/java/MySqlWebAuthnIT.java
```diff
@@ -0,0 +1,26 @@
...
+@QuarkusScenario
+public class MySqlWebAuthnIT extends AbstractWebAuthnTest {
+
+    private static final int MYSQL_PORT = 3306;
+
+    @Container(image = "${mysql.80.image}", port = MYSQL_PORT, expectedLog = "ready for connections")
+    static MySqlService database = new MySqlService();
+
+    @QuarkusApplication
+    static RestService app = new RestService().withProperties("mysql.properties")
```
michalvavrik, line 16: I've added following waiting which fixed native mode. I don't suggest it as fix, but I believe it provides you with valuable information - the issue is that Quarkus app starts before app is ready. Personally I'd try to compare this with other Hibernate Reactive tests and see what is difference (IMO there is none, but I only had quick look). I'll leave you there as I have my own tasks. Good luck.

```
.onPreStart(new Action() {
                @Override
                public void handle(Service service) {
                    try {
                        Thread.sleep(Duration.ofSeconds(5).toMillis());
                    } catch (InterruptedException e) {
                        throw new RuntimeException(e);
                    }
                }
            })
```

jcarranzan: I also tried with your code lines but it failed in my ocp instance with the same failures. Also I increased the seconds...

michalvavrik:
run tests

michalvavrik:
run tests

jcarranzan:
> run tests

All tests passed again.

michalvavrik (changes requested):
OC fails over added tests. Please have a look.

security/webauthn/src/test/java/AbstractWebAuthnTest.java
```diff
@@ -0,0 +1,228 @@
...
+                .contentType(ContentType.JSON)
+                .body("{\"name\": \"" + newUserName + "\"}")
+                .post(REGISTER_URL)
+                .then()
+                .statusCode(is(200)).extract();
+        given().filter(cookieFilter)
+                .get(PUBLIC_ME_API_URL)
+                .then()
+                .statusCode(200)
+                .body(Matchers.is("<not logged in>"));
```
michalvavrik, line 124: please also add call to secured endpoint that will return 401 or 403

jcarranzan: Well, in that test case, as the user is not logged the admin API endpoint will throw a 404, if the user is logged in correctly and the user is not an Admin the status code will be 403

michalvavrik (approved):

security/webauthn/src/test/java/OpenShiftWebAuthnIT.java
```diff
@@ -0,0 +1,9 @@
+import org.junit.jupiter.api.Disabled;
```
michalvavrik, line 1: please add packages to test classes

PR #1501, opened 2023-11-03
Fix Kafka Reactive Streams tests on Windows

michalvavrik:
### Summary

Fixes failing tests on Windows. 

- Kafka SSL fix is of same sort as we did in other modules (certificate issue).

- Processing of incoming topics during graceful shutdown just seems slower on Windows, I can't explain it. I've refactored it to avoid restart before testing graceful shutdown because amount of time needed to wait before app was accepting incoming topics was enormous. I spent hours trying to findout some timeout or other reasons (tried readyness checks calling etc.) and I can't find a reason why is it slower.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/messaging/kafka-streams-reactive-messaging/src/test/java/io/quarkus/ts/messaging/kafka/reactive/streams/KafkaGratefulShutdownIT.java
+++ b/messaging/kafka-streams-reactive-messaging/src/test/java/io/quarkus/ts/messaging/kafka/reactive/streams/KafkaGratefulShutdownIT.java
@@ -8,7 +8,11 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
 
 import org.jboss.logmanager.Level;
+import org.junit.jupiter.api.MethodOrderer;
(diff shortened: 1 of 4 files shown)


mjurc (approved):
LGTM

michalvavrik:
> This maybe warrants an issue, just to show there may be a problem

Oh yeah, I hoped you won't realize that. It maybe be an issue in native libraries, or it may also be issue in Kafka itself or somehow connected to running Kafka in Minikube - I have no idea. The extension isn't doing anything specific for Windows and it works as expected first time (I run tons of tests, it's not flaky). The application is idempotent, so it will start second time a same way (please note it is not related to the graceful shutdown flag present at build time, I tested it). So I think it is less likely to be in Quarkus. I can create a reproduce and open Quarkus issue if you like...

michalvavrik:
@mjurc https://github.com/quarkusio/quarkus/issues/36862 (still bit worried this will cost me a lot of time, but hopefully I'm wrong).

PR #1504, opened 2023-11-06
Adapt OpenShift MongoDB tests to work with official Docker image

michalvavrik:
### Summary

Official MongoDB Docker image requires read / write on `/data/db` directory and because it is not adapted to work with random user ID, this user doesn't have permission to edit `/data/db` by default and startup in OCP fails with `"DBException in initAndListen, terminating","attr":{"error":"IllegalOperation: Attempted to create a lock file on a read-only directory: /data/db"}}`.

I've considered using `pre` lifecycle hook to grant permissions to this file, but it also means to handle `entrypoint` ourselves. This PR proposes to use `emptyDir` volume to store `/data/db`. Storing this data on mounted volume is legit strategy explained https://hub.docker.com/_/mongo in the _Where to Store Data_ section.

Please note that in next PR I'll allow run MongoDB tests on Aarch64, I didn't forget about it, but it's a separate issue that needs to be tested differently.

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/nosql-db/mongodb-reactive/src/test/resources/test.properties
+++ b/nosql-db/mongodb-reactive/src/test/resources/test.properties
@@ -1,3 +1,5 @@
 ts.app.log.enable=true
 ts.database.log.enable=true
 ts.database.openshift.use-internal-service-as-url=true
+# store MongoDB data in 'emptyDir' volume to work-around write permission issue
+ts.database.openshift.template=/mongo-openshift-deployment-template.yml
--- a/pom.xml
+++ b/pom.xml
@@ -263,7 +263,7 @@
                                 <!-- Oracle image - we aim to use same version as Dev Services for Oracle so that we only download one image -->
                                 <oracle.image>docker.io/gvenzl/oracle-free:23-slim-faststart</oracle.image>
                                 <db2.image>quay.io/quarkusqeteam/db2:11.5.7.0</db2.image>
-                                <mongodb.image>docker.io/bitnami/mongodb:5.0</mongodb.image>
+                                <mongodb.image>docker.io/library/mongo:5.0</mongodb.image>
                                 <redis.image>docker.io/library/redis:6.0</redis.image>
                                 <wiremock.image>quay.io/ocpmetal/wiremock</wiremock.image>
                                 <!-- TODO use official image when/if this fixed https://github.com/hashicorp/docker-consul/issues/184 -->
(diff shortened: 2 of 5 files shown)


michalvavrik:
run tests

michalvavrik:
OCP failures unrelated.

nosql-db/mongodb-reactive/src/test/resources/test.properties
```diff
@@ -1,3 +1,5 @@
 ts.app.log.enable=true
 ts.database.log.enable=true
 ts.database.openshift.use-internal-service-as-url=true
+# store MongoDB data in 'emptyDir' volume to work-around write permission issue
+ts.database.openshift.template=/mongo-openshift-deployment-template.yml
```
fedinskiy, line 5: wow, I didn't know we can do **that**!

pom.xml
```diff
@@ -263,7 +263,7 @@
                                 <!-- Oracle image - we aim to use same version as Dev Services for Oracle so that we only download one image -->
                                 <oracle.image>docker.io/gvenzl/oracle-free:23-slim-faststart</oracle.image>
                                 <db2.image>quay.io/quarkusqeteam/db2:11.5.7.0</db2.image>
-                                <mongodb.image>docker.io/bitnami/mongodb:5.0</mongodb.image>
+                                <mongodb.image>docker.io/library/mongo:5.0</mongodb.image>
```
fedinskiy, line 266: Did you try this on podman? It also uses randomly assigned UID, so it may have the same problem on baremetal tests

michalvavrik: I created my own instance on OpenStack with Podman and it worked. I presume it goes down to how filesystem is shared or something, but I definitely don't know that (nor I care TBH).

nosql-db/mongodb-reactive/src/test/resources/test.properties
michalvavrik: I was actually looking for a way I can just customize default template, but it is not possible ATM and I believe it will change when Michal Jurc finishes migration to Fabric8 Java client (from OC tool).

michalvavrik: So then we can refactor this.

rsvoboda: @mjurc fyi

Issue #1146, opened 2023-04-02
Update tests using `mongodb.image` to be able to consume `arm64v8/mongo`


mjurc:
Our nosql coverage uses bitnami/mongodb, which doesn't release multiarch container right now. This is problematic for testing nosql capabilities on OCP on ARM.

[`arm64v8/mongo`](https://hub.docker.com/r/arm64v8/mongo) provides the container to run MongoDB on ARM, but the container is based on [`library/mongo`](https://hub.docker.com/_/mongo) rather than on bitnami container [we use](https://github.com/quarkus-qe/quarkus-test-suite/blob/main/pom.xml#L243). Replacing the currently used container with the [`arm64v8/mongo`](https://hub.docker.com/r/arm64v8/mongo) results in DB never reaching readiness.

We should either replace our default mongodb container by the [`library/mongo`](https://hub.docker.com/_/mongo) or create some sort of glue that would enable us to use both. I'd be inclined to use the official image.

PR #1507, opened 2023-11-07
Use Arm64 specific MongoDB image for OSD ARM

michalvavrik:
### Summary

closes: #1146

OSD ARM64 should use MongoDB image for platform ARM64, OS Linux and variant v8. I didn't find a way to specify platform in Pod object definition https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#Container and even though OSD is running on ARM64, I checked digest from `oc describe pod ...` and inspected it with 
 `docker inspect manifest docker.io/library/mongo@sha256:0d8482aef8d58e787fcf8e5e4aa81440f0ff1d9017c63958584b22312da10937` - it was still amd64. Stuff like `docker pull arm64v8/mongo` suggested here https://hub.docker.com/r/arm64v8/mongo doesn't even work for me locally, let alone in OCP (you need to specify --platform).

Fun fact, it works on AMD64 architecture image anyway, but requirement is to use ARM64 specific image.

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -789,6 +789,8 @@
                                         <mariadb.105.image>registry.redhat.io/rhel8/mariadb-105</mariadb.105.image>
                                         <amq-streams.image>registry.redhat.io/amq7/amq-streams-kafka-27-rhel7</amq-streams.image>
                                         <amq-streams.version>1.7.0</amq-streams.version>
+                                        <!-- run 'docker manifest inspect mongo:5.0' and use digest for architecture 'arm64', variant '8', os 'linux' -->
+                                        <mongodb.image>docker.io/library/mongo@sha256:6f851e31a1b317c6fa681b7dad5f94c622f1c3588477f3b769579dc5462ee659</mongodb.image>
                                     </systemPropertyVariables>
                                 </configuration>
                             </execution>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -789,6 +789,7 @@
                                         <mariadb.105.image>registry.redhat.io/rhel8/mariadb-105</mariadb.105.image>
                                         <amq-streams.image>registry.redhat.io/amq7/amq-streams-kafka-27-rhel7</amq-streams.image>
                                         <amq-streams.version>1.7.0</amq-streams.version>
+                                        <mongodb.image>docker.io/library/mongo@sha256:6f851e31a1b317c6fa681b7dad5f94c622f1c3588477f3b769579dc5462ee659</mongodb.image>
```
rsvoboda, line 792: Why doesn't the default `<mongodb.image>docker.io/library/mongo:5.0</mongodb.image>` work?
The intention to move to the official image instead of bitnami was to have OOTB experience on arm.

Also a note how to figure out the right image / sha256 would be handy.

michalvavrik: > Why doesn't the default <mongodb.image>docker.io/library/mongo:5.0</mongodb.image> definition work?

It works, but it is image build for platform amd64, which is not what we want to test

> The intention to move to the official image instead of bitnami was to have OOTB experience on arm.

move to the official image was so that we can use ARM64 platform specific image, please note that Docker images are built for different platforms. You can try `docker manifest inspect mongo:5.0` and see what are possibilities for different platforms

> Also a note how to figure out the right image / sha256 would be handy.

I wrote it into this description. I can write it into comment (well, I'll do it for sake of having this approved), but I think if someone want to change something, he should know what he is doing (or learn it from history).

michalvavrik: Added comment how to choose correct image digest.

michalvavrik: Sorry, I don't see in the PR's description, I had planned to write it there.

rsvoboda (approved):

PR #1508, opened 2023-11-07
Add test coverage for issue 36402

mocenas:
Create test coverage for issue: https://github.com/quarkusio/quarkus/issues/36402.

### Summary

Issue fails the Server Sent Events (SSE) while in native mode. New coverage adds app to simulate & test this behavior.
Original reproducer is [here](https://github.com/sdaschner/quarkus-playground/tree/36402-sse-native-cli).

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


http/http-sse/src/test/java/http/sse/HttpSseIT.java
```diff
@@ -0,0 +1,27 @@
...
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.scenarios.QuarkusScenario;
+
+@QuarkusScenario
+@Tag("QQE-257")
```
michalvavrik, line 13: This is very new thing, I'm not sure it is necessary, but it really depends on team discussion :-/ We are using tags for Quarkus tickets, but there is a difference, because this is a ticket used for work planning.

michalvavrik (commented):
Hey @jedla97 , I have more than one comment, but I have doubts we need a new module and so much code to test missing classes in the REST Client Classic. I'll verify you are right and get back to you when I have a time. Please wait till then, thank you.

michalvavrik (changes requested):
Hi @jedla97 ,

I have tested it a bit and all that I needed to reproduce the issue was following patch (I'm pretty sure you simplify that as well, but I don't have more time)

```
Subject: [PATCH] fix uma_protection
---
Index: http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/SseClient.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/SseClient.java b/http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/SseClient.java
new file mode 100644
--- /dev/null	(date 1699377305495)
+++ b/http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/SseClient.java	(date 1699377305495)
@@ -0,0 +1,18 @@
+package io.quarkus.ts.http.advanced;
+
+import jakarta.enterprise.event.Observes;
+import jakarta.enterprise.event.Startup;
+import jakarta.ws.rs.client.ClientBuilder;
+import jakarta.ws.rs.client.WebTarget;
+import jakarta.ws.rs.sse.SseEventSource;
+
+@ApplicationScoped
+public class SseClient {
+
+    void callSomeMethod(@Observes Startup startup) {
+        WebTarget target = ClientBuilder.newClient().target("http://localhost:8080/updates");
+        var src = SseEventSource.target(target).build();
+        System.out.println("I have a client " + src);
+    }
+}
```
with command `mvn clean verify -Dnative -Dit.test=HttpAdvancedIT -Dquarkus.platform.version=3.2.7.Final`

Please refactor it to your needs. Do you agree that it will cost use much less of execution time and it's easier to read? Please feel free to contradict. Thanks

jedla97:
Hi, @michalvavrik did you mean me or it was mistake as the@ mocenas created this PR and the questions more target PR creator.

http/http-sse/src/test/java/http/sse/HttpSseIT.java
jedla97: My POV on this is that it's ok use tag but I would change qqe to quarkus issue. It make more sense for me, also qqe is only internal.

michalvavrik:
Sorry @jedla97 for some reason I thought you opened issues. I thought it was weird considering to who is assigned to the JIRA ticket :-D

http/http-sse/src/test/java/http/sse/HttpSseIT.java
michalvavrik: Alright, thanks.

michalvavrik:
@mocenas the comment was for you

mocenas:
> Hi @jedla97 ,
> 
> I have tested it a bit and all that I needed to reproduce the issue was following patch (I'm pretty sure you can simplify that as well, but I don't have more time)
> 
> ```
> Subject: [PATCH] sse-native
> ---
> Index: http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/SseClient.java
> IDEA additional info:
> Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
> <+>UTF-8
> ===================================================================
> diff --git a/http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/SseClient.java b/http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/SseClient.java
> new file mode 100644
> --- /dev/null	(date 1699377305495)
> +++ b/http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/SseClient.java	(date 1699377305495)
> @@ -0,0 +1,18 @@
> +package io.quarkus.ts.http.advanced;
> +
> +import jakarta.enterprise.context.ApplicationScoped;
> +import jakarta.enterprise.event.Observes;
> +import jakarta.enterprise.event.Startup;
> +import jakarta.ws.rs.client.ClientBuilder;
> +import jakarta.ws.rs.client.WebTarget;
> +import jakarta.ws.rs.sse.SseEventSource;
> +
> +@ApplicationScoped
> +public class SseClient {
> +
> +    void callSomeMethod(@Observes Startup startup) {
> +        WebTarget target = ClientBuilder.newClient().target("http://localhost:8080/updates");
> +        var src = SseEventSource.target(target).build();
> +        System.out.println("I have a client " + src);
> +    }
> +}
> ```
> 
> with command `mvn clean verify -Dnative -Dit.test=HttpAdvancedIT -Dquarkus.platform.version=3.2.7.Final`
> 
> Please refactor it to your needs. Do you agree that it will cost us much less of execution time and it's easier to read? Please feel free to contradict. Thanks


@michalvavrik  I agree that your solution is simpler and faster, but I'm not sure I like it. 

1.  I though that goal of test coverage for backports is to have a test that will tell you if the issue is fixed or not (if this specific functionality works or not). your solution will just fail to start the tests and tell nothing - unless you know this exact issue, you'll have no idea why it failed (OK I can add a try-catch and print better exception message, but you get the point).
2. IMHO to verify that issue is fixed, there should be a test that has a pass run, proving that the features actually work. IMHO having just "module did not fail to start" is kind of bad proof that issue is indeed fixed.
3. IMHO new tests should not interfere with other ones. Your proposed solution, if failed, will cause entire http-advanced module to fail, because tested app will not start. And all other tests in it will just not run. That's why I've created a new module - to isolate the issue.
4. Due to bug description it requires picoli on client side. I don't know why it successfully fails in your example. But I wanted to move it to another module, so adding picoli will not interfere with other tests.

I'm new to issue-test-coverage this stuff so correct me if I'm wrong at my line of thinking.

http/http-sse/src/test/java/http/sse/HttpSseIT.java
mocenas: OK I will change this. Is it better to have it without tag, or can I add a link to the upstream issue to the tag? AFAIK there is no quarkus jira for this.

michalvavrik:
> I though that goal of test coverage for backports is to have a test that will tell you if the issue is fixed or not (if this specific functionality works or not)

Agreed, that's is why I spent time understanding a root cause of issue and principle how it works and created simplified reproducer. Issue has literally nothing to do with picocli.

> your solution will just fail to start the tests and tell nothing - unless you know this exact issue, you'll have no idea why it failed (OK I can add a try-catch and print better exception message, but you get the point).

My solution was POC for you, it's your job and I won't tell you how to do this. If I were to do this reproducer, I'd simply put what I added to startup observer to resource method and call it from a unit test, which tells you exactly what is tested

> IMHO to verify that issue is fixed, there should be a test that has a pass run, proving that the features actually work. IMHO having just "module did not fail to start" is kind of bad proof that issue is indeed fixed.

+1, please see answer above

> IMHO new tests should not interfere with other ones. Your proposed solution, if failed, will cause entire http-advanced module to fail, because tested app will not start. And all other tests in it will just not run. That's why I've created a new module - to isolate the issue.

no worry, if you put it to resource method, it interferes with nothing as you are bringing in no dependencies or application class changes, you can only add new resource class

> Due to bug description it requires picoli on client side. I don't know why it successfully fails in your example. But I wanted to move it to another module, so adding picoli will not interfere with other tests.

no it does not, I already explained it

http/http-sse/src/test/java/http/sse/HttpSseIT.java
michalvavrik: @mocenas this was more of a question, I really don't know. We can ask for @rsvoboda opinion, but as far as I'm concerned, there is nothing wrong on your approach. it just surprised me :-)

michalvavrik: > AFAIK there is no quarkus jira for this.

indeed

michalvavrik:
@mocenas one else thing - there is nothing wrong about your reproducer (which is adjusted upstream reproducer, right?), my argument is that resources are tight, my suggestion will add couple of microseconds to execution time, new module build in native will add couple of minutes on my workstation

mocenas:
@michalvavrik I've refactored the test to be in http-advanced module.

http/http-advanced/src/test/java/io/quarkus/ts/http/advanced/HttpSseIT.java
```diff
@@ -0,0 +1,32 @@
...
+public class HttpSseIT {
+
+    @QuarkusApplication
+    static RestService app = new RestService()
+            .withProperty("quarkus.oidc.enabled", "false")
+            .withProperty("quarkus.keycloak.policy-enforcer.enable", "false")
+            .withProperty("quarkus.keycloak.devservices.enabled", "false");
+    private static final String SSE_ERROR_MESSAGE = "java.lang.ClassNotFoundException: Provider for jakarta.ws.rs.sse.SseEventSource.Builder cannot be found";
+
+    @Test
```
michalvavrik, line 24: While I think it's nicely clear that you want to test `sse` as you have it in dedicated class, I'd still prefer if you added it to the `HttpAdvancedIT` because that way, application won't be re-build. There are miscellaneous test and if you choose fitting unit test method name, it will be clear enough. We can spare couple of minutes because `quarkus.oidc.enabled` is a build time property.

It's not irrational, because if you look how long our native CI is running, we need to have good reasons to add it as a separate class.

http/http-advanced/src/test/java/io/quarkus/ts/http/advanced/HttpSseIT.java
```diff
@@ -0,0 +1,32 @@
...
+import io.quarkus.test.services.QuarkusApplication;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@QuarkusScenario
+@Tag("https://github.com/quarkusio/quarkus/issues/36402")
+// this test should only fail on native
```
michalvavrik, line 14: I hope it never fails :-) No, seriously, if you want, you can only do `@EnabledOnNative` or you can just leave comment out. It's fine if you want to keep it.

http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/SseResource.java
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * Test runner assigns random ports, on which the app should run.
+     * Parse this port from the CLI and return it.
+     * If no parameter is specified, return the default (8080)
+     *
+     * @return port on which the application is running
+     */
+    private int getQuarkusPort() {
+        String value = System.getProperty("quarkus.http.port");
```
michalvavrik, line 69: please inject `HttpConfiguraton` and take port from `io.quarkus.vertx.http.runtime.HttpConfiguration#port`. that way, it won't ever be null (as you have a default value) and also, system properties are just one configuration source. we should never rely on them. You can also inject single property with `@ConfigProperty(name = "quarkus.http.port")`, choice is yours

http/http-advanced/src/test/java/io/quarkus/ts/http/advanced/HttpSseIT.java
michalvavrik: Now that I'm thinking about it, you probably wanted to avoid OCP tests. Well, I think it is positive to have this tested on OCP as well.

http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/SseResource.java
```diff
@@ -0,0 +1,90 @@
...
+    private String consumeSse() {
+        StringBuilder response = new StringBuilder();
+
+        /*
+         * Client connects to itself (to server endpoint running on same app),
+         * because for https://github.com/quarkusio/quarkus/issues/36402 to reproduce client must run on native app.
+         * Which cannot be done in test code itself.
+         * This method acts just as a client
+         */
+        WebTarget target = ClientBuilder.newClient().target("http://localhost:" + getQuarkusPort() + "/api/sse/server");
```
michalvavrik, line 45: `quarkus.http.host` will in production resolve to 0.0.0.0, I think it won't make difference here, but I don't fully understand why it exists otherwise. I'll let you decide.

mocenas: AFAIK setting `quarkus.http.host` to 0.0.0.0 means, that app will listen on all network interfaces. Which is not necessary here, as this app wants to connect to itself.

http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/SseResource.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+        return Integer.parseInt(value);
+    }
+
+    @GET
+    @Path("/server")
+    @Produces(MediaType.SERVER_SENT_EVENTS)
+    public void sendSseEvents(@Context SseEventSink eventSink) {
+        eventSink.send(createEvent("test234", "test"));
+        LockSupport.parkNanos(1_000_000_000L);
```
michalvavrik, line 81: I tried to find this in Quarkus upstream project or in its documentation, but in `io.quarkus.resteasy.test.sse.SseResource` (classic) I found this pattern

```
        SseBroadcaster sseBroadcaster = sse.newBroadcaster();

        sseBroadcaster.register(sink);
        sseBroadcaster.broadcast(sse.newEventBuilder().data("hello").build());
```
which I found repeatedly used in RESTEasy Reactive as well.

Are you certain it is necessary and wise? I really don't know that, so please provide a context if you want this pattern.

http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/SseResource.java
```diff
@@ -0,0 +1,90 @@
...
+            response.append("event: ").append(ev.getName()).append(" ").append(ev.readData());
+            response.append("\n");
+
+        }, thr -> {
+            response.append("Error in SSE, message: ").append(thr.getMessage()).append("\n");
+            response.append(Arrays.toString(thr.getStackTrace()));
+        });
+        updateSource.open();
+
+        LockSupport.parkNanos(2_000_000_000L);
```
michalvavrik, line 57: same question here in regards of parking as below, thx

http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/SseResource.java
michalvavrik: I'll trust you there.

michalvavrik (commented):
overall looks good, I need to leave now for a while, I'll review & merge this later today. only thing that really concerns me is whether we need dedicated test class (we have comment thread on that)

http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/SseResource.java
mocenas: This is a pattern from original reproducer. I didn't really thought of changing it.

http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/SseResource.java
mocenas: SSE client (event) is asynchronous, so client needs to wait until the event really comes before return result. We can use other synchronization tool, but this one is simple and effective.

http/http-advanced/src/test/java/io/quarkus/ts/http/advanced/HttpSseIT.java
mocenas: Moved SSE test to HttpAdvancedIT.

http/http-advanced/src/test/java/io/quarkus/ts/http/advanced/HttpSseIT.java
mocenas: I left this out.

http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/SseResource.java
michalvavrik: ok

http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/SseResource.java
michalvavrik: hmm

michalvavrik (approved):
lgtm, let see what CI says

PR #1520, opened 2023-11-15
Use OpenTelemetry extension in Build-tiime Analytics module as OpenTracing is deprecated and not released yet

michalvavrik:
### Summary

Fixes CI. https://github.com/quarkus-qe/quarkus-test-suite/pull/1519 follow-up

https://docs.quarkiverse.io/quarkus-smallrye-opentracing/dev/index.html says we should use OpenTelemetry. There is no deeper logic why this module used OpenTracing extension (like product requirement or RH ticket etc.), so here we can safely replace it with preferred version.

Please select the relevant options.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport
- [ ] New scenario (non-breaking change which adds functionality)
- [ ] This change requires a documentation update
- [ ] This change requires execution against OCP (use `run tests` phrase in comment)

### Checklist:
- [x] Methods and classes used in PR scenarios are meaningful
- [x] Commits are well encapsulated and follow [the best practices](https://cbea.ms/git-commit/)

PR #1349, opened 2023-08-05
Propagate max Java heap used during native image generation down to build time analytics app builder

michalvavrik:
### Summary

Native build fails with OOM as Github CI runners don't have that much of a memory. So we set maximum here https://github.com/quarkus-qe/quarkus-test-suite/pull/1346 but I didn't realize it is not propagated down to the application.

Please select the relevant options.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport
- [ ] New scenario (non-breaking change which adds functionality)
- [ ] This change requires a documentation update
- [ ] This change requires execution against OCP (use `run tests` phrase in comment)

### Checklist:
- [x] Methods and classes used in PR scenarios are meaningful
- [x] Commits are well encapsulated and follow [the best practices](https://cbea.ms/git-commit/)

PR #1523, opened 2023-11-16
Fix daily native build by limiting Java heap used and spread load in matrix

michalvavrik:
### Summary

- Currently native analytics daily build fails over OOM during native executable build, it started to happen after #1520 but I don't expect there is something wrong about OpenTelemetry, it might be larger than OpenTracing extension, but we definitely had problemw with memory in this module before: https://github.com/quarkus-qe/quarkus-test-suite/pull/1349
- I checked and Security profile takes least on average (~ 1 h 20 min) and root modules with spring takes up to for 4 h https://github.com/quarkus-qe/quarkus-test-suite/actions/runs/6765196692/job/18392571155 but it jumps heavily with hours in difference). IMO we should spread load by moving Spring modules to Security modules

```
2023-11-16T03:28:34.0733613Z 03:28:34,030 INFO   - 4.44GB of memory (28.5% of 15.61GB system memory, set via '-Xmx5g')
2023-11-16T03:28:34.0734917Z 03:28:34,030 INFO   - 4 thread(s) (100.0% of 4 available processor(s), determined at start)
2023-11-16T03:35:32.9806773Z 03:35:32,969 INFO  Terminating due to java.lang.OutOfMemoryError: GC overhead limit exceeded
```

IIUC https://www.graalvm.org/latest/reference-manual/native-image/optimizations-and-performance/MemoryManagement/ it says that GC also needs 2x of the maximum heap size is the worst case and https://quarkus.io/guides/native-reference says minimum allowed is 4 GB.

So this PR might slower down native CI execution, but GH runners seems weak, I suggest to try it and optimize based on results.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -92,9 +92,9 @@ jobs:
       matrix:
         java: [ 11 ]
         image: [ "ubi-quarkus-graalvmce-builder-image:jdk-17", "ubi-quarkus-mandrel-builder-image:23.0-java17" ]
-        profiles: [ "root-modules,spring-modules",
+        profiles: [ "root-modules",
                    "http-modules",
-                   "security-modules",
+                   "security-modules,spring-modules",
                     "sql-db-modules -pl env-info,sql-db/hibernate,sql-db/sql-app,sql-db/sql-app-compatibility,sql-db/multiple-pus,sql-db/panache-flyway,sql-db/hibernate-reactive",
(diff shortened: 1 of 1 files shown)


michalvavrik:
PS: this CI isn't really testing this PR :-) but let's wait at least for validation (thought that's not testing this PR either)...

jsmrcka (approved):
Well, it looks like it's worth a shot. Thanks for the analysis.

michalvavrik:
> Well, it looks like it's worth a shot. Thanks for the analysis.

Yeah, it's not worth testing it because native run execution time is really volatile, IMO we need to take a bit of guessing and wait.

PR #1531, opened 2023-11-20
Add test coverage for grpc GlobalInterceptor method annotation

jedla97:
### Summary

Add test coverage for https://github.com/quarkusio/quarkus/issues/21358. Testing both server and client interceptor. 

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/GrpcClientInterceptors.java
+++ b/http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/GrpcClientInterceptors.java
@@ -1,0 +1,55 @@
+package io.quarkus.ts.http.advanced;
+
+import static io.grpc.Metadata.ASCII_STRING_MARSHALLER;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.inject.Produces;
+
+import io.grpc.CallOptions;
+import io.grpc.Channel;
+import io.grpc.ClientCall;
+import io.grpc.ClientInterceptor;
+import io.grpc.ForwardingClientCall;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+import io.quarkus.grpc.GlobalInterceptor;
+
+class GrpcClientInterceptors {
+    public static Metadata.Key<String> CLIENT_METHOD = Metadata.Key.of("client-method-target", ASCII_STRING_MARSHALLER);
+    public static Metadata.Key<String> CLIENT_CLASS = Metadata.Key.of("client-class-target", ASCII_STRING_MARSHALLER);
+
+    @GlobalInterceptor
+    @ApplicationScoped
+    static class ClassTarget extends Base {
+    }
+
+    static class MethodTarget extends Base {
+    }
+
+    static class Producer {
+        @GlobalInterceptor
+        @Produces
+        MethodTarget methodTarget() {
+            return new MethodTarget();
+        }
+    }
+
+    abstract static class Base implements ClientInterceptor {
+        @Override
+        public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(MethodDescriptor<ReqT, RespT> method, CallOptions options,
+                Channel next) {
+            String interceptedTarget = getClass().getName();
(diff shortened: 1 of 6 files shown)


jedla97:
run tests

jedla97:
run tests

http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/GrpcResource.java
```diff
@@ -1,18 +1,24 @@
...
+import jakarta.ws.rs.core.Response;
 
 import io.quarkus.example.GreeterGrpc;
+import io.quarkus.example.HelloReply;
 import io.quarkus.example.HelloRequest;
 import io.quarkus.grpc.GrpcClient;
 
 @Path("/grpc")
 public class GrpcResource {
+    static Set<String> invoked = new HashSet<>();
```
michalvavrik, line 21: My impression is that this approach is spreading from previous PRs, but these kind of data should be passed along with request / response. Only reason why you can do it is because you call same gRPC server. But in production you will never use gRPC client to call anything inside very same application.

I know this can be dogmatic, but I'd like to follow approach closer to what users will do, especially because this way indirectly we test other things, not just isolated feature.

I'll leave below suggestions how to do this.

http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/GrpcServerInterceptors.java
```diff
@@ -0,0 +1,37 @@
...
+        MethodTarget methodTarget() {
+            return new MethodTarget();
+        }
+    }
+
+    abstract static class Base implements ServerInterceptor {
+        @Override
+        public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(ServerCall<ReqT, RespT> call, Metadata metadata,
+                ServerCallHandler<ReqT, RespT> next) {
+            GrpcResource.invoked.add(getClass().getName());
```
michalvavrik, line 33: check it's duplicated context https://quarkus.io/guides/duplicated-context#distinguish-between-root-and-duplicated-contexts (illegal state if it isn't) and put it to local data. Then access it in your endpoint and put it to response

```
io.quarkus.vertx.core.runtime.context.VertxContextSafetyToggle#validateContextIfExists
io.vertx.core.Context#putLocal
```

Note: don't know gRPC that well, you can maybe go with metadata instead? your choice

http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/GrpcClientInterceptors.java
```diff
@@ -0,0 +1,38 @@
...
+        MethodTarget methodTarget() {
+            return new MethodTarget();
+        }
+    }
+
+    abstract static class Base implements ClientInterceptor {
+        @Override
+        public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(MethodDescriptor<ReqT, RespT> method, CallOptions options,
+                Channel next) {
+            GrpcResource.invoked.add(getClass().getName());
```
michalvavrik, line 34: add new headers, one example I found in a second was `io.quarkus.grpc.examples.hello.HeaderCollectingInterceptor` looks like it comply your use case, but you will know better

michalvavrik (changes requested):

http/http-advanced/src/main/java/io/quarkus/ts/http/advanced/GrpcServerInterceptors.java
michalvavrik:
```
            VertxContextSafetyToggle.validateContextIfExists("", "");
            final var ctx = Vertx.currentContext();
            ctx.putLocal(TokenCredential.class.getName(), token);
```

jedla97:
Updated it with more realistic usecase. And adding tag to RH Quarkus jira.

Now it's workflow `intercept client` -> `intercept server` -> `server read added header by client` -> `server add to Context` -> `context is propagated to service`.

This testing global interceptor for class and methods which each intercepted call (method and class) have own constant used to pass client intercepted data to next step.

michalvavrik (approved):
Good research, nice job.

PR #1524, opened 2023-11-16
Add Redis cache tests and bump Redis version

jedla97:
### Summary

Hi, this PR main objective was cover https://github.com/quarkusio/quarkus/issues/35680. As it was need standalone module I added some test which was overlapping between caffeine cache and spring cache.  Test to cover 35680 is `quteShouldThrowError`. Tested manually the redis cache with rest of the api and I saw it cached in Redis (used `redis-cli keys "*"`).

In standalone commit is update of Redis version (Quarkus dev mode using latest).  

Please select the relevant options.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport
- [x] New scenario (non-breaking change which adds functionality)
- [ ] This change requires a documentation update
- [x] This change requires execution against OCP (use `run tests` phrase in comment)

### Checklist:
- [x] Methods and classes used in PR scenarios are meaningful
- [x] Commits are well encapsulated and follow [the best practices](https://cbea.ms/git-commit/)

PR #1533, opened 2023-11-20
Move cache extensions from root modules to spread load evenly in matrix (especially as way to mitigate native build OOM in dailys)

michalvavrik:
### Summary

I don't know what to do about dailys - https://github.com/quarkus-qe/quarkus-test-suite/actions/workflows/daily.yaml it seems to me like either we need to introduce some cleaning of resources, or we need to spread load to other runners. I already had 2 experimental attempts #1529 and #1523 without success.

But what is worse, new addition is coming right into root modules where we fight resources: #1524

I propose to separate cache modules as that way we can easily couple them with other profiles in matrix. Native executable build time is very volatile, but building HTTP modules still takes less than root-modules. Sometimes significantly less, see https://github.com/quarkus-qe/quarkus-test-suite/actions/runs/6819612674/job/18547594454 2 hours less.

I'll follow-up with Jenkins updates if this goes in.

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,7 +100,7 @@ jobs:
             echo "Running modules: ${MODULES_ARG}"
             echo "MODULES_MAVEN_PARAM=[\" -pl ${MODULES_ARG} -Dall-modules\"]" >> $GITHUB_OUTPUT
           else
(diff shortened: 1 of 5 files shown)


jedla97 (approved):
Thanks, LGTM!

Issue #1580, opened 2023-12-18
Tests with Keycloak container should not be started at Keycloak DEV mode


michalvavrik:
We keep repeating this pattern

https://github.com/quarkus-qe/quarkus-test-suite/blob/c92ab509aa30a6b7b17ce67578ac2e0a841d2e42/security/keycloak-authz-reactive/src/test/java/io/quarkus/ts/security/keycloak/authz/reactive/KeycloakAuthzSecurityReactiveIT.java#L13

mostly because we don't have configured HTTPS for all origins, but no users will actually do that in production, we need to have communication between Quarkus OIDC and Keycloak via secured channel and only start Keycloak in a production mode.

PR #1581, opened 2023-12-19
Keycloak fixes and Quarkus FW bump to 1.4.0.Beta5

michalvavrik:
### Summary

Main goal of this PR is to fix OCP failures related to the fact that Keycloak auth server URL has changed in our FW: https://github.com/quarkus-qe/quarkus-test-framework/pull/960

- `OpenShiftRhSsoOidcMtlsIT` will require much more thinkering, at the very least I hit that config map name is created with illegal chars when resource is in nested directory. But I think it will be even more difficult, I'll address it later this week.
- I don't want to address all that repeated stuff of KC commands till  https://github.com/quarkus-qe/quarkus-test-suite/issues/1580 is done
- I also think some of these Keycloak starting commands are unnecessary at some of scenarios (I replaced one accidentally and it still works), let's address it when we get rid of `start-dev` because it all seems like a one big workaround for proper solution to me
- `OidcRestClientIT` failure (both baremetal and OCP) goes down to the https://github.com/quarkusio/quarkus/pull/37788, I hope it's going to be merged today as that reviewer is active in other Quarkus PRs
- new KC image does not accept args joined by space, we need to pass them one by one (which is correct way to do it anyway)

- [x] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/security/keycloak-authz-reactive/src/test/java/io/quarkus/ts/security/keycloak/authz/reactive/KeycloakAuthzSecurityReactiveIT.java
+++ b/security/keycloak-authz-reactive/src/test/java/io/quarkus/ts/security/keycloak/authz/reactive/KeycloakAuthzSecurityReactiveIT.java
@@ -1,5 +1,9 @@
 package io.quarkus.ts.security.keycloak.authz.reactive;
 
+import static io.quarkus.test.bootstrap.KeycloakService.DEFAULT_REALM;
+import static io.quarkus.test.bootstrap.KeycloakService.DEFAULT_REALM_BASE_PATH;
+import static io.quarkus.test.bootstrap.KeycloakService.DEFAULT_REALM_FILE;
+
 import io.quarkus.test.bootstrap.KeycloakService;
 import io.quarkus.test.bootstrap.RestService;
 import io.quarkus.test.scenarios.QuarkusScenario;
@@ -10,8 +14,8 @@ import io.quarkus.test.services.QuarkusApplication;
 public class KeycloakAuthzSecurityReactiveIT extends BaseAuthzSecurityReactiveIT {
 
     //TODO Remove workaround after Keycloak is fixed https://github.com/keycloak/keycloak/issues/9916
-    @KeycloakContainer(command = { "start-dev --import-realm --hostname-strict=false" })
-    static KeycloakService keycloak = new KeycloakService("/keycloak-realm.json", REALM_DEFAULT, "/realms")
+    @KeycloakContainer(command = { "start-dev", "--import-realm" })
+    static KeycloakService keycloak = new KeycloakService(DEFAULT_REALM_FILE, DEFAULT_REALM, DEFAULT_REALM_BASE_PATH)
             .withProperty("JAVA_OPTS", "-Dcom.redhat.fips=false");
 
     @QuarkusApplication
--- a/security/keycloak-oauth2/src/test/java/io/quarkus/ts/security/keycloak/oauth2/OpenShiftRhSsoOauth2SecurityIT.java
+++ b/security/keycloak-oauth2/src/test/java/io/quarkus/ts/security/keycloak/oauth2/OpenShiftRhSsoOauth2SecurityIT.java
@@ -1,12 +1,16 @@
 package io.quarkus.ts.security.keycloak.oauth2;
 
+import static io.quarkus.test.bootstrap.KeycloakService.DEFAULT_REALM;
+import static io.quarkus.test.bootstrap.KeycloakService.DEFAULT_REALM_BASE_PATH;
+import static io.quarkus.test.bootstrap.KeycloakService.DEFAULT_REALM_FILE;
+
 import org.junit.jupiter.api.condition.DisabledIfSystemProperty;
 import org.junit.jupiter.api.condition.EnabledIfSystemProperty;
 
 import io.quarkus.test.bootstrap.KeycloakService;
 import io.quarkus.test.bootstrap.RestService;
 import io.quarkus.test.scenarios.OpenShiftScenario;
-import io.quarkus.test.services.Container;
+import io.quarkus.test.services.KeycloakContainer;
 import io.quarkus.test.services.QuarkusApplication;
 
 @OpenShiftScenario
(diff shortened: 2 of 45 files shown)


michalvavrik:
run tests

michalvavrik:
run tests

security/keycloak-authz-reactive/src/test/java/io/quarkus/ts/security/keycloak/authz/reactive/KeycloakAuthzSecurityReactiveIT.java
```diff
@@ -10,8 +14,8 @@
 public class KeycloakAuthzSecurityReactiveIT extends BaseAuthzSecurityReactiveIT {
 
     //TODO Remove workaround after Keycloak is fixed https://github.com/keycloak/keycloak/issues/9916
-    @KeycloakContainer(command = { "start-dev --import-realm --hostname-strict=false" })
-    static KeycloakService keycloak = new KeycloakService("/keycloak-realm.json", REALM_DEFAULT, "/realms")
+    @KeycloakContainer(command = { "start-dev", "--import-realm" })
```
jedla97, line 17: Is there reason why `--hostname-strict=false` is  removed when in other cases it stay.

michalvavrik:
Remaining OCP / baremetal failures are failing also without this PR, therefore they are not result of this PR.

security/keycloak-oauth2/src/test/java/io/quarkus/ts/security/keycloak/oauth2/OpenShiftRhSsoOauth2SecurityIT.java
```diff
@@ -1,24 +1,25 @@
...
 @DisabledIfSystemProperty(named = "ts.arm.missing.services.excludes", matches = "true", disabledReason = "https://github.com/quarkus-qe/quarkus-test-suite/issues/1145")
 @EnabledIfSystemProperty(named = "ts.redhat.registry.enabled", matches = "true")
 public class OpenShiftRhSsoOauth2SecurityIT extends BaseOauth2SecurityIT {
 
-    static final int KEYCLOAK_PORT = 8080;
-
-    @Container(image = "${rhsso.image}", expectedLog = "Http management interface listening", port = KEYCLOAK_PORT)
-    static KeycloakService keycloak = new KeycloakService(REALM_DEFAULT)
-            .withProperty("SSO_IMPORT_FILE", "resource::/keycloak-realm.json");
+    @KeycloakContainer(command = { "start-dev", "--import-realm" }, image = "${rhbk.image}")
```
jedla97, line 21: Why some of the command `KeycloakContainers` contains contain more commands that these two. When they were the same with using `Container`.

jedla97 (commented):
I looked at it. It's look good just two comments to clarify.

security/keycloak-authz-reactive/src/test/java/io/quarkus/ts/security/keycloak/authz/reactive/KeycloakAuthzSecurityReactiveIT.java
michalvavrik: yes, I wrote about this (vaguely) in the PR description, I changed it accidentally and it made no difference, so I suppose it doesn't really matter. If it doesn't cause test failure I don't think it needs to be there. Good catch, it was mistake. If it is somehow important for you, I'll change it in next PR when fixing mTLS.

security/keycloak-oauth2/src/test/java/io/quarkus/ts/security/keycloak/oauth2/OpenShiftRhSsoOauth2SecurityIT.java
michalvavrik: I removed `.withProperty("SSO_IMPORT_FILE", "resource::/keycloak-realm.json");` so I want to import it from command. I can't answer it in general, if you deem particular change wrong, please comment on the very change and I'll check.

security/keycloak-authz-reactive/src/test/java/io/quarkus/ts/security/keycloak/authz/reactive/KeycloakAuthzSecurityReactiveIT.java
jedla97: If it's work I'm fine with it. Hope that this not cause the problem in future.

michalvavrik: > If it's work I'm fine with it. Hope that this not cause the problem in future.

IMO tests need to be idempotent, otherwise our job will become very dangerous

michalvavrik: we will see other integration runs like Windows and Podman... I don't expect there should be difference.

security/keycloak-oauth2/src/test/java/io/quarkus/ts/security/keycloak/oauth2/OpenShiftRhSsoOauth2SecurityIT.java
jedla97: Just that somewhere `@Conatiner` change to this
```
@KeycloakContainer(command = { "start-dev", "--import-realm", "--hostname-strict=false",
            "--features=token-exchange" }, image = "${rhbk.image}")
```
and somewhere to this
```
@KeycloakContainer(command = { "start-dev", "--import-realm" }, image = "${rhbk.image}")
```

But this was answered in other comment so now there is probably no need for some of these commands.
But as it was written in PR description let's solve these problem as the `start-dev` is solved.

jedla97 (approved):
Comments were explained and the PR it's look fine.

security/keycloak-oauth2/src/test/java/io/quarkus/ts/security/keycloak/oauth2/OpenShiftRhSsoOauth2SecurityIT.java
michalvavrik: just note - I won't pretend I couldn't make mistake, but this is how I worked:

- if I remove import env var, I need to add `--import-realm`
- If I change RHSSO for RHBK it is change from Wildfly to Quarkus based KC, so I followed what is done in that very project for non-OCP tests. so typically if there were `--features=token-exchange` for baremetal tests I just added it as well. I didn't try to think whether it needs to be in baremetal tests at the first place.

PR #1585, opened 2023-12-21
Test MapStruct together with Quarkus

michalvavrik:
### Summary

Add test coverage for basic MapStruct features used with Quarkus. Integration point with Quarkus is mostly about CDI, as all other should work from annotation processor, but it's good to test common stuff because you never know.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/javaee-like-getting-started/src/test/java/io/quarkus/ts/javaee/gettingstarted/MapStructIT.java
+++ b/javaee-like-getting-started/src/test/java/io/quarkus/ts/javaee/gettingstarted/MapStructIT.java
@@ -1,0 +1,101 @@
+package io.quarkus.ts.javaee.gettingstarted;
+
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.restassured.RestAssured;
+
+@QuarkusScenario
+public class MapStructIT {
+
+    @Test
+    public void testInjectionWithJakartaCdiComponentModel() {
+        // tests:
+        // - Jakarta CDI component model (mapper is @ApplicationScoped bean)
+        // - mapping of beans to beans
+        // - getters used for nested properties
+        RestAssured
+                .get("/mapper/injected-jakarta-cdi")
+                .then()
+                .statusCode(200)
+                .body("greeting", Matchers.is("Hello Sebastian from Maximillian"));
+    }
+
+    @Test
+    public void testInjectionWithJakartaComponentModel() {
+        // tests:
+        // - Jakarta component model (mappers are named beans)
+        // - mapper decorator
+        // - default interface method is used to convert types (FarewellPartOne -> FarewellPartOneDTO)
+        // - @Mapping Java expression works
+        // - using records
+        // - injection of original (not decorated mapper) works as described in MapStruct documentation
+        // - injection of other mappers inside decorator works
+        // - @Mapping can use constant as value source
+        RestAssured
+                .given()
+                .get("/mapper/injected-jakarta-happy-impl")
+                .then()
+                .statusCode(200)
+                .body("partOne.data", Matchers.is("original_happy"))
+                .body("partTwo", Matchers.is("1"));
+        RestAssured
+                .given()
+                .get("/mapper/injected-jakarta-sad-impl")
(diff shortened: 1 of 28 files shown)


javaee-like-getting-started/src/test/java/io/quarkus/ts/javaee/gettingstarted/MapStructIT.java
```diff
@@ -0,0 +1,55 @@
...
+                .given()
+                .get("/mapper/injected-jakarta-default-method")
+                .then()
+                .statusCode(200)
+                .body("partOne.data", Matchers.is("original_default"))
+                .body("partTwo", Matchers.is("3"));
+    }
+
+    @Test
+    public void testInjectionWithCdiComponentModel() {
```
rsvoboda, line 47: I'm bit confused by method name because I didn't notice any injection for `/mapper/no-cdi` and `SeriousTalkMapper` uses the default componentModel.

```
@Mapper
public interface SeriousTalkMapper {
```
https://github.com/mapstruct/mapstruct/blob/main/core/src/main/java/org/mapstruct/Mapper.java#L158

javaee-like-getting-started/src/test/java/io/quarkus/ts/javaee/gettingstarted/MapStructIT.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.restassured.RestAssured;
+
+@QuarkusScenario
+public class MapStructIT {
+
```
rsvoboda, line 11: I would appreciate some context about the scenarios because this is first MapStruct addition here and having some context could help others with faster understanding.

I was thinking about JavaDoc details for test methods. Another option would be text in README, but I think having details close to the tests is the best.

rsvoboda (changes requested):

javaee-like-getting-started/src/test/java/io/quarkus/ts/javaee/gettingstarted/MapStructIT.java
michalvavrik: sure, name should be `testInjectionWithDefaultComponentModel` or something like that

rsvoboda: but where is injection used?

michalvavrik: there is no injection, it makes sense to test default model in native mode because it works based on static variables and you want to know there is no initialization issue. URL says it `/mapper/no-cdi`, so I just copied previous name.

javaee-like-getting-started/src/test/java/io/quarkus/ts/javaee/gettingstarted/MapStructIT.java
michalvavrik: I will add few comments. I think it tests what it tests, there is no requirement on list of features and I just tried to cover as much as possible in not so much code.

javaee-like-getting-started/src/test/java/io/quarkus/ts/javaee/gettingstarted/MapStructIT.java
michalvavrik: but I can't see `injected-cdi` called, I think I accidentally rewritten it to this "no-cdi", but I remember testing it, I'll try to remember how and add it.

michalvavrik: done

javaee-like-getting-started/src/test/java/io/quarkus/ts/javaee/gettingstarted/MapStructIT.java
michalvavrik: done

rsvoboda (approved):

PR #1599, opened 2024-01-08
Non json resteasy reactive

jcarranzan:
### Summary

Create test cases that evolve non-JSON payload in REST endpoints as described in the Jira task:
https://issues.redhat.com/browse/QQE-343

1. XML scenario:
Scenario: Verify that the endpoint properly consumes and produces XML payloads.

- Check the endpoint that returns a valid XML.
- Send a request with an XML payload to the endpoint.
- Validate the response.
- Ensure the response's content matches the expected from the XML structure.

2. YAML scenario

- Check the endpoint that returns a Yaml from a Yaml file.
- Send a request with a Yaml payload from a file to an endpoint.
- Validate the response 
- Ensure the response's content matches the expected from the YAML content.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/CityResource.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        String responseXML = serializer.toXML(cityListDTO);
+
+        return responseXML;
+    }
+
+    @POST
+    @Path("/cityYaml")
+    @Produces(MediaType.APPLICATION_JSON)
```
michalvavrik, line 83: Why is `handleYamlPostRequest` producing application/json?

jcarranzan: Why not? I mean the request already uses a YAML payload, and also in the @GET we already checked that it could be @Produce a Yaml here:    [ @Produces(APPLICATION_YAML)](https://github.com/quarkus-qe/quarkus-test-suite/blob/7db3158f3b330b551335d884f02c8bb0c6cc7f34/http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/CityResource.java#L100)

michalvavrik: ok

mjurc: Tbh it's kinda confusing at first gaze - why not just remove `Yaml` from the path? If people come back to this test in the future, they'll be scratching their head too :]

mjurc (changes requested):
Are you planning to add the image payload test too btw?

http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/CityResource.java
michalvavrik: truth be told, I didn't want to argue, but I think name should reflect that yaml is only consumed, not produced. like `handleConsumedYamlPostRequest` (just better :-)).

I also think it is rather rare to consume YAML and produce JSON and there is no scenario AFAICT that tests both consume/produce YAML, but I'll respect @jcarranzan preference.

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/NonJsonPayLoadIT.java
```diff
@@ -0,0 +1,176 @@
...
+            "    body:\n" +
+            "      cities:\n" +
+            "        - name: Tokio\n" +
+            "          country: Japan\n";
+
+    private final static String EXPECTED_RESPONSE_BODY_YAML_WINDOWS_OS = "- description: \"Send a request with a YAML payload to the endpoint, including cities data\"\r\n"
+            +
+            "  steps:\r\n" +
+            "    - action: \"POST\"\r\n" +
+            "      endpoint: \"/media-type/yaml\"\r\n" +
```
michalvavrik, line 72: this is hard to read, please use string block for all these long strings

michalvavrik: hmm, unless you need to backport it, but I don't think you do

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/NonJsonPayLoadIT.java
```diff
@@ -0,0 +1,176 @@
...
+        List<Object> cities = jsonPath.getList("expectedResponse.body.cities");
+        assertEquals("[{name=Tokio, country=Japan}]", cities.get(0).toString());
+    }
+
+    private String readYamlFile(String filePath) throws IOException {
+        Path path = Paths.get(filePath);
+        return new String(Files.readAllBytes(path));
+    }
+
+    private String convertCityListDTOToXml(CityListDTO cityListDTO) throws JAXBException {
```
michalvavrik, line 164: why not leave it to RestAssured? here https://www.james-willett.com/rest-assured-serialization-with-json-and-xml/#xml-serialization it looks like XML serialization should work.

http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/CityListWrapperSerializer.java
```diff
@@ -0,0 +1,53 @@
...
+import java.io.StringWriter;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.xml.bind.JAXBContext;
+import jakarta.xml.bind.JAXBException;
+import jakarta.xml.bind.Marshaller;
+import jakarta.xml.bind.Unmarshaller;
+import jakarta.xml.bind.annotation.XmlRootElement;
+
+import org.slf4j.Logger;
```
michalvavrik, line 13: we use JBOSS logger, right?

http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/CityResource.java
```diff
@@ -0,0 +1,119 @@
...
+    @Consumes("application/x-yaml")
+    public Object handleYamlPostRequest(@RequestBody final String yamlPayload) throws IOException {
```
michalvavrik, line 85: If this is `string`, there is no serialization in place, how does this actually test YAML content type consumption if you do serialization/deserialization yourself?

Please let me give you example: change `@Consumes` to JSON  or plain text and let me know if there is any difference. Thank you

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/NonJsonPayLoadIT.java
```diff
@@ -0,0 +1,176 @@
...
+    public void testExpectedYamlResponse() {
+
+        final Response response = app.given()
+                .contentType("application/yaml")
+                .get("/city/getYamlFile")
+                .then()
+                .statusCode(HttpStatus.SC_OK)
+                .extract().response();
+        if (OS.current() == OS.WINDOWS) {
+            assertThat(response.asPrettyString(), is(EXPECTED_RESPONSE_BODY_YAML_WINDOWS_OS));
```
michalvavrik, line 131: OS version :-) you prefer this over removing white spaces? you won't need to do that for response part if you avoid pretty string.

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/NonJsonPayLoadIT.java
```diff
@@ -0,0 +1,176 @@
...
+        String unescapedXml = doc.text();
+        assertThat(unescapedXml, equalTo(EXPECTED_XML_PAYLOAD));
+    }
+
+    @Test
+    public void testExpectedYamlResponse() {
+
+        final Response response = app.given()
+                .contentType("application/yaml")
+                .get("/city/getYamlFile")
```
michalvavrik, line 126: I'd like to point out that you don't actually test YAML. You test that you can send String from endpoint, consume string and because you manually serialized DTO to YAML, it is YAML.  The condition is always true because you do it manually.

http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/CityResource.java
jcarranzan: Yes, looking at it again, you're right I will tweak it to change the path and the name to `handleConsumedYamlPostRequest`

jcarranzan:
> Are you planning to add the image payload test too btw?

Yes, I will push it today if there are no setbacks.

http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/CityListWrapperSerializer.java
jcarranzan: Oh, I see, in vert.x test suite we used to use slf4j, changing it to JBOSS...

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/NonJsonPayLoadIT.java
jcarranzan: Yes, I'll use string block better.

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/NonJsonPayLoadIT.java
jcarranzan: good example, it works.

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/NonJsonPayLoadIT.java
jcarranzan: I removed it.

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/NonJsonPayLoadIT.java
michalvavrik: `testExpectedYamlResponse` is acceptable now. I happen to disagree

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/NonJsonPayLoadIT.java
```diff
@@ -0,0 +1,163 @@
...
+                .statusCode(200).extract().response();
+        JsonPath jsonPath = response.jsonPath();
+
+        List<Object> cities = jsonPath.getList("expectedResponse.body.cities");
+        assertEquals("[{name=Tokio, country=Japan}]", cities.get(0).toString());
+    }
+
+    @Test
+    public void testGetImage() {
+        byte[] imageData = RestAssured.given().get("/city/wolf_moon_howling.jpg").asByteArray();
```
michalvavrik, line 126: I'd prefer Howlin Wolf :/

No, now seriously, this image is not present in git project, therefore this test must always fail.

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/NonJsonPayLoadIT.java
```diff
@@ -0,0 +1,163 @@
...
+        assertThat(imageData, notNullValue());
```
michalvavrik, line 127: I apologize for my gruelling review, but I'm worried this could be true even when byte array is empty? Anyway, only way to assure it works properly is to compare actual byte arrays.

Maybe I am missing something, please let me know if you disagree.

michalvavrik: ha! exactly in a fashion you are doing it in the next test would be great!

http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/CityResource.java
```diff
@@ -0,0 +1,183 @@
...
+            cityList.add(new City("Brno", "Czech Republic"));
+            cityList.add(new City("Zaragoza", "Spain"));
+        }
+        return new CityListWrapper(cityList);
+    }
+
+    @POST
+    @Consumes(MediaType.APPLICATION_XML)
+    @Produces(value = MediaType.APPLICATION_XML)
+    public String createCity(String xml) {
```
michalvavrik, line 64:
```suggestion
    public String createCity(CityListDTO dto) {
```

http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/CityResource.java
```diff
@@ -0,0 +1,184 @@
...
+    @POST
+    @Path("/cities")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes("application/x-yaml")
+    public Object handleConsumedYamlPostRequest(@RequestBody final String yamlPayload) throws IOException {
+
+        LOG.info("Received YAML payload: " + yamlPayload);
+
+        try {
+            return objectMapper.readValue(yamlPayload, Object.class);
```
michalvavrik, line 97: This looks like a smoke test for `YAMLMapper` as deserializing it as object is lenient. While I think it's nice to know that the `YAMLMapper` works in native, I don't think this is how users should write it. 

Microservice won't have one endpoint, right? It depends on use case, but more often than not will endpoints consume same content type (at least we can say more than one)? If your answer is yes, than having boilerplate in every endpoint method doesn't make sense to me. That is why I think users should consume DTOs unless in special cases. I think you test transfer of string and if you didn't deserialize it with YamlMapper, it could as well be JSON etc. and the test would pass with same consumed content type.

In short: there is not a single endpoint that consumes DTO and I disagree with it.

jcarranzan: Well, in fact with my last changes, now there is an endpoint that consumes DTO https://github.com/quarkus-qe/quarkus-test-suite/blob/17584497b1ca70f5a5382d8c3dd363ab9f7fafcf/http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/CityResource.java#L65

jcarranzan: oh srry , this  @Consumes(MediaType.APPLICATION_XML)..., ok I will try to think it around to change it.

jcarranzan: With the last changes I think this thread we could mark it as resolved, wdyt @michalvavrik ?

http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/YamlProvider.java
```diff
@@ -0,0 +1,66 @@
...
+import jakarta.ws.rs.ext.MessageBodyWriter;
+import jakarta.ws.rs.ext.Provider;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+
+import io.quarkus.logging.Log;
+
+@Provider
+@Consumes("application/yaml")
```
michalvavrik, line 23: interesting, so given you didn't specify `@Produces`  this will be applied to all content types? Are you sure this is never applied to `public CityListDTO getCities() {`?

michalvavrik (commented):
LGTM if you can answer well my comment, thanks

http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/YamlProvider.java
jcarranzan: Well, it will be applied to all resource methods that consume "application/yaml".
If a method resource @produce another different content type , it takes precedence over the general provider.
In the case that you mentioned,it is:
```
@Produces(value = MediaType.APPLICATION_XML)
    public CityListDTO getCities() {
```

michalvavrik: The `MessageBodyWriter` javadoc only speaks about `produces`, not `consumes` - https://jakarta.ee/specifications/platform/8/apidocs/javax/ws/rs/ext/messagebodywriter So I don't say you are not right, but I'm not sure if this is expected behavior.

michalvavrik (approved):

http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/YamlProvider.java
jcarranzan: added` @Produces("application/yaml")` to the provider to be more consistent. Thanks @michalvavrik .

mjurc (approved):

mjurc (changes requested):
@jcarranzan thanks for the coverage, code LGTM, but please squash the commits and update the commit message to be more helpful

PR #1616, opened 2024-01-17
Add test coverage for containerResponseFilter issue

mocenas:
### Summary

Adding test coverage for issue https://github.com/quarkusio/quarkus/issues/31024.

For issue to manifest we need a quarkus app without resteasy dependency. It is probably cleaner and easier to maintain, to have a new module for this, that creating some system to exclude this dependency for some test.  It would require a new build of quarkus app anyway, so this approach should not bring any more performance drawbacks.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/http/rest-client-reactive-vanilla/src/main/java/io/quarkus/ts/http/restclient/reactive/RestCallerService.java
+++ b/http/rest-client-reactive-vanilla/src/main/java/io/quarkus/ts/http/restclient/reactive/RestCallerService.java
@@ -1,0 +1,30 @@
+package io.quarkus.ts.http.restclient.reactive;
+
+import java.net.URI;
+
+import jakarta.annotation.PostConstruct;
+
+import org.eclipse.microprofile.rest.client.RestClientBuilder;
+
+import io.quarkus.runtime.Startup;
+
+/**
+ * This class is used to reproduce issue https://github.com/quarkusio/quarkus/issues/31024
+ * Also requires {@link UselessRestApi} and {@link VersionHeaderFilter} for this
+ */
+@Startup
+public class RestCallerService {
+
+    @PostConstruct
+    void init() {
+        callRest();
+    }
+
+    void callRest() {
+        RestClientBuilder builder = RestClientBuilder.newBuilder()//
+                .baseUri(URI.create("localhost"));
+
+        // API needs to be created for issue to manifest
+        UselessRestApi api = builder.build(UselessRestApi.class);
+    }
+}
--- a/http/rest-client-reactive-vanilla/src/main/java/io/quarkus/ts/http/restclient/reactive/VersionHeaderFilter.java
+++ b/http/rest-client-reactive-vanilla/src/main/java/io/quarkus/ts/http/restclient/reactive/VersionHeaderFilter.java
@@ -1,0 +1,30 @@
+package io.quarkus.ts.http.restclient.reactive;
+
+import java.io.IOException;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.ws.rs.container.ContainerRequestContext;
+import jakarta.ws.rs.container.ContainerResponseContext;
+import jakarta.ws.rs.container.ContainerResponseFilter;
+import jakarta.ws.rs.container.PreMatching;
+import jakarta.ws.rs.ext.Provider;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+/**
+ * Used for {@link RestCallerService}
+ */
+@Provider
+@PreMatching
+@ApplicationScoped
+public class VersionHeaderFilter implements ContainerResponseFilter {
+    @ConfigProperty(name = "quarkus.application.version")
+    String applicationVersion;
+
+    public VersionHeaderFilter() {
+    }
+
+    public void filter(ContainerRequestContext requestContext, ContainerResponseContext responseContext) throws IOException {
+        responseContext.getHeaders().add("Build-Version", applicationVersion);
+    }
+}
--- a/http/rest-client-reactive-vanilla/src/main/java/io/quarkus/ts/http/restclient/reactive/VersionHeaderFilter.java
+++ b/http/rest-client-reactive-vanilla/src/main/java/io/quarkus/ts/http/restclient/reactive/VersionHeaderFilter.java
@@ -1,7 +1,5 @@
 package io.quarkus.ts.http.restclient.reactive;
 
-import java.io.IOException;
-
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.ws.rs.container.ContainerRequestContext;
 import jakarta.ws.rs.container.ContainerResponseContext;
@@ -9,22 +7,18 @@ import jakarta.ws.rs.container.ContainerResponseFilter;
 import jakarta.ws.rs.container.PreMatching;
 import jakarta.ws.rs.ext.Provider;
 
-import org.eclipse.microprofile.config.inject.ConfigProperty;
-
 /**
  * Used for {@link RestCallerService}
+ * Does nothing we just need quarkus to have a possibility to use a ContainerResponseFilter
  */
 @Provider
 @PreMatching
 @ApplicationScoped
 public class VersionHeaderFilter implements ContainerResponseFilter {
-    @ConfigProperty(name = "quarkus.application.version")
-    String applicationVersion;
-
     public VersionHeaderFilter() {
     }
 
-    public void filter(ContainerRequestContext requestContext, ContainerResponseContext responseContext) throws IOException {
-        responseContext.getHeaders().add("Build-Version", applicationVersion);
+    public void filter(ContainerRequestContext requestContext, ContainerResponseContext responseContext) {
+        // do nothing. implements this method just because ContainerResponseFilter requires it
     }
 }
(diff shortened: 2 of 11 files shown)


mocenas:
run tests

http/rest-client-reactive-vanilla/src/test/java/io/quarkus/ts/http/restclient/reactive/VanillaReactiveRestClientIT.java
```diff
@@ -0,0 +1,34 @@
...
+    @Test
+    @Tag("https://github.com/quarkusio/quarkus/issues/31024")
+    public void restClientContainerFilterTest() {
+        List<String> logs = app.getLogs();
+
+        // app produces error log (not stack trace) after startup in its log if issue in not fixed,
+        // search fox this error text
+        for (String log : logs) {
+            if (log.contains(CONTAINER_FILTER_ERROR_TEXT)) {
+                fail("Detected failure-indicating text in app's log. Issue https://github.com/quarkusio/quarkus/issues/31024is not fixed.");
```
michalvavrik, line 30:
```suggestion
                fail("Detected failure-indicating text in app's log. Issue https://github.com/quarkusio/quarkus/issues/31024 is not fixed.");
```

it may not be fixed, but _is not fixed_ is probably incorrect, there is plenty other reasons this can be logged

http/rest-client-reactive-vanilla/src/main/java/io/quarkus/ts/http/restclient/reactive/VersionHeaderFilter.java
```diff
@@ -0,0 +1,30 @@
...
+@ApplicationScoped
+public class VersionHeaderFilter implements ContainerResponseFilter {
+    @ConfigProperty(name = "quarkus.application.version")
+    String applicationVersion;
+
+    public VersionHeaderFilter() {
+    }
+
+    public void filter(ContainerRequestContext requestContext, ContainerResponseContext responseContext) throws IOException {
+        responseContext.getHeaders().add("Build-Version", applicationVersion);
```
michalvavrik, line 28: This filter is only applied when you actually call Jakarta REST resource (which you have none), so I'm not sure why you bother with `Build-Version` headers. Please at least give it context with comment or dropped dead code as IMO noone will remember why it is there in a week. Thanks

michalvavrik: If the point is to test that you have been and config property injection works, well, you still don't need headers that are never set.

http/rest-client-reactive-vanilla/src/main/java/io/quarkus/ts/http/restclient/reactive/RestCallerService.java
```diff
@@ -0,0 +1,30 @@
...
+    void init() {
+        callRest();
+    }
+
+    void callRest() {
+        RestClientBuilder builder = RestClientBuilder.newBuilder()//
+                .baseUri(URI.create("localhost"));
+
+        // API needs to be created for issue to manifest
+        UselessRestApi api = builder.build(UselessRestApi.class);
```
michalvavrik, line 28: Can you actually call something? Building REST client just to check static registration inside REST client extension feels very much like unit test. Then it belongs to Quarkus project (please feel free to contribute it there).

My preference is to write actual scenarios, as if we already pay a price of building this in JVM / native etc. one Vert.x setup and REST call is almost undetectable but we will have another test coverage of extension code (end-to-end not some isolated test).

http/rest-client-reactive-vanilla/src/test/java/io/quarkus/ts/http/restclient/reactive/VanillaReactiveRestClientIT.java
```diff
@@ -0,0 +1,34 @@
...
+public class VanillaReactiveRestClientIT {
+    private static final String CONTAINER_FILTER_ERROR_TEXT = "CDI: programmatic lookup problem detected";
+
+    @QuarkusApplication
+    static RestService app = new RestService();
+
+    @Test
+    @Tag("https://github.com/quarkusio/quarkus/issues/31024")
+    public void restClientContainerFilterTest() {
+        List<String> logs = app.getLogs();
```
michalvavrik, line 24: IMO you should add Vert.x HTTP extension and call the route which uses REST client to call another route to receive response. Then you have end to end test.

michalvavrik: Another argument I can think of is not hanging on log message when not necessary. Problem is that the message can be slightly altered and we would never know (which means test would not fail with a regression).

http/rest-client-reactive-vanilla/src/test/java/io/quarkus/ts/http/restclient/reactive/VanillaReactiveRestClientIT.java
```diff
@@ -0,0 +1,34 @@
...
+
+        // app produces error log (not stack trace) after startup in its log if issue in not fixed,
+        // search fox this error text
```
michalvavrik, line 27:
```suggestion
        // search for this error text
```

http/rest-client-reactive-vanilla/src/main/java/io/quarkus/ts/http/restclient/reactive/RestCallerService.java
mocenas: In this case, it probably belongs to quarkus project. There is no reason to actually use the API, neither to do any HTTP calls to the service. AFAIK only detectable way the issue manifest itself is nit the app logs.

WDYT @rsvoboda is it OK if we scap this and move this test to upstream, or should we have it in our TS?

michalvavrik: okay, I'll respect whatever you go with.

just to be clear - I think you should just use the client and that's it.

michalvavrik:
@mocenas I've tried it and the exception is not detectable, you were right about logging, let's stick with it.

http/rest-client-reactive-vanilla/src/test/java/io/quarkus/ts/http/restclient/reactive/VanillaReactiveRestClientIT.java
mocenas: fixed.

http/rest-client-reactive-vanilla/src/test/java/io/quarkus/ts/http/restclient/reactive/VanillaReactiveRestClientIT.java
mocenas: As stated below, we agreed that there is no better way, than to look into logs.

http/rest-client-reactive-vanilla/src/main/java/io/quarkus/ts/http/restclient/reactive/VersionHeaderFilter.java
mocenas: I've changed the filter and removed the header altering from it. Now the filter just "exists", so quarkus can inject it, but filter has no real actions. I've also added comments to it, why it is here.

http/rest-client-reactive-vanilla/src/main/java/io/quarkus/ts/http/restclient/reactive/RestCallerService.java
mocenas: We've agreed, that there is no real way to use REST or any other API to detect the failure.

http/rest-client-reactive-vanilla/src/main/java/io/quarkus/ts/http/restclient/reactive/VersionHeaderFilter.java
```diff
@@ -0,0 +1,24 @@
+package io.quarkus.ts.http.restclient.reactive;
```
michalvavrik, line 1: I think it should also contain _vanilla_ in package name because it is easier to identify Maven module just from package name when reading Jenkins.

I don't insist on this.

http/rest-client-reactive-vanilla/src/main/java/io/quarkus/ts/http/restclient/reactive/RestCallerService.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+@Startup
+public class RestCallerService {
+
+    @PostConstruct
+    void init() {
+        callRest();
+    }
+
+    void callRest() {
```
michalvavrik, line 23: The name _callRest_ is not really precise as there is no call. I don't insist on changing it, just saying :-)

michalvavrik (approved):
LGTM

AFAIC you can merge it or wait for @rsvoboda if he finds a time. Thanks

michalvavrik:
Remainder: Whoever merge this must use _Squash and merge_ option.

http/rest-client-reactive-vanilla/src/main/java/io/quarkus/ts/http/restclient/reactive/VersionHeaderFilter.java
mocenas: You are right, it would be better for investigation. I've changed it.

http/rest-client-reactive-vanilla/src/main/java/io/quarkus/ts/http/restclient/reactive/RestCallerService.java
mocenas: I renamed it, and further simplified this class.

mocenas:
> Sorry, I just realized you need to add note to README about new module. Then it's fine with me.

added

PR #1633, opened 2024-01-29
Re-run failing tests to limit flakiness

michalvavrik:
### Summary

Re-runs baremetal failures twice and OCP failures once. There is minimal cost of doing that on baremetal as JVM build takes little while and native build will be re-used when I make adjustments in the Test Framework (right now custom build is not re-used, but there is little point of testing Quarkus build indempotency).

This is part of efforts to limit false failures. After that must come way to detect flaky tests that were re-run and report it so that we address flakyness.

This is how logs will look like (OCP will be only re-run once though):

```
[ERROR] io.quarkus.ts.http.minimum.reactive.OpenShiftHttpMinimumReactiveIT.fail -- Time elapsed: 5.265 s <<< FAILURE!
org.opentest4j.AssertionFailedError
	at io.quarkus.ts.http.minimum.reactive.HttpMinimumReactiveIT.fail(HttpMinimumReactiveIT.java:33)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR] io.quarkus.ts.http.minimum.reactive.OpenShiftHttpMinimumReactiveIT.fail
[ERROR]   Run 1: OpenShiftHttpMinimumReactiveIT>HttpMinimumReactiveIT.fail:33
[ERROR]   Run 2: OpenShiftHttpMinimumReactiveIT>HttpMinimumReactiveIT.fail:33
[ERROR]   Run 3: OpenShiftHttpMinimumReactiveIT>HttpMinimumReactiveIT.fail:33
[INFO] 
[INFO] 
[ERROR] Tests run: 7, Failures: 1, Errors: 0, Skipped: 0
[INFO] 

```

Please select the relevant options.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Dependency update
- [x] Refactoring
- [ ] Backport
- [ ] New scenario (non-breaking change which adds functionality)
- [ ] This change requires a documentation update
- [ ] This change requires execution against OCP (use `run tests` phrase in comment)

### Checklist:
- [x] Methods and classes used in PR scenarios are meaningful
- [x] Commits are well encapsulated and follow [the best practices](https://cbea.ms/git-commit/)

PR #1634, opened 2024-01-30
[3.2] Re-run failing tests to limit flakiness

michalvavrik:
### Summary

Backports https://github.com/quarkus-qe/quarkus-test-suite/pull/1633.

- [ ] Dependency update
- [ ] Refactoring
- [x] Backport

### Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -177,6 +177,9 @@
                 <plugin>
                     <artifactId>maven-surefire-plugin</artifactId>
                     <version>${surefire-plugin.version}</version>
+                    <configuration>
+                        <rerunFailingTestsCount>2</rerunFailingTestsCount>
+                    </configuration>
                 </plugin>
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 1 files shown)


michalvavrik:
run tests

rsvoboda:
I know this is backport, but why different numbers of re-runs (1 vs. 2) and wouldn't be parametrization of this through a property more handy?

michalvavrik:
> I know this is backport, 

np, I'm open to changes in main or in backport (and port them)

> but why different numbers of re-runs (1 vs. 2)

- Re-run of JVM is not a big deal as it only starts app again, no JAR build etc. (and starting containers etc.)
- Re-run in native is even lesser deal on main when this change https://github.com/quarkus-qe/quarkus-test-framework/pull/1021 is propagated to TS, because native executable starts even quicker, so re-run is cheap (it is not true for 3.2 for there build works differently and it will not be re-used, but still, it should be rather stable without failures, right?)
- Re-run of OpenShift requires creating new project, pushing it to the OpenShift + same as above, I think such a re-run will take significantly longer depending on the scenario.

However @mjurc had same doubts, I'm not sure, if you feel it doesn't matter, we can also re-run it twice in OpenShift. I was afraid of these serial failures when there is bump of Fabric8 and many tests are not working, it would take longer.

> and wouldn't be parametrization of this through a property more handy?

I'd expect you can easily overrun this with `-Dfailsafe.rerunFailingTestsCount=3` if you need to. Didn't try it, but command line should have higher priority, right? In which case I don't see an advantage.

michalvavrik:
FYI https://maven.apache.org/surefire/maven-failsafe-plugin/examples/rerun-failing-tests.html#rerun-failing-tests

rsvoboda (approved):

PR #1637, opened 2024-02-01
Add tests coverage for REST client reactive using HTTP2

jedla97:
### Summary

Adding test coverage for [QUARKUS-658](https://issues.redhat.com/browse/QUARKUS-658).

This cover basic scenarios which Quarkus users can use. 

The `http2ClientAsync` and `http2ClientSync` which are removed is covered by new tests. In addition the class these test were using were also removed. The ones from http-advanced were removed completely as Quarkus doesn't plan to support the non-reactive client.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


jedla97:
The `HttpVersionAlpnWithConfigKeyIT.testAsyncResponseWithSingleClientOnSingleClient` and `HttpVersionAlpnWithConfigKeyIT.testSyncResponseWithSingleClientOnSingleClient` are expected to fail because the https://github.com/quarkusio/quarkus/issues/38514 but I PR for it is created and passing so it will be merged pretty soon.

Cookie test failing also on daily so it's not caused by this PR

jcarranzan:
> The `HttpVersionAlpnWithConfigKeyIT.testAsyncResponseWithSingleClientOnSingleClient` and `HttpVersionAlpnWithConfigKeyIT.testSyncResponseWithSingleClientOnSingleClient` are expected to fail because the [quarkusio/quarkus#38514](https://github.com/quarkusio/quarkus/issues/38514) but I PR for it is created and passing so it will be merged pretty soon.
> 
> Cookie test failing also on daily so it's not caused by this PR

The issue is already closed so we can run again the tests

jedla97:
The failing test are just cookies which will be fixed by https://github.com/quarkus-qe/quarkus-test-suite/pull/1639

http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/resources/HttpVersionClientResource.java
```diff
@@ -0,0 +1,145 @@
...
+    public Response httpSynchronousForClientWithKey() throws MalformedURLException {
+        Response response = httpVersionClientWithConfigKey.getClientHttpVersion(NAME);
+        return checkHttpVersion(response);
+    }
+
+    @GET
+    @Path("https-asynchronous-for-client-with-key")
+    public Response httpsAsynchronousForClientWithKey() throws MalformedURLException {
+        Response response = httpsVersionClientWithConfigKey.getClientHttpVersionAsync(NAME)
+                .await().atMost(Duration.ofSeconds(10));
```
jcarranzan, line 85: Consider using a constant for seconds (10).

http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/resources/HttpVersionClientResource.java
```diff
@@ -0,0 +1,145 @@
...
+
+    /**
+     * Check response from the REST client if http 2 is used
+     *
+     * @param response response of REST client
+     * @return Response contain original response from endpoint or error message containing actual http version used
+     */
+    private Response checkHttpVersion(Response response) {
+        String httpVersion = ((ClientResponseImpl) response).getHttpVersion();
+        if (httpVersion.equals("HTTP_2")) {
```
jcarranzan, line 139: I think there is some method like `HttpVersion.HTTP_2.name()` it would be good to see it instead of the string "HTTP_2"

http/rest-client-reactive/src/test/java/io/quarkus/ts/http/restclient/reactive/HttpVersionAlpnWithConfigKeyIT.java
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.QuarkusApplication;
+
+@QuarkusScenario
+public class HttpVersionAlpnWithConfigKeyIT {
+
+    String SUCESSFUL_RESPONSE_STRING = "Hello " + NAME;
```
jcarranzan, line 16: Typo in  the word should be SUCCESSFUL_, you could change it in all SUCESSFUL_RESPONSE_STRING

jcarranzan (commented):
Hi @jedla97, I've left some minor comments, but overall, the code looks good to me. Additionally, I was thinking that you might consider including a couple of tests that intentionally induce failures or check for the possibility of receiving a status code other than 200. 
Thank you

http/rest-client-reactive/src/main/resources/httpVersion.properties
```diff
@@ -0,0 +1,6 @@
+// server settings
+quarkus.http.http2=true
```
jcarranzan, line 2: According to the guide: https://es.quarkus.io/guides/http-reference#http2-support , the http2 property is already set by default, so you can remove it.

http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/resources/HttpVersionClientResource.java
```diff
@@ -0,0 +1,145 @@
...
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.eclipse.microprofile.rest.client.RestClientBuilder;
+import org.eclipse.microprofile.rest.client.inject.RestClient;
+import org.jboss.resteasy.reactive.client.impl.ClientResponseImpl;
+
+import io.quarkus.ts.http.restclient.reactive.HttpVersionClient;
+import io.quarkus.ts.http.restclient.reactive.HttpVersionClientWithConfigKey;
+import io.quarkus.ts.http.restclient.reactive.HttpsVersionClientWithConfigKey;
+
+@Path("http2")
```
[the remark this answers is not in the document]
jedla97: What I know it doesn't matter it's always treated same. But I'll will add it as I see that is used everywhere else.

From docs of path
> For the purposes of absolutizing a path against the base URI , a leading '/' in a path is ignored and base URIs are treated as if they ended in '/'

http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/resources/HttpVersionClientResource.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @GET
+    @Path("http-asynchronous")
+    public Response httpAsynchronous() throws MalformedURLException {
+        return createAsynchronousResponseForClient(baseUri);
+    }
+
+    @GET
+    @Path("https-synchronous-for-client-with-key")
+    public Response httpsSynchronousForClientWithKey() throws MalformedURLException {
```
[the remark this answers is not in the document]
jedla97: Thanks for the catch

http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/resources/HttpVersionClientResource.java
```diff
@@ -0,0 +1,145 @@
...
+    @RestClient
+    HttpVersionClientWithConfigKey httpVersionClientWithConfigKey;
+
+    @Inject
+    @RestClient
+    HttpsVersionClientWithConfigKey httpsVersionClientWithConfigKey;
+
+    public HttpVersionClientResource(@ConfigProperty(name = "quarkus.http.port") int httpPort,
+            @ConfigProperty(name = "quarkus.http.ssl-port") int sslHttpPort) {
+        this.baseUri = URI.create("http://localhost:" + httpPort);
```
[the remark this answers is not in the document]
jedla97: Good idea I change it

jedla97:
Thanks for review @jcarranzan I update things which you have pointed out.

> Additionally, I was thinking that you might consider including a couple of tests that intentionally induce failures or check for the possibility of receiving a status code other than 200.

In this I will try thing about negative scenario and if I have some idea I'll create standalone PR for it.
In case of checking for 200 status in tests.

It's there yous to check but not testing the client itself. The test calling endpoint where is the client created/injected and that endpoint calling another endpoint as rest client. So status in tests is there just to check that's response is as expected (client use/not use http2). If there will be some problem with client it should return 500.

jedla97:
Just note why we can't use the rest client directly in test:

https://github.com/quarkusio/quarkus/issues/25041

Tl;DR : At this moment quarkus rest client reactive is not meant to be used standalone.

http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/resources/HttpVersionClientResource.java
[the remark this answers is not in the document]
jcarranzan: I keep seeing that, Have you already tweaked it in some way?

jcarranzan (commented):
Thank you for the tweaks, just only a comment about the MalformedURLException signature and I'll approve it.

http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/resources/HttpVersionClientResource.java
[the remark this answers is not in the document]
jedla97: Sorry I must forget to update it. Now it should be fine

http/rest-client-reactive/src/main/resources/httpVersion.properties
```diff
@@ -0,0 +1,3 @@
+// general REST client setting
```
michalvavrik, line 1: 1. properties comments shouldn't created with `//` whatever your framework works. If our framework really requires this, please open issue in Framework project
2. I don't think this comment is valuable. If you follow this pattern, you will have to comment out everything. One of million examples - `quarkus.http.ssl.certificate.key-store-file` in the file right above, is it better known to readers than following properties?
3. The comment doesn't seem true, this is client setting specific to concrete named clients?

I'm not against comments at all, but this one particular strikes me. Thouths @jedla97 ?

michalvavrik (commented):
I don't have time to review this right now (not that I was asked to :-D), but once this is approved please do not merge. There is a lot of changes (nothing against them!) and I'd like to check when I find a time. Thanks

http/rest-client-reactive/src/main/resources/httpVersion.properties
jedla97: Removed the comment. I had there part for server side settings so that make sense for me. But now agree that it's not needed.

http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/HttpVersionClientWithConfigKey.java
```diff
@@ -0,0 +1,23 @@
...
+@Path("/httpVersion")
+public interface HttpVersionClientWithConfigKey {
+
+    @GET
+    @Path("synchronous")
+    Response getClientHttpVersion(@QueryParam("name") String name);
+
+    @GET
+    @Path("asynchronous")
+    Uni<Response> getClientHttpVersionAsync(@QueryParam("name") String name);
```
jcarranzan, line 22: Hi @jedla97 , I noticed some code duplication in the method interfaces between HttpVersionClientWithConfigKey and HttpVersionClient. Perhaps it could be improved by applying inheritance to a single interface containing the methods getClientHttpVersion and getClientHttpVersionAsync, and then having the rest of the interfaces extend from this "BaseInterface" with those methods.

jedla97: @jcarranzan updated. I done it between the `ClientWithConfigKey` interfaces but not the rest. I agree that inherit it is better way. The `HttpVersionClientWithConfigKey` and `HttpsVersionClientWithConfigKey` inheriting from `HttpVersionClient` as they do same thing and are just specify name of `RestClient`. 

Personally don't want to introduce one new interface (base interface) and have 3 another interfaces inheriting from it. I see this base interface as something which are not needed + the inheritance from `HttpVersionClient` make sense for me.

jcarranzan: Thank you now it make sense to me also :+1:

http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/resources/HttpVersionClientResource.java
```diff
@@ -0,0 +1,149 @@
...
+
+import io.quarkus.ts.http.restclient.reactive.HttpVersionClient;
+import io.quarkus.ts.http.restclient.reactive.HttpVersionClientWithConfigKey;
+import io.quarkus.ts.http.restclient.reactive.HttpsVersionClientWithConfigKey;
+import io.vertx.core.http.HttpVersion;
+
+@Path("/http2")
+public class HttpVersionClientResource {
+
+    public static final String WRONG_HTTP_VERSION = "The HTTP version should be HTTP_2 but is ";
```
michalvavrik, line 25:
```suggestion
    public static final String WRONG_HTTP_VERSION = "The HTTP version should be HTTP/2 but is ";
```

http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/resources/HttpVersionBaseResource.java
```diff
@@ -0,0 +1,26 @@
...
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.QueryParam;
+import jakarta.ws.rs.core.Response;
+
+import io.smallrye.mutiny.Uni;
+
+@ApplicationScoped
+@Path("/httpVersion")
```
michalvavrik, line 12:
```suggestion
@Path("/http-version")
```
That's usual convention, or are you testing whether it works with case sensitive path? If so, let's document.

http/rest-client-reactive/src/test/java/io/quarkus/ts/http/restclient/reactive/HttpVersionAlpnWithConfigKeyIT.java
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.QuarkusApplication;
+
+@QuarkusScenario
+public class HttpVersionAlpnWithConfigKeyIT {
+
+    String SUCCESSFUL_RESPONSE_STRING = "Hello " + NAME;
```
michalvavrik, line 16:
```suggestion
    private static final String SUCCESSFUL_RESPONSE_STRING = "Hello " + NAME;
```

http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/resources/HttpVersionClientResource.java
```diff
@@ -0,0 +1,149 @@
...
+     * @param uri base url for client to use
+     * @return Response contain original response from endpoint or error message containing actual http version used
+     */
+    private Response createAsynchronousResponseForClient(URI uri) throws MalformedURLException {
+        HttpVersionClient httpVersionClient = RestClientBuilder.newBuilder()
+                .baseUrl(uri.toURL())
+                .build(HttpVersionClient.class);
+
+        Response response = httpVersionClient.getClientHttpVersionAsync(NAME).await()
+                .atMost(Duration.ofSeconds(TIMEOUT_SECONDS));
```
michalvavrik, line 131: I think blocking here is just laziness, there is no point to it. Asynchronous test should be processed in a reactive manner and return stuff asynchronously from resource as well. Please feel free to prove me wrong. Thanks

http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/resources/HttpVersionClientResource.java
```diff
@@ -0,0 +1,149 @@
...
+                .baseUrl(uri.toURL())
+                .build(HttpVersionClient.class);
+
+        Response response = httpVersionClient.getClientHttpVersion(NAME);
+        return checkHttpVersion(response);
+    }
+
+    /**
+     * Create HttpVersionClient and make REST client asynchronous request to `httpVersion/asynchronous endpoint`
+     * Response from the client is checked if http 2 was used
```
michalvavrik, line 120:
```suggestion
     * Response from the client is checked if HTTP/2 was used
```
please check other comments as well. As you have already decided to write nice Java doc.

http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/resources/HttpVersionBaseResource.java
```diff
@@ -0,0 +1,26 @@
...
+public class HttpVersionBaseResource {
+
+    @GET
+    @Path("synchronous")
+    public Response sayHello(@QueryParam("name") String name) {
+        return Response.ok("Hello " + name).build();
```
michalvavrik, line 18: Hello is fine, you can keep it if you want as it tests query param. It would be nice to have incoming request protocol in response. This way, you can check that REST client is telling you a true about used HTTP version. I propose to inject `RoutingContext`, then `io.vertx.ext.web.RoutingContext#request` -> `io.vertx.core.http.HttpServerRequest#version` should always return `HTTP_2`.

michalvavrik (approved):
This looks fine. @jcarranzan please have a look next week, review and merge. If you need more time, can you maybe create dedicated ticket for yourself in JIRA, please? Let' s move on. Thxc

michalvavrik:
@jedla97 let's add info about HTTP/2 for REST client into root project README

http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/resources/HttpVersionClientResource.java
michalvavrik: Now I'm thinking of it, you probably used actual response type value as they use underscore, fine then.

jedla97:
@jcarranzan I update all the comments mentioned by @michalvavrik so can you have quick look. Also add mention of these test to the README.

jcarranzan:
Changes LGTM, and tests are correct as well. However I noticed the test classes seem to contain repetitive tests, and perhaps using inheritance could consolidate tests into a single class. Nonetheless, I respect the current approach, it's simply a different way of approaching the tests. 
On the other hand, try to resolve the comments from Michal, still I can see some unresolved threads, thank you.

jcarranzan:
Changes LGTM, and tests are correct as well. However I noticed the test classes seem to contain repetitive tests, and perhaps using inheritance could consolidate tests into a single class. Nonetheless, I respect the current approach, it's simply a different way of approaching the tests. 
On the other hand, try to resolve the comments from Michal, still I can see some unresolved threads, thank you.

jedla97:
Thanks for review
> However I noticed the test classes seem to contain repetitive tests, and perhaps using inheritance could consolidate tests into a single class

When writing the test I was thinking about it, but if we want one class as base for multiple tests there would be subset of 2 tests for all 4 tests classes.
When we would have 2 base classes the subset of test would be 4 (all https tests). So in the result we shouldn't have 4 classes but 6. 

In my POV it's not that useful in this scenario.

> On the other hand, try to resolve the comments from Michal, still I can see some unresolved threads, thank you.

I update the solution but sum it in one comment as I was resolve all the comment on code side (not GitHub side)

PR #1655, opened 2024-02-12
HTTP2 REST test cases

jcarranzan:
### Summary

Create HTTP/2 test cases using RestEasy reactive, Vert.x HttpClient.

- HttpClient Vertx HTTP/2 Test: Verifies that the HTTP client is functioning correctly with HTTP/2, ensuring that the response status code is OK, and the version is as expected.

- Test max header size exceeded: Verifies that when the maximum header size is exceeded using HTTP/2, the appropriate error message is returned. It tests this scenario by incrementally adding headers until the limit is reached.

- X-Content-Type-Options Header test: Checks that the X-Content-Type-Options header is correctly set to "nosniff" in the HTTP/2 response headers. It also ensures that the status code, version, and response body are as expected.
- 

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


http/http-advanced-reactive/src/test/resources/http2.properties
```diff
@@ -0,0 +1,7 @@
+quarkus.oidc.enabled=false
+quarkus.tls.trust-all=true;
```
michalvavrik, line 2:
```suggestion
quarkus.tls.trust-all=true
```

http/http-advanced-reactive/pom.xml
```diff
@@ -89,5 +98,13 @@
             <groupId>commons-io</groupId>
             <artifactId>commons-io</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-undertow</artifactId>
```
michalvavrik, line 103: This is bit dangerous because we lose coverage of RESTEasy Reactive without Undertow and their integration is cumbersome. Can you please move serverlet part to the `servlet-undertow` module? If yo need RR, add it via force deps there.

michalvavrik (changes requested):
I'll review later, but undertow portion needs to be moved.

http/http-advanced-reactive/pom.xml
jcarranzan: Sure, I can move it to the servlet-undertow module.

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,219 @@
...
+                .onSuccess(body -> {
+                    assertThat("Body response", body.toString().contains(GREETING));
+                    checkpoint.flag();
+                })
+                .onFailure(vertxTestContext::failNow);
+    }
+
+    @AfterAll
+    public static void closeConnections() {
+        httpClient.close();
```
michalvavrik, line 215: this class creates 5 HTTP clients and closes one.

http/http-advanced-reactive/src/main/resources/META-INF/web.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<web-app>
+    <servlet>
+        <servlet-name>io.quarkus.ts.http.advanced.reactive.Servlet</servlet-name>
+        <servlet-class>io.quarkus.ts.http.advanced.reactive.Servlet</servlet-class>
+    </servlet>
+    <servlet-mapping>
+        <servlet-name>io.quarkus.ts.http.advanced.reactive.Servlet</servlet-name>
```
michalvavrik, line 8: how this works? shouldn't it be removed?

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,219 @@
...
+                    assertThat("Body response", body.toString().contains(GREETING));
+
+                    requestCheckpoint.flag();
+
+                }).onFailure(Throwable::printStackTrace));
+
+    }
+
+    @Test
+    @DisplayName("HttpClient Vertx HTTP/2 protocol Test with ssl, alpn, trustAll true and verifyHost false")
```
michalvavrik, line 80:
```suggestion
    @DisplayName("HttpClient Vertx HTTP/2 protocol test with ssl, alpn, trustAll true and verifyHost false")
```

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,219 @@
...
+        Checkpoint requestCheckpoint = vertxTestContext.checkpoint(2);
+        assertNotNull(vertx);
+        httpClient = vertx.createHttpClient(new HttpClientOptions()
+                .setSsl(true)
+                .setUseAlpn(true)
+                .setTrustAll(true)
+                .setVerifyHost(false)
+                .setProtocolVersion(HttpVersion.HTTP_2));
+        assertNotNull(httpClient);
+        vertxTestContext.verify(() -> httpClient.request(HttpMethod.GET, 8443, baseUri.getHost(), BASE_ENDPOINT)
```
michalvavrik, line 92: `8443`? I think you should use `@QuarkusApplication(ssl = true)` and get port from `app.getURI(HTTPS)`

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,219 @@
...
+import io.vertx.core.http.HttpMethod;
+import io.vertx.core.http.HttpVersion;
+import io.vertx.junit5.Checkpoint;
+import io.vertx.junit5.VertxExtension;
+import io.vertx.junit5.VertxTestContext;
+
+@QuarkusScenario
+@ExtendWith(VertxExtension.class)
+public class Http2IT {
+    private static final Logger LOG = Logger.getLogger(Http2IT.class);
```
michalvavrik, line 34: I'm sorry but even though there is no hard convention, it's not customary to sort properties like this. please group _private static final_ and under them static non-final. Thanks

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,219 @@
...
+        vertxTestContext.verify(() -> httpClient.request(HttpMethod.GET, baseUri.getPort(), baseUri.getHost(), BASE_ENDPOINT)
+                .compose(req -> req.send()
+                        .compose(httpClientResponse -> {
+                            assertEquals(HttpStatus.SC_OK, httpClientResponse.statusCode());
+                            assertEquals(expectedHTTP1Version, httpClientResponse.version().toString());
+                            requestCheckpoint.flag();
+                            return httpClientResponse.body();
+                        }))
+                .onSuccess(body -> {
+                    LOG.info("Got data " + body.toString("ISO-8859-1"));
```
michalvavrik, line 70: Personally I'd only be interested in this information if something goes wrong, like if the assertion fails you should log it as part of the reason of assertion failure. Apart of logging being expensive (which is no big deal here), I just think how would tests look like if everyone logged information like this. We are in a different situation from production because we can reproduce things and if we can't, we have the information logged on failure.

michalvavrik: Same goes to other logging in this class.

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,219 @@
...
+    @DisplayName("HttpClient Vertx HTTP/2 protocol Test with ssl, alpn, trustAll true and verifyHost false")
+    public void httpClientVertxHttp2(VertxTestContext vertxTestContext) {
+
+        Checkpoint requestCheckpoint = vertxTestContext.checkpoint(2);
+        assertNotNull(vertx);
+        httpClient = vertx.createHttpClient(new HttpClientOptions()
+                .setSsl(true)
+                .setUseAlpn(true)
+                .setTrustAll(true)
+                .setVerifyHost(false)
```
michalvavrik, line 89: Please help me to understand scope of this PR - I think to test HTTP/2 with trust all = true and verify-host = false avoids anything that could run on production. This is not real-world use case, who will use this? Why not to actually test use generate the certificate and use it? I comprehend it's going to be self-signed certificate so there will be tweaks.

michalvavrik (commented):
Please document how you generated `keystore.jks` because these certificates usually have expiration time and if it ever expires or anyone needs to make any adjustment, they will not be able to use it or adjust it. Please use local project README.

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
jcarranzan: you are right I will correct it, good catch, thanks.

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
jcarranzan: Yes, I agree, sometimes I do some "`System.out.println`", in the tests to check that everything is working as expected, I changed to LOG for best practices, but that's all.

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
jcarranzan: I've removed old keystore, and I've created a new one, now we can remove the HttpClient option `.setVerifyHost(false)` and use the keystore.jks. However if we can also remove trust all= true we would need a valid SSL certificate from a trusted Certificate Authority (CA) , maybe using https://quarkus.io/guides/security-customization#registering-security-providers. But I don't think it's the purpose of these tests....

michalvavrik: sure, let's stick with trust all

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,219 @@
...
+import io.vertx.junit5.VertxExtension;
+import io.vertx.junit5.VertxTestContext;
+
+@QuarkusScenario
+@ExtendWith(VertxExtension.class)
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
+public class Http2IT {
+    @QuarkusApplication(ssl = true, classes = { MorningResource.class }, properties = "http2.properties")
+    static RestService app = new RestService();
+    // .withProperty("quarkus.rest-client.http2", "true");
```
michalvavrik, line 40:
```suggestion

```
?

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,219 @@
...
+                    testContext.completeNow();
+                    break;
+                } else {
+                    testContext.failNow(e);
+                }
+            }
+        }
+    }
+
+    private void request(AtomicInteger headerCount, VertxTestContext testContext) {
```
michalvavrik, line 160: It's not important, but please kindly make the method static, thanks.

michalvavrik (commented):
@jcarranzan as I don't understand what is target scope, I can only review what I can see - looks good. Please make last required changes (very small things) and add a note to TS README about http-advanced-reactive HTTP/2 Vert.x test.

michalvavrik:
Issue description _Create HTTP/2 test cases using RestEasy reactive client_ is not true, you are not using RESTEasy Reactive client, please fix that.,

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,219 @@
...
+        HttpClientOptions options = new HttpClientOptions()
+                .setSsl(true)
+                .setUseAlpn(true)
+                .setTrustAll(true)
+                .setProtocolVersion(HttpVersion.HTTP_2);
+        httpClient = vertx.httpClientBuilder().with(options).build();
+
+        int expectedLimit = 2048;
+
+        for (int i = 0; i < expectedLimit; i++) {
```
michalvavrik, line 144: So basically you make almost concurrently 2048 requests? Do you realize that your calls are not waiting for responses so that you actually call pretty random number till you reached limit?

I'm not completely against it, but you don't even test that it actually failed on 2048, I think you should make it more deterministic and when `e.getMessage().contains("Header size exceeded max allowed size")` is true, you should check `i`

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,219 @@
...
+            AtomicInteger headerCount = new AtomicInteger(i);
+
+            try {
+                request(headerCount, testContext);
+            } catch (Exception e) {
```
michalvavrik, line 149: How can this ever happen that you can catch Exception here? I would expect `onFailure` be called otherwise it is a bug?

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,219 @@
...
+                .onFailure(throwable -> {
+                    if (throwable instanceof Http2Exception &&
+                            throwable.getMessage().contains("Header size exceeded max allowed size")) {
+                        testContext.completeNow();
+                    } else {
+                        testContext.failNow(throwable);
+                    }
+                })
+                .onSuccess(response -> {
+                    testContext.failNow(new AssertionError("Unexpected successful response"));
```
michalvavrik, line 180: Can you add `i` to the error message please?

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,219 @@
...
+                .onSuccess(body -> {
+                    assertThat("Body response", body.toString().contains(GREETING));
+                    checkpoint.flag();
+                })
+                .onFailure(vertxTestContext::failNow);
+    }
+
+    @AfterAll
+    public static void closeConnections() {
+        httpClient.close();
```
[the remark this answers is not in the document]
jcarranzan: Srry, I was confusing, I though just it was created once in BeforeAll, re-using it in the test and closing at the @AfterAll method..

michalvavrik:
> > Issue description _Create HTTP/2 test cases using RestEasy reactive client_ is not true, you are not using RESTEasy Reactive client, please fix that.,
> 
> Well with my last change in MorningResource, I think I use it, if not just explain to me pls.

There is no client used in the `MorningResource`, btw there is no such thing as RESTEasy Reactive client anyway, I am just asking you to improve description :-)

http/http-advanced-reactive/src/test/resources/http2.properties
```diff
@@ -0,0 +1,8 @@
+quarkus.oidc.enabled=false
+quarkus.tls.trust-all=true
```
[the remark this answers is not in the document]
jcarranzan: remove it. And I've removed also setTrustAll from the tests using keystore.jks , good catch Michal, thanks!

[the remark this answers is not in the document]
michalvavrik: perfect, thanks

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,261 @@
...
+                    });
+                    return pong;
+                })
+                .result();
+    }
+
+    @AfterEach
+    public void closeConnections() {
+        httpClient.close();
+        vertx.close();
```
michalvavrik, line 258: both returns future, IMO you should make sure that process has finished before you continue. thoughts?

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,261 @@
...
+                            context.verify(() -> {
+                                assertEquals(PING_DATA, frame.payload().toString().toLowerCase());
+                                entry.flag();
+                            });
+                        });
+                    });
+                    request.exceptionHandler(context::failNow);
+                    vertx.setPeriodic(1000, timerID -> {
+                        timerId.set(timerID);
+                        request.writeCustomFrame(10, 0, Buffer.buffer("ping"));
```
michalvavrik, line 247: I have hard time to understand this. It goes to me, but it is really complex test, could you document that test in a code, please? I doubt most of us know how event.connection().ping etc works.

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,261 @@
...
+                        request.putHeader(headerName, headerValue);
+                    }
+                    return request.send();
+                })
+                .onFailure(throwable -> {
+                    if (throwable instanceof Http2Exception &&
+                            throwable.getMessage().contains("Header size exceeded max allowed size ")) {
+                        testContext.completeNow();
+                    } else {
+                        testContext.failNow(throwable.getMessage() + headerCount);
```
michalvavrik, line 185:
```suggestion
                        testContext.failNow(throwable.getMessage() + " ; header count is" + headerCount);
```
or some better text. the point is there is an exceptionmessage51 (where 51 will be headerCount => no space)

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
michalvavrik: just hints will do

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
jcarranzan: From the vert.x docs:

https://vertx.io/docs/vertx-junit5/java/#_closing_and_removal_of_vertx_objects 
_"Injected Vertx objects are being automatically closed and removed from their corresponding scopes.
Similarly, when it is being created by a @BeforeEach method, it is being closed after possible @AfterEach methods have completed."_
https://vertx.io/docs/vertx-core/java/#_closing_the_underlying_connection
_you can [close](https://vertx.io/docs/apidocs/io/vertx/core/http/HttpServerResponse.html#close--) the underlying TCP connection with
close.
Non keep-alive connections will be automatically closed by Vert.x when the response is ended._

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
jcarranzan: I've changed it the complex logic to an easier with `CompletableFuture`, let me know if you are ok with.

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
michalvavrik: > Yes returns Future but if I don't recall wrong vert.x framework already managed properly those instances (vert.x and httpclient) and is no need to close it manually.

I certainly know nothing about this framework, but it is highly suspicious from reading docs because:
1. https://vertx.io/docs/vertx-junit5/java/#_closing_and_removal_of_vertx_objects speaks about injected (you don't do that)
2. https://vertx.io/docs/vertx-junit5/java/#_closing_and_removal_of_vertx_objects also speaks about `Similarly, when it is being created by a @BeforeEach method` but it is unclear whether they mean 
```
@BeforeEach
  @DisplayName("Deploy a verticle")
  void prepare(Vertx vertx, VertxTestContext testContext) {
    vertx.deployVerticle(new SomeVerticle(), testContext.succeedingThenComplete());
  }
```
or what you do (`Vertx.vertx()`)
4. if I am to believe they are able to automatically close `Vertx.vertx()` how they exactly know it was created and how they access it? I suppose there are some events or do they bytecode transform code or something?

If you are 100 % sure, then alright. Let's keep it.

http/http-advanced-reactive/README.md
```diff
@@ -0,0 +1,5 @@
+File `keystore.jks` was generated with command:
+
+```bash
+keytool -genkey -alias mykey -keyalg RSA -keysize 2048 -validity 365 -dname "CN=localhost" -keystore test/resources/keystore.jks -storepass password
```
michalvavrik, line 4:
```suggestion
keytool -genkey -alias mykey -keyalg RSA -keysize 2048 -validity 365 -dname "CN=localhost" -keystore src/test/resources/keystore.jks -storepass password
```

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
michalvavrik: Still, I do not doubt you know more about this and I'm sorry for digging into it but here https://vertx.io/docs/vertx-junit5/java/#_running_tests_on_a_vert_x_context I can see that

1. they are not using static field for the `Vertx` instance. is that significant? should we do the same? I think it might because they don't use the `Vertx` instance, so it looks like they can access it.
2. they create Vertx with `vertx = rtoc.vertx();` is that preferred way? I can see that in the very first test they created with `Vertx.vertx()` and didn't care about closing so it signal it is alright.

So here is my suggestion:

- don't use static field or please give me something I can read up about why it's fine

tbh I'd be calmer if you injected the Vert.x, but you are the expert here so I'll trust you.

jcarranzan: Yes, I agree with injecting Vert.x in this test context. I've done that now.

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    @Test
+    @Order(3)
+    @DisplayName("HttpClient Vertx just with HTTP/2 protocol option set and default port")
+    void http2ProtocolTest(Vertx vertx, VertxTestContext vertxTestContext) {
+        HttpClientOptions httpClientOptions = new HttpClientOptions()
+                .setTrustOptions(new JksOptions().setPath("keystore.jks").setPassword("password"))
+                .setProtocolVersion(HttpVersion.HTTP_2);
+        httpClient = vertx.httpClientBuilder().with(httpClientOptions).build();
```
michalvavrik, line 121: you create `httpClient` inside `BeforeEach` then here and then close it once in `AfterEach`, I suggest it must be one to one relationship (not create it in before each).

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,260 @@
...
+                } else {
+                    testContext.failNow(e);
+                }
+            }
+        }
+    }
+
+    private static void request(int headerCount, VertxTestContext testContext) {
+
+        httpClient.request(HttpMethod.GET, app.getPort(Protocol.HTTPS), baseUri.getHost(), BASE_ENDPOINT)
```
michalvavrik, line 168: `getPort` method is already deprecated (I agree it is convenient, but that's how it is), please use `getURI`

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,259 @@
...
+                        String headerName = "Header" + i;
+                        String headerValue = "Value" + i;
+                        request.putHeader(headerName, headerValue);
+                    }
+                    return request.send();
+                })
+                .onFailure(throwable -> {
+                    if (throwable instanceof Http2Exception &&
+                            throwable.getMessage().contains("Header size exceeded max allowed size ")) {
+                        testContext.completeNow();
```
michalvavrik, line 179: Here is how I understand what is happening there:

1. incrementally call Quarkus in asynchronous manner (=when `i` is incremented, the previous request didn't end for sure); first header count is zero
2. when first HTTP request happens, it happens with header count zero, therefore there is no header
3. when first request succeeds, there is no reason to fail, therefore it should succeed, however on success an error is thrown `Unexpected successful response...`.
4. only point at which this test succeed is when header count exceeds maximum limit that you have for `200`

Why does it work? You send 40 HTTP requests that are not rejected by Quarkus because the header limit has not been reached and they are being processed while after that, you send 159 HTTP requests that fail and is rejecting them before processing is faster, you get success result. Am I right?

So my conclusions:

1. You can't reliably say that first 40 requests will be processed after next 159 requests, so there is a potential for flakiness
2. You set `quarkus.http.limits.max-header-list-size=2048` which is reached with `headerCount=41` on my computer. But your test is also successful if you send only `headerCount=90` or any other `headerCount > 40`. Is there a relation between 40 headers and header size limit 2048? My point is that you are testing there is _**some limit applied**_, but you have no idea what that size limit is. If that is the point, IMO you should document it properly with comments.

Please note that number of requests required may differ depending depending on time (future) and environment.

I don't want to force you to rework this here are my thoughts, use it or refuse it:

1. If you don't really care about `2048` limit, you just try to figure if there is a limit, you just need few requests, not sure why there is 200 of them (determine range safe for all envs and future chagnes)
2. You test framework doesn't give a damn about failed results as long as there is at least one successful result (again, I have no idea how the framework works), because I've rewrote your test to this with `headerCount=1`:
```
.onSuccess(response -> {
                    request(150, testContext);
                    System.out.println("response is " + response + " for " + headerCount);
                    testContext.failNow(new AssertionError("Unexpected successful response " + " ; header count is "
                            + headerCount));
                });
```
and test still succeeded even though you clearly can see that `testContext.failNow` is going to be executed (FAIL) and it is most likely executed before `request(150, testContext);` receive response because I/O is more expensive and is executed asynchronously.

I am thinking that this Vert.x framework is probably reliable for one `testContext`, not 200 hundred of them, otherwise how do we know that test is failing?

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,259 @@
...
+
+        int expectedLimit = 200;
+        int headerCount;
+
+        for (int i = 0; i < expectedLimit; i++) {
+            headerCount = i;
+
+            try {
+                request(headerCount, testContext);
+            } catch (Exception e) {
```
michalvavrik, line 154:
```suggestion

```
Please mention one example when there can be an exception. One is enough. I can't imagine that `e.getMessage().contains("Header size exceeded max allowed size")` will be true and it is not a bug in Vert.x. Can you point me to docs that says that? Vertx. should handle failure as standard state and not to throw exceptions.

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,246 @@
...
+                        String errorMessage = ar.cause().getMessage();
+                        if (errorMessage.contains("Header size exceeded max allowed size")) {
+                            testContext.completeNow();
+                        } else {
+                            testContext.failNow("Unexpected error message: " + errorMessage);
+                        }
+                    }
+                });
+    }
+    private String generateHeaders(int headerCount) {
```
michalvavrik, line 168: please add empty line

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
```diff
@@ -0,0 +1,246 @@
...
+    void testMaxHeaderSizeExceeded(Vertx vertx, VertxTestContext testContext) {
+        HttpClientOptions options = new HttpClientOptions()
+                .setSsl(true)
+                .setUseAlpn(true)
+                .setTrustOptions(new JksOptions().setPath("keystore.jks").setPassword("password"))
+                .setProtocolVersion(HttpVersion.HTTP_2);
+        httpClient = vertx.httpClientBuilder().with(options).build();
+        // Define the expected maximum header size limit
+        // Note: Consider the configured limit, which is 2048 in this case because we are testing quarkus.http.limits.max-header-list-size=2048,
+        // although it may be considerably lower in practice
```
michalvavrik, line 146: I like the note, it's almost correct. I think you confused number of headers with overall headers size. The name of the exception is confusing, but I believe that Netty is referring to the headers size in byte, so when you say 2048 it means total size of headers you generated

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/Http2IT.java
jcarranzan: I've simplified the test, thank you Michal for your points.

michalvavrik (approved):

PR #1656, opened 2024-02-12
Add http/2 test with vertx httpClient into servlet-undertow module

jcarranzan:
### Summary

Developing test cases about the topic HTTP/2, I am moving the servlet undertow test case to the correct sub-module.
The test case uses vert.x HttpClient.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


http/servlet-undertow/src/test/java/io/quarkus/ts/http/undertow/HttpServletWithSessionListenerIT.java
```diff
@@ -51,6 +57,30 @@ public void sessionSecured() {
         thenCheckActiveSessionsEqualTo(1);
     }
 
+    @Test
+    @DisplayName("undertow servlet test http2")
+    public void underTowHttp2Servert(VertxTestContext vertxTestContext) {
+        final String HTTP_2_VERSION = "HTTP_2";
```
michalvavrik, line 63: `HttpVersion.HTTP_2.toString()`? (I didn't check it out, but I suppose you should compare in `assertEquals` response that is basically the `HttpVersion.HTTP_2` with constant in Vert.x? no biggie though

http/servlet-undertow/src/test/java/io/quarkus/ts/http/undertow/HttpServletWithSessionListenerIT.java
```diff
@@ -51,6 +57,30 @@ public void sessionSecured() {
...
+        Vertx vertx = Vertx.vertx();
```
michalvavrik, line 64: this could lead to resource leak, Vertx is closeable.

http/servlet-undertow/src/test/java/io/quarkus/ts/http/undertow/HttpServletWithSessionListenerIT.java
```diff
@@ -51,6 +57,30 @@ public void sessionSecured() {
...
+        String uriUnderTow = "/app/servlet/hello";
+        String expectedResponseFromUndertow = "Hello World";
+        HttpClientOptions options = new HttpClientOptions().setProtocolVersion(HttpVersion.HTTP_2);
+        HttpClient httpClient = vertx.createHttpClient(options);
```
michalvavrik, line 68: this could lead to resource leak, HttpClient is closeable

http/servlet-undertow/src/test/java/io/quarkus/ts/http/undertow/HttpServletWithSessionListenerIT.java
```diff
@@ -51,6 +57,30 @@ public void sessionSecured() {
...
+        httpClient.request(HttpMethod.GET, 1101, "localhost", uriUnderTow)
+                .compose(request -> request.send()
+                        .compose(httpClientResponse -> {
+                            System.out.println("Got response {} with protocol {}" + httpClientResponse.statusCode() + " "
```
michalvavrik, line 72: We are using logger, I was never actually sure why, but if you dig in (or ask around) I'm quite sure there is a good reason for it (not just features!)

http/servlet-undertow/pom.xml
```diff
@@ -23,5 +23,19 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-security</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-vertx</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>jakarta.ws.rs</groupId>
+            <artifactId>jakarta.ws.rs-api</artifactId>
```
michalvavrik, line 32: Why is the API needed, is there something using it, please?

http/servlet-undertow/src/test/java/io/quarkus/ts/http/undertow/HttpServletWithSessionListenerIT.java
```diff
@@ -51,6 +57,30 @@ public void sessionSecured() {
         thenCheckActiveSessionsEqualTo(1);
     }
 
+    @Test
+    @DisplayName("undertow servlet test http2")
```
michalvavrik, line 61: Undertow is a name, so it should be capitalized. I'm also not sure http2 is according to the convention (HTTP/2)

http/servlet-undertow/src/test/java/io/quarkus/ts/http/undertow/HttpServletWithSessionListenerIT.java
```diff
@@ -51,6 +57,30 @@ public void sessionSecured() {
...
+    public void underTowHttp2Servert(VertxTestContext vertxTestContext) {
```
michalvavrik, line 62: - Undertow shouldn't be written as `underTow` as it suggests it is compound of 2 words
- `Servert` is probably a typo

http/servlet-undertow/src/test/java/io/quarkus/ts/http/undertow/HttpServletWithSessionListenerIT.java
```diff
@@ -51,6 +57,30 @@ public void sessionSecured() {
...
+        final String HTTP_2_VERSION = "HTTP_2";
+        Vertx vertx = Vertx.vertx();
+        String uriUnderTow = "/app/servlet/hello";
```
michalvavrik, line 65: 1. `UnderTow`?
2. I'm not sure what is a purpose of assign a String to a variable and using it exactly once. But that's a side note, I respect your preference and you shouldn't feel obliged to change it.

michalvavrik (changes requested):

http/servlet-undertow/src/test/java/io/quarkus/ts/http/undertow/HttpServletWithSessionListenerIT.java
jcarranzan: Added 
```
httpClient.close();
vertx.close();
```

http/servlet-undertow/pom.xml
```diff
@@ -23,5 +23,19 @@
...
+            <artifactId>quarkus-vertx</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.vertx</groupId>
+            <artifactId>vertx-junit5</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
```
michalvavrik, line 37: Can you please confirm this dependency is required by removing this dependency and re-run tests?  I think it's strange addition and I'd like to understand it. Thanks

http/servlet-undertow/pom.xml
```diff
@@ -23,5 +23,19 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-security</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-vertx</artifactId>
```
michalvavrik, line 28: I suspect this dependency is only serving for your tests so it can be in the test scope?

http/servlet-undertow/src/test/java/io/quarkus/ts/http/undertow/HttpServletWithSessionListenerIT.java
```diff
@@ -51,6 +63,27 @@ public void sessionSecured() {
...
+                .compose(request -> request.send()
+                        .compose(httpClientResponse -> {
+                            assertEquals(HttpStatus.SC_OK, httpClientResponse.statusCode());
+                            assertEquals(HttpVersion.HTTP_2, httpClientResponse.version());
+                            return httpClientResponse.body();
+                        }))
+                .onSuccess(body -> vertxTestContext.verify(() -> {
+                    assertThat("Body response", body.toString().contains("Hello World"));
+                    vertxTestContext.completeNow();
+                    httpClient.close();
```
michalvavrik, line 82: so you are not going to close the `httpClient` and `vertx` on failure? We are re-running failed tests, are you sure this won't leave Vert.x unclosed during other tests? I suppose when:

- on success (done)
- on failure
- on timeout - by which I mean that you call vertxTestContext.completeNow, therefore I presume if yo never call it, test will eventually timeout. If you never receive response from backend, it will depend whether test timeout is shorter than Http client timeout. What is the default client timeout? I don't expect any so we need to take into consideration test timeout.

I don't think it is very important, but we can do it right because I think it shouldn't be hard. Let me know if you disagree.

http/servlet-undertow/src/test/java/io/quarkus/ts/http/undertow/HttpServletWithSessionListenerIT.java
```diff
@@ -51,6 +63,27 @@ public void sessionSecured() {
         thenCheckActiveSessionsEqualTo(1);
     }
 
+    @Test
+    @DisplayName("Undertow servlet test http2")
```
michalvavrik, line 66: It would be nice if you addressed my comments from the other PR here, considering it is same code and I've already suggested changes. Mainly `underTow` is not correct.

http/servlet-undertow/pom.xml
jcarranzan: yes put it in test scope

http/servlet-undertow/src/test/java/io/quarkus/ts/http/undertow/HttpServletWithSessionListenerIT.java
jcarranzan: I don't get it at all, I removed unnecessary test code related to Undertow in my other PR (https://github.com/quarkus-qe/quarkus-test-suite/pull/1655) and put it in this new PR.

http/servlet-undertow/src/test/java/io/quarkus/ts/http/undertow/HttpServletWithSessionListenerIT.java
```diff
@@ -51,6 +62,31 @@ public void sessionSecured() {
...
+    public void underTowHttp2Servlet(VertxTestContext vertxTestContext) {
```
michalvavrik, line 67:
```suggestion
    public void undertowHttp2Servlet(VertxTestContext vertxTestContext) {
```

http/servlet-undertow/src/test/java/io/quarkus/ts/http/undertow/HttpServletWithSessionListenerIT.java
```diff
@@ -51,6 +62,31 @@ public void sessionSecured() {
         thenCheckActiveSessionsEqualTo(1);
     }
 
+    @Test
+    @DisplayName("Undertow servlet test http2")
```
michalvavrik, line 66:
```suggestion
    @DisplayName("Undertow servlet test HTTP/2")
```

http/servlet-undertow/src/test/java/io/quarkus/ts/http/undertow/HttpServletWithSessionListenerIT.java
michalvavrik: if unit test times out, aftereach method is called; I'll close this, let's keep it.

http/servlet-undertow/src/test/java/io/quarkus/ts/http/undertow/HttpServletWithSessionListenerIT.java
```diff
@@ -51,6 +73,36 @@ public void sessionSecured() {
...
+        HttpClientOptions options = new HttpClientOptions().setProtocolVersion(HttpVersion.HTTP_2);
+        options.setSsl(true)
+                .setUseAlpn(true)
+                .setTrustOptions(new JksOptions().setPath("keystore.jks").setPassword("password"))
+                .setProtocolVersion(HttpVersion.HTTP_2);
+        httpClient = vertx.createHttpClient(options);
+        vertxTestContext
+                .verify(() -> {
+                    httpClient
+                            .request(HttpMethod.GET, app.getURI(Protocol.HTTPS).getPort(), baseUri.getHost(),
```
michalvavrik, line 90: `app.getURI(Protocol.HTTPS).getPort(), baseUri.getHost()` seems bit strange, to make var for host and not for port

michalvavrik:
seems ok

http/servlet-undertow/src/test/java/io/quarkus/ts/http/undertow/HttpServletWithSessionListenerIT.java
jcarranzan: I've removed that variable to use directly `app.getURI().getHost()`

http/servlet-undertow/src/test/java/io/quarkus/ts/http/undertow/HttpServletWithSessionListenerIT.java
```diff
@@ -90,4 +140,9 @@ private void thenWaitToEvictSessionsAndCheckActiveSessionsEqualTo(int value) {
                 .pollInterval(REST_ASSURANCE_POLL_INTERVAL)
                 .until(() -> getActiveSessions() == value);
     }
+
+    @AfterAll
+    public static void closeHttpClientConnection() {
+        httpClient.close();
```
michalvavrik, line 146: FYI - closing `httpClient` that is only created in one of 3 test methods without null check is wrong because during testing you can just run one of the test methods (e.g. when debugging stuff)

michalvavrik (approved):

http/servlet-undertow/src/test/java/io/quarkus/ts/http/undertow/HttpServletWithSessionListenerIT.java
jcarranzan: I understand your point but I think in this test class context the `httpClient` variable is static, so it persists across the execution of the test class,so in this context using `@AfterAll` and closing the `httpclient` connection is still valid, and you can execute the other tests independently and it won't produce any failure...

michalvavrik: you can't execute other tests independently without NPE, just try to run `sessionEviction`

jcarranzan: Screenshot from 2024-03-01 17-38-20

image

jcarranzan: Executed independently through IDE(IntelliJ) and with mvn command ` servlet-undertow]$ mvn clean verify -Dit.test=HttpServletWithSessionListenerIT#sessionSecured
` Both passed

michalvavrik: Thank you, I didn't know. NPE is thrown ans swallowed https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/extension/LifecycleMethodExecutionExceptionHandler.html#implementation-guidelines. This NPE would only have negative effect if there were other `@AfterAll` (multiple) methods, which there is not. So it okay-ish (I still don't like it...). Thanks

Issue #1397, opened 2023-09-05
Fix OpenShift scenario for transaction recovery on MySQL


michalvavrik:
`io.quarkus.ts.transactions.OpenShiftMysqlTransactionGeneralUsageIT` is disabled as it doesn't work. I can't tell why, because baremetal scenario works and it seems like it could be upstream issue (maybe even MySQL one), but it needs to be tested and analyzed (and possibly we need to create reproducer which needs to use MySQL XA driver directly, without our `CrashingXAResource` wrapper).

```
18:38:00,593 INFO  [app] 16:37:57,142 ARJUNA016027: Local XARecoveryModule.xaRecovery got XA exception XAException.XAER_RMERR: com.mysql.cj.jdbc.MysqlXAException: XAER_RMERR: Fatal error occurred in the transaction branch - check your data for consistency
18:38:00,594 INFO  [app] 	at com.mysql.cj.jdbc.MysqlXAConnection.mapXAExceptionFromSQLException(MysqlXAConnection.java:344)
18:38:00,594 INFO  [app] 	at com.mysql.cj.jdbc.MysqlXAConnection.recover(MysqlXAConnection.java:189)
18:38:00,594 INFO  [app] 	at com.mysql.cj.jdbc.MysqlXAConnection.recover(MysqlXAConnection.java:117)
18:38:00,594 INFO  [app] 	at io.quarkus.ts.transactions.recovery.driver.CrashingXAResource.recover(CrashingXAResource.java:82)
18:38:00,594 INFO  [app] 	at io.agroal.narayana.RecoveryXAResource.recover(RecoveryXAResource.java:31)
18:38:00,595 INFO  [app] 	at com.arjuna.ats.internal.jta.recovery.arjunacore.XARecoveryModule.xaRecoveryFirstPass(XARecoveryModule.java:711)
18:38:00,595 INFO  [app] 	at com.arjuna.ats.internal.jta.recovery.arjunacore.XARecoveryModule.periodicWorkFirstPass(XARecoveryModule.java:225)
18:38:00,595 INFO  [app] 	at com.arjuna.ats.internal.jta.recovery.arjunacore.XARecoveryModule.getNewXAResource(XARecoveryModule.java:333)
18:38:00,595 INFO  [app] 	at com.arjuna.ats.internal.jta.recovery.arjunacore.XARecoveryModule.getNewXAResource(XARecoveryModule.java:425)
[...]

PR #1665, opened 2024-02-15
Switch MySQL 5.7 to 8 as 5.7 is EOL

jedla97:
### Summary

Update docker image of MySQL as 5.7 reach EOL in October

- [x] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/sql-db/narayana-transactions/src/test/java/io/quarkus/ts/transactions/MysqlTransactionGeneralUsageIT.java
+++ b/sql-db/narayana-transactions/src/test/java/io/quarkus/ts/transactions/MysqlTransactionGeneralUsageIT.java
@@ -1,6 +1,11 @@
 package io.quarkus.ts.transactions;
 
+import static io.quarkus.test.services.containers.DockerContainerManagedResource.DOCKER_INNER_CONTAINER;
+
+import java.io.IOException;
+
 import org.junit.jupiter.api.Tag;
+import org.testcontainers.containers.GenericContainer;
 
 import io.quarkus.test.bootstrap.MySqlService;
 import io.quarkus.test.bootstrap.RestService;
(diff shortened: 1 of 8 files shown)


fedinskiy:
@jedla97 what about test framework?

Also, we probably need to run tests on Openshift

fedinskiy (commented):
@jedla97 what about test framework?

Also, we probably need to run tests on Openshift

jedla97:
run tests

jedla97:
> what about test framework?

In framework I only see MySQL 8 so it should be fine. Only place was here and beefy-scenarios

jedla97:
Issue seems related, looking at it

jedla97:
I look at it and disable it for the moment. It's the same issue as https://github.com/quarkus-qe/quarkus-test-suite/issues/1397 . There will need to be closer look how to fix that (automatically when spinning up image) but I describe what causing it there and how to fix it on own mysql instance.

fedinskiy (commented):
@jedla97 
1. The issue is for Openshift, but the test is disabled everywhere. I that expected?
2. There is a suggested workaround in the comments to the issue. Do you want to try to apply it?

jedla97:
> The issue is for Openshift, but the test is disabled everywhere. I that expected?

Yes it is as this is caused by MySQL permission setting.

> There is a suggested workaround in the comments to the issue. Do you want to try to apply it?

@fedinskiy Yes I'll try it and also believe that it will work. Will update this PR later this day and let you know

jedla97:
run tests

jedla97:
@fedinskiy I fix it on baremetal. For openshift it's more complicated. I need to look at it more but the scope of this PR is update the MySQL images which is done.

The last push after is just removing the one comment so I won't be running openshift test again

fedinskiy:
run tests

jedla97:
Looking at the errors and they are not related to this PR. GitHub CI is caused by bump of postgres in Quarkus (we see it in daily), the ocp one is seem like flaky test as it's root module which this PR not touching

jedla97:
run tests

jedla97:
@fedinskiy can you look at this, I rebase it and from my side it;s look ok. Openshift failures are not related

sql-db/narayana-transactions/src/test/java/io/quarkus/ts/transactions/MysqlTransactionGeneralUsageIT.java
```diff
@@ -15,9 +20,22 @@
...
 
     static final int MYSQL_PORT = 3306;
-
-    @Container(image = "${mysql.57.image}", port = MYSQL_PORT, expectedLog = "port: " + MYSQL_PORT)
-    static final MySqlService database = new MySqlService();
+    @Container(image = "${mysql.80.image}", port = MYSQL_PORT, expectedLog = "port: " + MYSQL_PORT)
+    static final MySqlService database = new MySqlService().onPostStart(service -> {
+        // enable transactions recovery for user
+        var self = (MySqlService) service;
+        // RH MySQL image allow login root user locally without using password and ignoring set root password
```
fedinskiy, line 27: Doesn't that mean, that we can always add password and do not bother ourselves with the ternary on the next line?

jedla97: I tried it, and RH build seems to ignore the password value totally as it's stated in doc/description. So we need this or more like I add it if we in future run it with upstream MySQL image (now only RH image is used by default).

Description of RH MySQL image states:

> The root user has no password set by default, only allowing local connections. You can set it by setting the MYSQL_ROOT_PASSWORD environment variable. This will allow you to login to the root account remotely. Local connections will still not require a password.

fedinskiy (approved):

PR #1701, opened 2024-03-06
Add keystore file type as it's now required

jedla97:
### Summary

It's the same problem as https://github.com/quarkus-qe/quarkus-test-suite/pull/1700 

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


michalvavrik (approved):
That is absolutely fine. I wonder if we shouldn't have rename it somewhere instead of specifying file type everywhere. Just so that we have file extension detection tested somewhere (I don't expect problems). That's just for extra points :-D

jedla97:
> I wonder if we shouldn't have rename it somewhere instead of specifying file type everywhere. Just so that we have file extension detection tested somewhere (I don't expect problems). That's just for extra points :-D

Looking and in https://github.com/quarkus-qe/quarkus-test-suite/blob/main/http/rest-client-reactive/src/main/resources/application.properties we don't specify type I don't see any error there. Probably we have this somewhere else also. If you meant for managment I can update these test to use `.jks`/`.p12`.

When looking at this I see that this `.keystore` is `PKCS12` type so I changing it.

PR #1741, opened 2024-04-03
Add coverage for Quarkus Security JPA extension

gtroitsk:
### Summary

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


michalvavrik (commented):
I am yet to have a deeper look, but I can already tell you that anything of this magnitude would give our CIs a big hit. 30 x custom native build.

There is absolutely no chance we can afford it, our GH CI is already near its limits and have also our Jenkins with it's limits too, though we can afford to be more creative there.

It's DEV Preview and we simply can't go in such a depth:

I suggest to have exactly one test per one database. So if you want to test Sha515 do it for Oracle for example, Bcrypt for MySQL, and so on.

I would go with:

- Oracle
- MySQL
- MariaDB

Because H2 is comprehensively tested in upstream test coverage and PG is tested in Quarkus Quickstarts, which we run as part of the acceptance pipeline.

gtroitsk:
> I am yet to have a deeper look, but I can already tell you that anything of this magnitude would give our CIs a big hit. 30 x custom native build.
> 
> There is absolutely no chance we can afford it, our GH CI is already near its limits and have also our Jenkins with it's limits too, though we can afford to be more creative there.
> 
> It's DEV Preview and we simply can't go in such a depth:
> 
> I suggest to have exactly one test per one database. So if you want to test Sha515 do it for Oracle for example, Bcrypt for MySQL, and so on.
> 
> I would go with:
> 
> * Oracle
> * MySQL
> * MariaDB
> 
> Because H2 is comprehensively tested in upstream test coverage and PG is tested in Quarkus Quickstarts, which we run as part of the acceptance pipeline.

@michalvavrik Appreciate your notes. Let's keep test sha256, sha512 and md5  with MariaDB, Oracle, MySQL in our TS, overall 3 tests. Other scenarious except plaintext are covered in upstream and QS. I will update TP

security/jpa/src/test/resources/import.sql
```diff
@@ -0,0 +1,8 @@
+INSERT INTO md5_entity (id, username, password, role) VALUES (1, 'admin', '21232f297a57a5a743894a0e4a801fc3', 'admin');
+INSERT INTO md5_entity (id, username, password, role) VALUES (2, 'user','ee11cbb19052e40b07aac0ca060c23ee', 'user');
```
michalvavrik, line 2: Well, you can't say I didn't warn you that if you request my review I'll want also scenario where you create the user :-)

What you test is fine, but please, test whole flow, users are not a static data and there is a good change that application will create these users, so I think you should have a smoke test that persist new user and then uses it. Nothing complicated.

Choice is yours, but for example (do what you want!) I would just replace one of imported users, like create 'user' with 'admin' while 'admin' is important in this SQL.

Point behind this request: test real-world scenarios.

michalvavrik (commented):
Looks good, I have only minor requests:

1. Security JPA identity provider blocks, your endpoints only return things in a synchronous manner. Some of resources should either return reactive data type or be annotated with `NonBlocking` so that we know blocking is performed correctly and doesn't cause issues when on IO thread.
2. You only use standard security annotations (which is fine and mainstream use case), however authorization using configuration is performed earlier and that is relevant here is it runs on IO thread. Please add a smoke test. One of the HTTP requests should be secured by roles-policy https://quarkus.io/guides/security-authorize-web-endpoints-reference#authorization-using-configuration.
3. Add a smoke test for the `io.quarkus.security.jpa.RolesValue` annotation

security/jpa/src/test/resources/import.sql
michalvavrik: It might also be of interested to delete user, but we can leave it for now as it's a dev preview. Your call.

michalvavrik: Correction: you didn't request my review, it was Rostislav :-D

README.md
```diff
@@ -627,6 +627,11 @@ Bouncy castle providers:
 Verifies form-based authentication.
 Verifies that Basic Authentication is not used as a fallback option when it is explicitly disabled.
 
+### `security/jpa`
+
+Verifies Basic authentication with Jakarta Persistence identity provider, enabling role-based access control.
+User password is hashed by one of algorithm: MD5, SHA-256, SHA-512.
```
michalvavrik, line 633:
```suggestion
User password is hashed by one of algorithms: MD5, SHA-256, SHA-512.
```

? not sure TBH

security/jpa/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkus.ts.qe</groupId>
+        <artifactId>parent</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+        <relativePath>../..</relativePath>
+    </parent>
+    <artifactId>security-jpa</artifactId>
+    <packaging>jar</packaging>
+    <name>Quarkus QE TS: Security: JPA</name>
```
michalvavrik, line 12:
```suggestion
    <name>Quarkus QE TS: Security: Security Jakarta Persistence</name>
```

security/jpa/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-rest</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-hibernate-orm-panache</artifactId>
```
michalvavrik, line 20:
```suggestion
            <artifactId>quarkus-hibernate-orm</artifactId>
```
AFAICT you are not using Panache anywhere.

michalvavrik (approved):

michalvavrik:
Nice!

gtroitsk:
@michalvavrik CI is green. Can we merge this PR?

PR #1759, opened 2024-04-22
[3.8] Do not expect index page in DEV mode on Windows

michalvavrik:
### Summary

The check in question returns 404 on Windows due to https://github.com/quarkusio/quarkus/issues/39909#issuecomment-2070739091. It seems like it's related to the CLI version installed on our Windows runners with a Docker. I didn't look further into cause of this behavior because I do not consider this very important as long as there is opened known issue marked as won't f i x https://github.com/quarkusio/quarkus/issues/39909.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/QuarkusCliCreateJvmApplicationIT.java
+++ b/quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/QuarkusCliCreateJvmApplicationIT.java
@@ -79,7 +79,13 @@ public class QuarkusCliCreateJvmApplicationIT {
 
         // Start using DEV mode
         app.start();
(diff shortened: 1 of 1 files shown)


jedla97 (approved):
Thanks for PR and some investigation!

We should probably look into this when there will be more time as what I remeber [codestart.quarkus](https://code.quarkus.io/) put blank page even on linux and I would expect the same with cli. I'll create task for this

michalvavrik:
> We should probably look into this when there will be more time as what I remeber [codestart.quarkus](https://code.quarkus.io/) put blank page even on linux and I would expect the same with cli. I'll create task for this

true, it will need proper analysis, thanks

PR #1818, opened 2024-06-04
Coverage for several issues, which will be fixed in 3.8.5 via backporting

fedinskiy:
### Summary

The issues:
- https://github.com/quarkusio/quarkus/pull/40718
- https://github.com/quarkusio/quarkus/issues/37323
- https://github.com/quarkusio/quarkus/issues/40307

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/security/basic/pom.xml
+++ b/security/basic/pom.xml
@@ -13,11 +13,16 @@
     <dependencies>
         <dependency>
             <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-resteasy</artifactId>
+            <artifactId>quarkus-resteasy-reactive</artifactId>
+            <!-- Only reactive dependency can cause https://github.com/quarkusio/quarkus/issues/40307-->
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-elytron-security-properties-file</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-reactive-routes</artifactId>
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
--- a/security/basic/src/main/java/io/quarkus/ts/openshift/security/basic/SecurityFilter.java
+++ b/security/basic/src/main/java/io/quarkus/ts/openshift/security/basic/SecurityFilter.java
@@ -1,0 +1,16 @@
+package io.quarkus.ts.openshift.security.basic;
+
+import jakarta.inject.Singleton;
+
+import io.quarkus.vertx.web.RouteFilter;
+import io.vertx.ext.web.RoutingContext;
+
+@Singleton
+// if https://github.com/quarkusio/quarkus/issues/40307 is not fixed, then presence of a filter leads to 500 error
+public class SecurityFilter {
+
+    @RouteFilter(401)
+    void filter(RoutingContext routingContext) {
+        routingContext.next();
+    }
+}
--- a/http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/failures/FailingClientResource.java
+++ b/http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/failures/FailingClientResource.java
@@ -1,0 +1,31 @@
+package io.quarkus.ts.http.restclient.reactive.failures;
+
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.MediaType;
+
+import org.eclipse.microprofile.rest.client.inject.RestClient;
+
+import io.smallrye.mutiny.Uni;
+
+@Path("/client/failing")
+public class FailingClientResource {
+    @RestClient
+    FailureClient client;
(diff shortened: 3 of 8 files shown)


rsvoboda:
This mono commit doesn't follow items checked in description:
Screenshot 2024-06-05 at 11 24 15

 - There are 3 topics handled by this PR, there should be 3 commits.
 - Commit message doesn't give any idea about the concrete scenarios, I don't really care the stuff get there because of 3.8.5 release, commit message should cover highlight of the change
 - I miss explanatory failure messages for the persona facing the test failure
 - Test methods/classes could have a comment or JavaDoc entry to give a bit of context

security/basic/src/main/java/io/quarkus/ts/openshift/security/basic/SecurityFilter.java
```diff
@@ -0,0 +1,22 @@
...
+import io.quarkus.vertx.web.RouteFilter;
+import io.vertx.ext.web.RoutingContext;
+
+@Singleton
+public class SecurityFilter {
+
+    @RouteFilter(401)
+    void addAuthHeader(RoutingContext routingContext) {
+        try {
+            if (routingContext.request().headers().get("Sec-WebSocket-Protocol") != null) {
```
mocenas, line 14: What is a purpose of this? I don't understand it. 
If I understood the issue https://github.com/quarkusio/quarkus/issues/40307 correctly, then this is just a workaround for issue that is fixed in 3.10.1. Is that correct?
If yes, do we really want to have a workaround in our TS? At least, it should have a comment linking the issue, fix and explaining it.
WDYT @michalvavrik ?

mocenas (changes requested):
I'm not sure about the security changes.

security/basic/src/main/java/io/quarkus/ts/openshift/security/basic/SecurityFilter.java
michalvavrik: it never had anything to do with websockets. https://github.com/quarkusio/quarkus/issues/40307 is about existence of `@RouteFilter(404)`. You can have it empty and it will still be a reproducer. What you want to do is to have Reactive Route filter as then, Reactive Routes activates CDI request context.

security/basic/pom.xml
```diff
@@ -13,12 +13,16 @@
     <dependencies>
         <dependency>
             <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-resteasy</artifactId>
+            <artifactId>quarkus-resteasy-reactive</artifactId>
```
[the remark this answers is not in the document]
michalvavrik: it should be `quarkus-rest` (no reactive in name); I think Fedor just want to avoid changes in backports :-) I would prefer to avoid `quarkus-resteasy-reactive` in main branch, no other module uses it. But I am not a reviewer here, so I'll leave you to decide.

security/basic/src/main/java/io/quarkus/ts/openshift/security/basic/SecurityFilter.java
fedinskiy: As Michal noted correctly, this is not a workaround, but condition for an issue to be reproduced. The condition, on the other hand is a workaround for Websockets, but that is another story :)

I will try to clean the filter from everything that is not required for the bug to resurface, though

fedinskiy: Everything related to websockets was removed from the filter

security/basic/pom.xml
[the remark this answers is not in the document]
mocenas: Still IMHO the issue should be linked here, or put some comment here, why it needs to be reactive. Otherwise someone will be very confused some day, when changing something here.

security/basic/src/main/java/io/quarkus/ts/openshift/security/basic/SecurityFilter.java
```diff
@@ -0,0 +1,15 @@
+package io.quarkus.ts.openshift.security.basic;
+
+import jakarta.inject.Singleton;
+
+import io.quarkus.vertx.web.RouteFilter;
+import io.vertx.ext.web.RoutingContext;
+
+@Singleton
+public class SecurityFilter {
```
mocenas, line 9: Please add explanation, why is the filter here (ideally link the issue + comment that issue only appears if security filter is here).
There is a piece of code, that does nothing on a first glance. It would be really good to have an explanation why is it necessary.

mocenas (changes requested):
Please add comments/link about the security issue.
Otherwise it looks good.

rsvoboda:
@mocenas waiting for feedback from @michalvavrik before merging?

mocenas:
> @mocenas waiting for feedback from @michalvavrik before merging?

I don't think @michalvavrik was requested for feedback any more. IMHO we can merge it now.

michalvavrik:
@fedinskiy can you follow-up after this PR is merged with change of `quarkus-resteasy-reactive` to `quarkus-rest`? This makes it only place were we need relocation, which is not a problem for CI, but I normally don't do them. There is no reason.

Issue #1832, opened 2024-06-06
Test gRPC DEV UI messaging


michalvavrik:
The gRPC DEV UI allows to test gRPC client - send a message from DEV UI to gRPC server and receive answer. We have done it with the old DEV UI, but to do same with a new DEV UI would be very fragile. There are no hooks for selectors and with the Playwright it's harder to navigate in shadow roots. First, we need to change Quarkus main project and add there id's or classes that we can use as selectors during testing and then, we can start testing it again.

See:

- https://github.com/quarkus-qe/quarkus-test-suite/blob/main/http/http-advanced/src/test/java/io/quarkus/ts/http/advanced/DevModeGrpcIntegrationIT.java
- https://github.com/quarkus-qe/quarkus-test-suite/blob/main/http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/DevModeGrpcIntegrationReactiveIT.java

PR #1833, opened 2024-06-07
Fix and enable DEV UI gRPC tests

michalvavrik:
### Summary

Enables disabled gRPC DEV UI tests. The test differs because the new DEV UI is SPA and cannot be tested in a same way as static pages. It's bit hard to test it as there are no hooks we could use for selectors, so I reimplemented gRPC service assertion and opened https://github.com/quarkus-qe/quarkus-test-suite/issues/1832 for gRPC communication (follow-up). Another thing that has changed is that we use gRPC server on a same HTTP server (not a separate port as in past), but our FW is not adapted to this, so I only worked around it for now. I'll address it later in FW.

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/http/http-advanced-reactive/pom.xml
+++ b/http/http-advanced-reactive/pom.xml
@@ -98,5 +98,10 @@
             <groupId>com.aayushatharva.brotli4j</groupId>
             <artifactId>brotli4j</artifactId>
         </dependency>
+        <dependency>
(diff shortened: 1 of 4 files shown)


jedla97 (approved):
LGTM!

Failure is not related. Daily have same failure.

PR #1853, opened 2024-06-25
Add test coverage for improvement multipart encoded mode in rest-client

jcarranzan:
### Summary

Adding TD for the next backport:
[Improve the multipart encoded mode handling in the rest client](https://github.com/quarkusio/quarkus/pull/39770)

This PR adds test coverage for the improved multipart encoder mode handling in the REST client. 
The tests cover the encoder modes (HTML5, RFC1738, RFC3986), verifying the correct encoding of multipart requests and successful parsing by a Vert.x server. Data integrity checks, including file content verification via size and headers, are performed.

Note: Direct verification of the specific encoder mode used during the transmission is not directly feasible, but 
the tests ensure that the resulting content is correctly parsed and interpreted by the server, providing confidence that the appropriate encoder mode was applied.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/http/rest-client-reactive/pom.xml
+++ b/http/rest-client-reactive/pom.xml
@@ -41,5 +41,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-smallrye-fault-tolerance</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.vertx</groupId>
+            <artifactId>vertx-junit5</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/multipart/EncoderModeRestClient.java
+++ b/http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/multipart/EncoderModeRestClient.java
@@ -1,0 +1,26 @@
+package io.quarkus.ts.http.restclient.reactive.multipart;
+
+import static jakarta.ws.rs.core.MediaType.MULTIPART_FORM_DATA;
+
+import jakarta.ws.rs.Consumes;
+import jakarta.ws.rs.FormParam;
+import jakarta.ws.rs.POST;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.MediaType;
+
+import org.jboss.resteasy.reactive.PartType;
+import org.jboss.resteasy.reactive.RestForm;
+
+@Path("/encoder-mode")
+public interface EncoderModeRestClient {
+
+    @POST
+    @Consumes(MULTIPART_FORM_DATA)
+    @Produces(MediaType.APPLICATION_JSON)
+    MyMultipartDTO doAPostRequestToThisResource(
+            @PartType(MediaType.TEXT_PLAIN) @FormParam("file1") java.nio.file.Path file1,
+            @PartType(MediaType.TEXT_PLAIN) @FormParam("file2") java.nio.file.Path file2,
+            @RestForm String otherField);
+
+}
--- a/http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/multipart/Html5EncoderModeRestClient.java
+++ b/http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/multipart/Html5EncoderModeRestClient.java
@@ -1,0 +1,8 @@
+package io.quarkus.ts.http.restclient.reactive.multipart;
+
+import org.eclipse.microprofile.rest.client.inject.RegisterRestClient;
+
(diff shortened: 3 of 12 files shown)


rsvoboda (changes requested):
Hi @jcarranzan.

I did the first quick check of this PR and there are some things I would like to have:

 - PR description doesn't add too much details about the test coverage, it points to the PR where I can get to the discussion, but that's 2 hops, short summary would make the life better for the reviewer
 - Short scenario summary should be present in HttpSimpleEncodeModeIT as JavaDoc, people from the team will be looking into this class sooner or later and having some context is better than needing to go through the commits history and reading the code
 - Having stuff in `io.quarkus.ts.http.restclient.reactive` is not intuitive as you are mixing new classes with the original ones, put new classes into `io.quarkus.ts.http.restclient.reactive.multipart`. Test can stay in the same package as it is now.
 - What's the motivation to use Vert.x server as the second server and not 2 Quarkus applications, just curious, not asking to rework stuff
 - I didn't notice any check to verify that the different `encoderMode` was used (not sure there is an easy way to do, but this is a thing that could be mentioned in PR description that it was considered), the test is checking that the correct content was received, but not how it was encoded ... in theory the multipart could be encoded using the same strategy all the time

jcarranzan:
> Hi @jcarranzan.
> 
> I did the first quick check of this PR and there are some things I would like to have:
> 
>     * PR description doesn't add too much details about the test coverage, it points to the PR where I can get to the discussion, but that's 2 hops, short summary would make the life better for the reviewer
> 
>     * Short scenario summary should be present in HttpSimpleEncodeModeIT as JavaDoc, people from the team will be looking into this class sooner or later and having some context is better than needing to go through the commits history and reading the code
> 
>     * Having stuff in `io.quarkus.ts.http.restclient.reactive` is not intuitive as you are mixing new classes with the original ones, put new classes into `io.quarkus.ts.http.restclient.reactive.multipart`. Test can stay in the same package as it is now.
> 
>     * What's the motivation to use Vert.x server as the second server and not 2 Quarkus applications, just curious, not asking to rework stuff
> 
>     * I didn't notice any check to verify that the different `encoderMode` was used (not sure there is an easy way to do, but this is a thing that could be mentioned in PR description that it was considered), the test is checking that the correct content was received, but not how it was encoded ... in theory the multipart could be encoded using the same strategy all the time

I agree, I will improve it with your suggestions, thanks @rsvoboda .

michalvavrik:
> I didn't notice any check to verify that the different encoderMode was used ......

All that is true, but it's not possible to check this. Netty doesn't provide any API access to raw message that could give us difference. Decoder result is always true. IMO:

- only way to check **specific** mode is to have access to lower lever, so that we could do check similar to this: https://github.com/netty/netty/blob/151dfa083d28e995a18f7d2c73d4a7d3b7ab73b2/codec-http/src/test/java/io/netty/handler/codec/http/multipart/HttpPostRequestEncoderTest.java#L264
  - it will require different stack or some method I missed
- only way to check that **different** mode is used when different configuration property is set would be to reproduce this https://github.com/quarkusio/quarkus/discussions/39751#discussioncomment-8939486, however it works for multiple files regardless of modes, so there must be something else, like "you cannot reproduce it without undertow" or something; this can be done

michalvavrik:
hmm, I wonder if you could test this with Wiremock? Let's try that.

jcarranzan:
I've improved the next steps:

- PR description
- JavaDoc in the test class
- Rename of the test class

Answering some questions here:

1. What's the motivation to use Vert.x server as the second server and not 2 Quarkus applications...

 _While initially, it seemed possible to verify the encoding mode directly using the Vert.x server as a lightweight option, further investigation didn't reveal a straightforward approach. I didn' consider about 2 Quarkus application, but yes could be other option._

2. @rsvoboda  about the last point "I didn't notice any check to verify that the different `encoderMode` was used (not sure there is an easy way to do, but this is a thing that could be mentioned in PR description that it was considered), the test is checking that the correct content was received, but not how it was encoded ... in theory the multipart could be encoded using the same strategy all the time" 

_It's true,  directly inspecting the encoding mode in transit isn't straightforward, and I though checking  the structure of the received multipart content was enough. I also tried the reproduced described here https://github.com/quarkusio/quarkus/discussions/39751#discussioncomment-8939486 but I noticed some things as server-undertow, etc that was out of scope of this testing._

rsvoboda:
What about proposed Wiremock approach? That could be a follow-up, not necessary to be in this PR.

rsvoboda (approved):
I think this is good enough.

I have 2 suggestions for the future improvement if you feel like it:
 - content of the uploaded files could be checked ... extend the DTO and store there `buffer.toString()`
 - add negative scenario - e.g. with incorrect `multipart-post-encoder-mode`

jcarranzan:
Ok now I've added those last 2 suggestions (check for the content of the uploaded files.

PR #1861, opened 2024-06-27
[3.8] Use Red Hat PostgreSQL image in OpenShift and fix OpenShiftPostgresqlTransactionGeneralUsageIT when its run with PostgreSQL image from Red Hat registry

michalvavrik:
### Summary

- IBM PE team runs this test with `registry.access.redhat.com/rhel8/postgresql-16` (also because of Docker pull rate, but it's right way anyway). They set it as system property via additional maven arguments. This image doesn't accept `--max_prepared_transactions=100` and fails with `/usr/bin/container-entrypoint: line 3: exec: --: invalid option`. Documented way https://catalog.redhat.com/software/containers/rhel9/postgresql-16/657b03866783e1b1fb87e142?architecture=amd64&image=66612ca95ea3c2839df21ce1 is to use `POSTGRESQL_MAX_PREPARED_TRANSACTIONS` environment variable. It also works with PG 15 Red Hat image, but it doesn't work with images from Docker hub. 
- our own OpenShift tests were using Docker hub image instead of the Red Hat one, even though we did set the RH one. The reason for that was that effective POM prioritize execution-specific system variables over non-execution specific. Therefore I moved them inside the OpenShift profile into the execution

Therefore this PR:

- makes sure that image from Red Hat image registry is always used in OpenShift (main OpenShift profile already did that, but IBM and ARM64 profiles didn't)
- use the environment property for Red Hat images
- PostgreSQL container image from the Red Hat registry is passing before starting to the logs directory instead to STDOUT, so I opened https://github.com/quarkus-qe/quarkus-test-framework/issues/1183 and changed waiting for OpenShift tests to the last line logged to STDOUT
  - ad safety - it takes about second to start it after logging to log dir only while pushing image to OpenShift and starting app there takes longer so it's not a big deal but we should address it

Hint: there are 2 commits, one can be reverted when FW issue is fixed

P.S. doing all of this in the 3.8 as we are processing release

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/external-applications/src/test/java/io/quarkus/ts/external/applications/AbstractTodoDemoIT.java
+++ b/external-applications/src/test/java/io/quarkus/ts/external/applications/AbstractTodoDemoIT.java
@@ -1,0 +1,43 @@
+package io.quarkus.ts.external.applications;
+
+import static org.hamcrest.Matchers.is;
+
+import org.apache.http.HttpStatus;
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.annotations.DisabledOnNative;
+import io.restassured.http.ContentType;
+
+@DisabledOnNative(reason = "This scenario is using uber-jar, so it's incompatible with Native")
+public abstract class AbstractTodoDemoIT {
+
+    protected static final String TODO_REPO = "https://github.com/quarkusio/todo-demo-app.git";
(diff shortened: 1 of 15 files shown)


michalvavrik:
run tests

mjurc (approved):
LGTM, thanks for the fix

rsvoboda:
What about main? Is there a PR already or you want FW fix first?

michalvavrik:
> What about main? Is there a PR already or you want FW fix first?

I need this to be ready for IBM in 3.8. Port is here https://github.com/quarkus-qe/quarkus-test-suite/pull/1862. I'll address FW when I have QQE-710 assigned again.

PR #1868, opened 2024-07-02
Quarkus CLI update test

mocenas:
### Summary
This PR does not work, as it requires https://github.com/quarkus-qe/quarkus-test-framework/pull/1188 to be effective here.

This makes a proposal how Quarkus CLI test could look in the future. Idea is to create common tests and other support methods in parent class and making some kind of a "framework". For each new tested version we can add new class, which should be more or else simple.

Edit: This PR requires https://github.com/quarkus-qe/quarkus-test-framework/pull/1236 and other things in framework 1.5.0.Beta13. **So do not merge until this is released and updated in TS.** 

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/update/AbstractQuarkusCliUpdateIT.java
+++ b/quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/update/AbstractQuarkusCliUpdateIT.java
@@ -1,0 +1,82 @@
+package io.quarkus.ts.quarkus.cli.update;
+
+import static io.quarkus.test.util.QuarkusCLIUtils.getQuarkusAppVersion;
+import static io.quarkus.test.utils.AwaitilityUtils.untilAsserted;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.io.IOException;
+import java.time.Duration;
+
+import jakarta.inject.Inject;
+
+import org.apache.http.HttpStatus;
+import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
+import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.bootstrap.QuarkusCliClient;
+import io.quarkus.test.bootstrap.QuarkusCliRestService;
+import io.quarkus.test.logging.Log;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.scenarios.annotations.DisabledOnNative;
+import io.quarkus.test.services.quarkus.model.QuarkusProperties;
+import io.quarkus.test.util.DefaultQuarkusCLIAppManager;
+import io.quarkus.test.util.IQuarkusCLIAppManager;
+import io.quarkus.test.utils.AwaitilityUtils;
+
+@QuarkusScenario
+@DisabledOnNative // Only for JVM verification
+public abstract class AbstractQuarkusCliUpdateIT {
+    @Inject
+    static QuarkusCliClient cliClient;
+
+    protected final DefaultArtifactVersion oldVersionStream;
+    protected final DefaultArtifactVersion newVersionStream;
+    protected final String newVersionFromProperties;
+    protected final IQuarkusCLIAppManager quarkusCLIAppManager;
+
+    public AbstractQuarkusCliUpdateIT(DefaultArtifactVersion oldVersionStream, DefaultArtifactVersion newVersionStream) {
+        this.oldVersionStream = oldVersionStream;
+        this.newVersionStream = newVersionStream;
+
+        // takes quarkus.platform.version from maven parameters. If present, it will update to this exact BOM version
+        // otherwise it will default to update to stream
+        this.newVersionFromProperties = QuarkusProperties.getVersion();
+        this.quarkusCLIAppManager = createAppManager();
+    }
+
+    protected IQuarkusCLIAppManager createAppManager() {
+        if (this.newVersionFromProperties != null && this.newVersionFromProperties.contains("redhat")) {
+            return new RHBQPlatformAppManager(cliClient, oldVersionStream, newVersionStream,
+                    new DefaultArtifactVersion(newVersionFromProperties));
+        }
+        return new DefaultQuarkusCLIAppManager(cliClient, oldVersionStream, newVersionStream);
+    }
+
+    /**
+     * Perform basic update to new stream and check version in pom.xml matches
+     */
+    @Test
+    public void versionUpdateTest() throws IOException, XmlPullParserException {
+        QuarkusCliRestService app = quarkusCLIAppManager.createApplication();
+
+        quarkusCLIAppManager.updateApp(app);
+
+        DefaultArtifactVersion updatedVersion = getQuarkusAppVersion(app);
+        assertEquals(newVersionStream.getMajorVersion(), updatedVersion.getMajorVersion(),
+                "Major version for app updated to " + newVersionStream + "should be " + newVersionStream.getMajorVersion());
+        assertEquals(newVersionStream.getMinorVersion(), updatedVersion.getMinorVersion(),
+                "Minor version for app updated to " + newVersionStream + " should be " + newVersionStream.getMinorVersion());
+        // check that updated app is using RHBQ
+        assertTrue(updatedVersion.toString().contains("redhat"),
+                "Updated app is not using \"redhat\" version. Found version: " + updatedVersion);
+
+        Log.info("Starting updated app");
+        // start the updated app and verify that basic /hello endpoint works
+        app.start();
+        untilAsserted(() -> app.given().get("/hello").then().statusCode(HttpStatus.SC_OK),
+                AwaitilityUtils.AwaitilitySettings.usingTimeout(Duration.ofSeconds(10))
+                        .timeoutMessage("Updated app failed to expose working /hello endpoint"));
+    }
+}
--- a/quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/update/RHBQPlatformAppManager.java
+++ b/quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/update/RHBQPlatformAppManager.java
@@ -1,0 +1,30 @@
+package io.quarkus.ts.quarkus.cli.update;
+
+import static io.quarkus.test.bootstrap.QuarkusCliClient.UpdateApplicationRequest.defaultUpdate;
+
+import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
+
+import io.quarkus.test.bootstrap.QuarkusCliClient;
+import io.quarkus.test.bootstrap.QuarkusCliRestService;
+import io.quarkus.test.logging.Log;
+import io.quarkus.test.util.DefaultQuarkusCLIAppManager;
+
+/**
+ * AppManager designed to update app to specific quarkus-bom version, instead of to stream as DefaultQuarkusCLIAppManager
+ */
+public class RHBQPlatformAppManager extends DefaultQuarkusCLIAppManager {
+    protected final DefaultArtifactVersion newPlatformVersion;
+
+    public RHBQPlatformAppManager(QuarkusCliClient cliClient, DefaultArtifactVersion oldStreamVersion,
+            DefaultArtifactVersion newStreamVersion, DefaultArtifactVersion newPlatformVersion) {
+        super(cliClient, oldStreamVersion, newStreamVersion);
+        this.newPlatformVersion = newPlatformVersion;
+    }
+
+    @Override
+    public void updateApp(QuarkusCliRestService app) {
+        Log.info("Updating app to version: " + newPlatformVersion);
+        app.update(defaultUpdate()
+                .withPlatformVersion(newPlatformVersion.toString()));
+    }
+}


quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/update/AbstractQuarkusCliUpdateIT.java
```diff
@@ -0,0 +1,138 @@
...
+
+        Log.info("Starting updated app");
+        // start the updated app and verify that basic /hello endpoint works
+        app.start();
+        untilAsserted(() -> app.given().get("/hello").then().statusCode(HttpStatus.SC_OK),
+                AwaitilityUtils.AwaitilitySettings.usingTimeout(Duration.ofSeconds(10))
+                        .timeoutMessage("Updated app failed to expose working /hello endpoint"));
+    }
+
+    protected void checkPropertiesUpdate(Properties oldProperties, Properties expectedNewProperties) throws IOException {
```
michalvavrik, line 74: this task will repeat therefore I'd prefer to have it in the FW.
If you know that you:

- have app
- run update
- expect properties to be xyz

then it should be one method call with "app" and "expected property after", "expected property missing" etc. all depends on common task

michalvavrik:
run tests

michalvavrik:
please ignore `run tests` comment, I just need to test OCP CI and there was no better PR atm.

quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/update/AbstractQuarkusCliUpdateIT.java
mocenas: I'm not sure if this should be shifted into the framework. Do we expect to test the updates elsewhere?
I don't like the idea of splitting the test logic into two projects. And this method seems to me to be tightly integrated with the test logic.
Also having this somehow complicated method (it needs, to create app, update it - all with versions and app's seteup, run asserts, ...) in a framework would be IMHO very impractical if we want to change the methods behaviour or ad hoc disable failing assert or something in the test suite.

michalvavrik: well, I obviously disagree, but you are at Helm, follow your guts. if it should be in TS, let it be in TS. please choose different reviewer as I am OOO.

michalvavrik: cc @rsvoboda

mocenas:
I've added actual tests. @gtroitsk Please have a look at them.

gtroitsk:
> I've added actual tests. @gtroitsk Please have a look at them.

I will look at it later today

gtroitsk (commented):
I afraid that actual TC is not enough. I think we need to test wider set of changes, scope is up to you. Maybe Michal V. or Rostislav have some comments. 
For testing properties ideally we should create real app reflecting the customer's usage scenario including test coverage (taken from TP). We can't relate that new names are syntactically correct. 
Also can you please add an option/tests for RHBQ apps.

quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/update/AbstractQuarkusCliUpdateIT.java
rsvoboda: Hey. +1 to have this in framework / [QuarkusCLIUtils.java](https://github.com/quarkus-qe/quarkus-test-framework/blob/main/quarkus-test-cli/src/main/java/io/quarkus/test/util/QuarkusCLIUtils.java)

We are not the only consumers of test framework, Camel Quarkus QE team is using it too.

mocenas:
> I afraid that actual TC is not enough. I think we need to test wider set of changes, scope is up to you. Maybe Michal V. or Rostislav have some comments. For testing properties ideally we should create real app reflecting the customer's usage scenario including test coverage (taken from TP). We can't relate that new names are syntactically correct. Also can you please add an option/tests for RHBQ apps.

I agree that we should tests entire apps. I'm working on it (created first app for hibernate update). I plan to add more for methods and packages renames.

Could you be more specific what would you like to see as TC "wider set of changes"? I'm following recipes and migration guides and testing stuff from them.

Yes I will add RBHQ updates.

gtroitsk:
> > I afraid that actual TC is not enough. I think we need to test wider set of changes, scope is up to you. Maybe Michal V. or Rostislav have some comments. For testing properties ideally we should create real app reflecting the customer's usage scenario including test coverage (taken from TP). We can't relate that new names are syntactically correct. Also can you please add an option/tests for RHBQ apps.
> 
> I agree that we should tests entire apps. I'm working on it (created first app for hibernate update). I plan to add more for methods and packages renames.
> 
> Could you be more specific what would you like to see as TC "wider set of changes"? I'm following recipes and migration guides and testing stuff from them.
> 
> Yes I will add RBHQ updates.

Under TC "wider set of changes" I meant test as must as possible recipes and changes from migration guides. You follow it, so I am fine.

mocenas:
Testing `quarkus udpate` on RHBQ has a problem that I didn't forsee. Updating between two RHBQ versions requires to have both their repos available and (AFAIK) in single local maven repo. This doesn't fit our workflow for quarkus-test-suite and IMHO it is not worth it to change the workflow.

I think it would be better to drop this PR and move update tests into separate GH repo, for which we can setup appropriate workflow. WDYT @gtroitsk @rsvoboda @michalvavrik ?

michalvavrik:
> Testing quarkus udpate on RHBQ has a problem that I didn't forsee. Updating between two RHBQ versions requires to have both their repos available and (AFAIK) in single local maven repo. This doesn't fit our workflow for quarkus-test-suite and IMHO it is not worth it to change the workflow.

If that's a true it's a problem. I'll be honesty, I don't understand why updating from released RHBQ to RHBQ that is being released would required the released RHBQ locally. I'd expect that update command use remote Maven repository. Can you clarify this a bit?

mocenas:
> > Testing quarkus udpate on RHBQ has a problem that I didn't forsee. Updating between two RHBQ versions requires to have both their repos available and (AFAIK) in single local maven repo. This doesn't fit our workflow for quarkus-test-suite and IMHO it is not worth it to change the workflow.
> 
> If that's a true it's a problem. I'll be honesty, I don't understand why updating from released RHBQ to RHBQ that is being released would required the released RHBQ locally. I'd expect that update command use remote Maven repository. Can you clarify this a bit?

 I didn't check but `quarkus update` probably can work with remote maven repo. But:
1. I though we want this to work also with unreleased bits.
2. AFAIK goal is to test if update works with bits that we ship. And do we have some repos with RHBQ bits we ship, and ONLY those bits? I already found an issue with some artifact missing in released repo which caused `quarkus update` to fail.

michalvavrik:
But this is not right platform to discuss this. Let's move this to Slack. I don't want to be more specific here.

quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/update/AbstractQuarkusCliUpdateIT.java
mocenas: PR for these test methods into framework - https://github.com/quarkus-qe/quarkus-test-framework/pull/1239

mocenas: methods moved to framework.

mocenas:
After updating local quarkus CLI setup to work with RHBQ, current test design work surprisingly well with RHBQ.

mocenas:
I've added support for testing updates to specific bom version.

mocenas:
Can you please review this @gtroitsk @michalvavrik ? The tests should be done, I'm keeping it as a draft before framework is released.
PS: I know it seems that a lot of migration changes are not covered, but these are usually new feature/deprecated/removed/breaking change/etc, which means how have to manually update you app anyway, so I don't see way in testing it.

quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/update/Quarkus32to38CliUpdateIT.java
```diff
@@ -0,0 +1,263 @@
...
+                "quarkus-rest-client-jackson should be updated to quarkus-resteasy-client-jackson in child project");
+
+        // TODO: it would be good to actually run the updated app, but FW currently cannot run multi module app
+    }
+
+    /**
+     * Update resteasy app, while also upgrading it from java 11 to 17
+     */
+    @Test
+    public void updateAndRunApp() throws XmlPullParserException, IOException {
```
michalvavrik, line 232: `old` app (the one you are updating) uses RHBQ, it would be nice to check that new one is still RHBQ.

michalvavrik (commented):
I skimmed through changes, they LGTM. I am afraid I don't have proper time to in depth review in the next week and half. AFIC it's ok. Maybe you will need additional reviewer. Thanks!

quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/update/Quarkus213to38CliUpdateIT.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.http.HttpStatus;
+import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.bootstrap.QuarkusCliRestService;
+import io.quarkus.test.logging.Log;
+import io.quarkus.test.scenarios.annotations.EnabledOnQuarkusVersion;
+import io.quarkus.test.utils.AwaitilityUtils;
+
+@EnabledOnQuarkusVersion(version = "3.8.*", reason = "This class is testing only updates to 3.8.* versions")
```
gtroitsk, line 16: Does this condition work? As I understand, tests will run only if quarkus version in project is 3.8 stream

mocenas: You understand it right. This test class will run only if testing 3.8 stream. Idea is, that only in this case it has the 3.8 RHBQ repo available.

In all my local trying/testing this condition worked.

michalvavrik: Well, I missed that. It doesn't really make sense to add to `main` branch test that only run with 3.8 for which we have dedicated branch. We should never run `main` branch with `3.8` so it doesn't makes sense.

mocenas: I didn't realize that. Should I close this PR and merge it only to 3.8? 
I can put into main only the stuff that is not version specific but those are no actual tests.

michalvavrik: > I didn't realize that. Should I close this PR and merge it only to 3.8?

I think tests specific to 3.8 and lesser as well as code and resources only required by these tests (including test code unless it's planned for reuse) should be only merged to 3.8, yes.

> I can put into main only the stuff that is not version specific but those are no actual tests.

There is 3.9, 3.10, 3.11, 3.12, 3.13 and soon 3.14.  I can see recipes:
- https://github.com/quarkusio/quarkus-updates/blob/main/recipes/src/main/resources/quarkus-updates/core/3.9.alpha1.yaml
- https://github.com/quarkusio/quarkus-updates/blob/main/recipes/src/main/resources/quarkus-updates/core/3.10.alpha1.yaml
- https://github.com/quarkusio/quarkus-updates/blob/main/recipes/src/main/resources/quarkus-updates/core/3.11.yaml
- https://github.com/quarkusio/quarkus-updates/blob/main/recipes/src/main/resources/quarkus-updates/core/3.12.yaml
- https://github.com/quarkusio/quarkus-updates/blob/main/recipes/src/main/resources/quarkus-updates/core/3.13.alpha1.yaml

These migrations will be part of the next LTS, therefore also in RHBQ and customers will use them when migrating between RHBQs. Maybe you could consider testing some of that?

mocenas: > These migrations will be part of the next LTS, therefore also in RHBQ and customers will use them when migrating between RHBQs. Maybe you could consider testing some of that?

 I can do that - is it OK to have this stuff in main?


> I think tests specific to 3.8 and lesser as well as code and resources only required by these tests (including test code unless it's planned for reuse) should be only merged to 3.8, yes.

OK I will rework it into multiple PRs and send 3.8 specific stuff only to 3.8 branch.

michalvavrik: > I can do that - is it OK to have this stuff in main?

IMO yes.

> OK I will rework it into multiple PRs and send 3.8 specific stuff only to 3.8 branch.

Cool, thanks.

quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/update/Quarkus32to38CliUpdateIT.java
michalvavrik: Thanks for long answer, maybe I am slow but I need little more context:

> This is kind of difficult to do, since it presumes that quarkus CLI, which will do the operations, is configured to run with RHBQ. Meaning even if it receives just 3.8.5 as a parameter it will work with RHBQ 3.8 stream, not with upstream quarkus. And if it receives anything else than specified version (in this case 3.8.*) it will not run at all.

I thought based on this comment https://github.com/quarkusio/quarkus/issues/36777#issuecomment-1831709951 that unless you specify `--platform` or `--stream` then neither group id or stream is going to change. What I'd like to do is to check that group id didn't change, hence RHBQ is still used so that we know that we accidentally didn't update customer to community. Which was my motivation for this comment. What is stopping you?

> Test work a bit differently if they receive just 3.8.5 as a parameter or 3.8.5-redhat-. If they receive specific RH build, it will go for that exact BOM, if they receive just version, they will go for a stream. Either way, they should work with RHBQ.

Either way, I'd expect that update app is still Red Hat build of Quarkus regardless of the registry. Do you mean that if config is like

```
---
registries:
- registry.quarkus.io
- registry.quarkus.redhat.com
```

then I'll get community app instead? If so, sure, I'd test that depending on https://github.com/quarkus-qe/quarkus-test-framework/blob/main/quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/model/QuarkusProperties.java#L35 you are able to tell what is expected result. Because if Quarkus version is RHBQ, you are likely to test it with RH registry. Nevertheless maybe locally not...

Just to be clear, this is no biggie, if there is an issue with checking this, I don't require it.

michalvavrik: Great, thanks.

quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/update/Quarkus213to38CliUpdateIT.java
mocenas: In the end, I removed the version specific stuff, it will be another PR into the 3.8 branch. In this PR there is only version-independent stuff. If it is OK I would like to merge this and put tests for 3.8+ updates into separate PR.

quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/update/Quarkus32to38CliUpdateIT.java
mocenas: added

quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/update/Quarkus213to38CliUpdateIT.java
mocenas: ^^ @michalvavrik

michalvavrik: ok

michalvavrik:
@mocenas 8 commits for 2 new files is bit excessive, please squash your commits.

mocenas:
> @mocenas 8 commits for 2 new files is bit excessive, please squash your commits.

I intended to squash them during merge, but OK done.

michalvavrik (approved):
LGTM

mocenas:
Failures are not related, so I'm merging this.

PR #1916, opened 2024-08-07
Feature/podman build

fedinskiy:
### Summary

- Add coverage for podman-build extension in a new module
- Move the module and docker-build to the same folder

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


build/podman/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+    <artifactId>podman-build</artifactId>
+    <packaging>jar</packaging>
+    <name>Quarkus QE TS: Podman-build</name>
+    <properties>
+        <quarkus.container-image.build>true</quarkus.container-image.build>
+    </properties>
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy-jackson</artifactId>
```
michalvavrik, line 19:
```suggestion
            <artifactId>quarkus-rest-jackson</artifactId>
```

build/podman/src/test/java/io/quarkus/ts/docker/PodmanBuildIT.java
```diff
@@ -0,0 +1,40 @@
...
+import io.quarkus.test.utils.DockerUtils;
+
+@QuarkusScenario
+@EnabledIfEnvironmentVariable(named = "DOCKER_HOST", matches = ".*podman.*")
+/**
+ * We run the tests only when podman is enabled and does not pretend to be Docker
+ * since otherwise we won't be able to distinguish between the extension being broken
+ * and a situation, when podman is not installed.
+ *
+ * Unfortunately, that means, that the test is disabled on Windows and Mac
```
michalvavrik, line 23: This feels clumsy, why don't you try `podman` cli command instead, or are there no Podman specific environment variables or something else? We have Windows tests with Podman, so this must be tested there, or not?

build/podman/src/main/resources/application.properties
```diff
@@ -0,0 +1,8 @@
+quarkus.container-image.group=qe
+# DON'T remove the spaces that are at the end of quarkus app name
+quarkus.application.name=hello-world-app      
+quarkus.container-image.tag=1.0.0
+
+# native debug configuration
+quarkus.native.debug.enabled=true
+quarkus.native.additional-build-args=-H:GenerateDebugInfo=1, -H:-SpawnIsolates, -H:DebugInfoSourceCacheRoot=test
```
michalvavrik, line 8: I presume this has a price as more things will be done, are you positive we will need it and we cannot just enable it when there is a problem?

build/podman/src/main/resources/application.properties
```diff
@@ -0,0 +1,8 @@
+quarkus.container-image.group=qe
+# DON'T remove the spaces that are at the end of quarkus app name
+quarkus.application.name=hello-world-app
```
michalvavrik, line 3: I'd prefer to call it differently than in the Docker module if you don't mind. Please add `-podman` postfix or something, especially as the image is deleted in `@AfterAll` not before the tests. If all conditions play right (like `@Tag("podman-incompatible")`, both modules shouldn't be run together, but both modules are still built, right?

BTW what happens when you build `docker-build` module with Podman, did you test it?

build/podman/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-container-image-podman</artifactId>
+        </dependency>
+    </dependencies>
+    <profiles>
+        <!-- Skipped on Windows when availability of Linux containers is not explicitly declared -->
```
michalvavrik, line 27: I am thinking - this `skip-tests-on-windows` will also skip tests when Podman/Docker is available, unless we set property. We could get rid of this if we skipped Quarkus JAR/native build with Quarkus maven plugin with `<quarkus.build.skip>true</quarkus.build.skip>` and let framework to perform the build only when Linux containers are available.

michalvavrik: full disclosure: we use Mave Quarkus plugin when it is a native build to build the executable and we could adjust FW to do that as well, but OK. Closing this thread.

build/podman/src/test/java/io/quarkus/ts/docker/PodmanBuildIT.java
michalvavrik: oki

michalvavrik (approved):

fedinskiy:
@michalvavrik I rebased the changed on the newer main to make CI happy. Do you still want to take another look?

michalvavrik:
> @michalvavrik I rebased the changed on the newer main to make CI happy. Do you still want to take another look?

ah, sorry, ok

PR #1919, opened 2024-08-09
Reproducer for leaking connection in XA mode

fedinskiy:
### Summary
When XA is enabled, connections to MS SQL may leak. This tests checks, that this is not happening.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/sql-db/narayana-transactions/src/test/java/io/quarkus/ts/transactions/MssqlTransactionGeneralUsageIT.java
+++ b/sql-db/narayana-transactions/src/test/java/io/quarkus/ts/transactions/MssqlTransactionGeneralUsageIT.java
@@ -17,7 +17,6 @@ import io.quarkus.test.bootstrap.RestService;
 import io.quarkus.test.bootstrap.SqlServerService;
 import io.quarkus.test.scenarios.QuarkusScenario;
 import io.quarkus.test.scenarios.annotations.DisabledOnFipsAndJava17;
-import io.quarkus.test.scenarios.annotations.DisabledOnQuarkusSnapshot;
 import io.quarkus.test.services.QuarkusApplication;
 import io.quarkus.test.services.SqlServerContainer;
 import io.quarkus.ts.transactions.recovery.TransactionExecutor;
@@ -76,7 +75,7 @@ public class MssqlTransactionGeneralUsageIT extends TransactionCommons {
     }
 
     @Test
-    @DisabledOnQuarkusSnapshot(reason = "The test is too long to run it on CI")
+    @Tag("long-running")
     @Tag("QUARKUS-4185")
     //on average, there is one leaking connection every 2 minutes
     void connectionLeak() throws InterruptedException {
--- a/sql-db/narayana-transactions/src/test/java/io/quarkus/ts/transactions/MssqlTransactionGeneralUsageIT.java
+++ b/sql-db/narayana-transactions/src/test/java/io/quarkus/ts/transactions/MssqlTransactionGeneralUsageIT.java
@@ -3,20 +3,30 @@ package io.quarkus.ts.transactions;
 import static io.quarkus.test.services.containers.DockerContainerManagedResource.DOCKER_INNER_CONTAINER;
 
 import java.io.IOException;
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
 
+import org.apache.http.HttpStatus;
+import org.jboss.logging.Logger;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
 import org.testcontainers.containers.GenericContainer;
 
 import io.quarkus.test.bootstrap.RestService;
 import io.quarkus.test.bootstrap.SqlServerService;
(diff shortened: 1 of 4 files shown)


michalvavrik (approved):
roughly 4,5 minutes gone, that's sad, I'd prefer if we could avoid this test. changes lgtm, thanks

sql-db/narayana-transactions/src/test/java/io/quarkus/ts/transactions/MssqlTransactionGeneralUsageIT.java
```diff
@@ -65,4 +72,22 @@ protected Operation[] getExpectedJdbcOperations() {
                 new Operation(actualOperationName -> actualOperationName.startsWith("UPDATE msdb.")) };
     }
 
+    @Test
+    @Tag("QUARKUS-4185")
+    //on average, there is one leaking connection every 2 minutes
+    void connectionLeak() throws InterruptedException {
+        int before = getConnections();
+        Duration later = Duration.of(4L, ChronoUnit.MINUTES).plus(Duration.ofSeconds(15));
+        Thread.sleep(later.toMillis());
```
michalvavrik, line 81: I wonder if adding logging like "going to sleep" would help when reading logs. It's purely optional.

rsvoboda: +1 to add such a message

fedinskiy: Added the log

fedinskiy:
@michalvavrik @rsvoboda I am in doubts wrt this one. If we enable it, we lose several minutes on every run, otherwise we risk the test too turn stale and for the error to reappear.
Alternative solutions:
1. Disable it for main, but enable in branches
2. Enable it only when there is an environment variable/system property(which one).

What do you think? Maybe you have other suggestions?

michalvavrik:
> @michalvavrik @rsvoboda I am in doubts wrt this one. If we enable it, we lose several minutes on every run, otherwise we risk the test too turn stale and for the error to reappear. Alternative solutions:
> 
> 1. Disable it for main, but enable in branches
> 2. Enable it only when there is an environment variable/system property(which one).
> 
> What do you think? Maybe you have other suggestions?

I feel same way. I think it would be nice to run this test with a product Quarkus version. Problem with branches is that we will forget to put it into next branches and miss possible regression. Enable it explicitly with env var is fine with me as well, though it is something new we would have to add to Jenkins (maybe discuss this with @mjurc as well).

mjurc:
I would keep it in main too personally, this should be ran in weeklies for regression checks.

I don't mind excluding this for CI (especially considering that databases module already takes ages), but I would like to have this in main. As for method of exclusion - Maven exclusion, envvar - I don't really have a strong opinion here

fedinskiy:
@mjurc do we have weeklies, that run on baremetal? Periodic build in Jenkins has only OCP.

Disabled the test on SNAPSHOTS, I suppose it is a good compromise, which doesn't require any drastic changes

sql-db/narayana-transactions/src/test/java/io/quarkus/ts/transactions/MssqlTransactionGeneralUsageIT.java
```diff
@@ -65,4 +75,24 @@ protected Operation[] getExpectedJdbcOperations() {
                 new Operation(actualOperationName -> actualOperationName.startsWith("UPDATE msdb.")) };
     }
 
+    @Test
+    @DisabledOnQuarkusVersion(version = ".*-SNAPSHOT", reason = "The test is too long to run it on CI")
```
michalvavrik, line 79: I kinda expected `DisabledOnQuarkusSnapshot` to do same trick.

michalvavrik (approved):

mjurc:
@fedinskiy Sorry, I meant the nightlies. Does this risk pushing the db modules out of allowed execution time for the runners? I don't think timeliness is that much of concern for the nightly build.

fedinskiy:
@mjurc I do not think four minutes will affect us too much.

I do not want to implement this feature in a hurry, so I created https://issues.redhat.com/browse/QQE-876 to discuss a way to do it properly.

rsvoboda (approved):
Failed test is not related to the PR

rsvoboda:
Talking to Michal and he has good point, that 999-SNAPSHOT is used all the time and waiting till the product testing is risky.

We could for example define a property in https://github.com/quarkus-qe/quarkus-test-suite/blob/main/.github/workflows/ci.yml and skip the test case run using assume from JUnit.

rsvoboda (changes requested):
We need to make the skip relevant just for GH Actions connected with PRs.

michalvavrik (approved):

mjurc (approved):

mjurc:
I'm going to merge this, the failure is unrelated and there's two approval.

Thanks!

PR #1941, opened 2024-08-22
Add file with implicit call to blocking method

fedinskiy:
### Summary

Reproducer for this fix: https://github.com/quarkusio/quarkus/pull/40721

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


http/rest-client-reactive/src/main/java/com/redhat/BrokenBlockingApi.java
```diff
@@ -0,0 +1,26 @@
+package com.redhat; // this bug can not be reproduced in io.io.quarkus package
```
jcarranzan, line 1: why can it not be reproduced in our packages?

jcarranzan (commented):
I left some questions.
Is it finished or is it a draft?
I tried to reproduce the gradle issue opened here with mvn but I didn't get it: 
https://github.com/quarkusio/quarkus/issues/38275#issuecomment-2115117993

http/rest-client-reactive/src/main/java/com/redhat/BrokenBlockingApi.java
michalvavrik: I read upstream review, fix PR and also don't get it. The comment needs to be improved to explain circumstances.

michalvavrik (changes requested):
Please:

- rebase on current main to fix CI
- adding client without any test that calls it and without any link to the upstream issue won't do. I presume you expect build to fail when the client is present and so is bug, but:
  - you need to add comment so that we do understand what is tested
  - you need to add comment so that we do understand why is it not called
  - you need to link upstream issue

fedinskiy:
@jcarranzan it is finished. The bug can be reproduced with maven, but only for quarkus 3.10 or older.
@michalvavrik added the required comments and rebased the PR. I did not add any test calls due to YAGNI principle, but I can add them, if you insist.

michalvavrik:
> I did not add any test calls due to YAGNI principle, but I can add them, if you insist.

I don't insist, thanks for the changes. I don't quite agree with YAGNI in here but I have already explained my motivation for using the client, so let's keep your code as is.

michalvavrik (approved):

jcarranzan:
Ok  Fedor, could you put the maven command to know how to reproduce the bug? It would be useful to me, and then let's merge it, thank you.

jcarranzan (approved):

PR #1946, opened 2024-08-25
Verify log file rotation for periods longer than day

gtroitsk:
### Summary

Verifies issue https://github.com/quarkusio/quarkus/issues/40016

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/logging/jboss/src/main/resources/application.properties
+++ b/logging/jboss/src/main/resources/application.properties
@@ -8,3 +8,9 @@
 # By default min-level is set to DEBUG
 #quarkus.log.min-level=DEBUG
 quarkus.log.level=TRACE
+
+quarkus.log.file.enable=true
+quarkus.log.file.rotation.file-suffix=.yyyy-MM
+# Max file size value is set to this value to ensure that both backup logs are created
+quarkus.log.file.rotation.max-file-size=100
+quarkus.log.file.rotation.max-backup-index=2
--- a/logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/DefaultMinLogLevelIT.java
+++ b/logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/DefaultMinLogLevelIT.java
@@ -1,13 +1,32 @@
 package io.quarkus.ts.logging.jboss;
 
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import org.awaitility.Awaitility;
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
 import io.quarkus.test.bootstrap.RestService;
 import io.quarkus.test.scenarios.QuarkusScenario;
 import io.quarkus.test.services.QuarkusApplication;
+import io.quarkus.test.utils.FileUtils;
 
 @QuarkusScenario
 public class DefaultMinLogLevelIT {
+
+    private static final String LOG_FILE_NAME = "quarkus.log";
+    private static final DateTimeFormatter DATE_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM");
+    private static final String EXPECTED_LOG_MESSAGE = "Example log message";
 
     @QuarkusApplication
     static RestService app = new RestService();
@@ -24,5 +43,36 @@ public class DefaultMinLogLevelIT {
 
         // the value of minimum logging level overrides the logging level
         app.logs().assertDoesNotContain("Trace log example");
+    }
+
+    @Test
+    @Tag("https://github.com/quarkusio/quarkus/issues/40016")
+    public void checkLogRotationContent() throws IOException {
+        Path logDir = app.getServiceFolder();
+        // one main log file plus others rotated
+        int logFilesNumber = Integer.parseInt(app.getProperty("quarkus.log.file.rotation.max-backup-index").get()) + 1;
+
+        app.given().when().get("/log").then().statusCode(204);
+
+        Awaitility.await().atMost(10, TimeUnit.SECONDS).until(() -> {
+            File[] rotatedFiles = logDir.toFile().listFiles((dir, name) -> name.startsWith(LOG_FILE_NAME));
+            return rotatedFiles != null && rotatedFiles.length == logFilesNumber;
+        });
+
+        // Verify that the rotated file with the expected suffix exists
+        String expectedSuffix = "." + DATE_FORMATTER.format(LocalDate.now()) + "\\.\\d+$";
+        List<Path> rotatedFiles = Files.list(logDir)
+                .filter(path -> path.getFileName().toString().matches(LOG_FILE_NAME + expectedSuffix)).toList();
+
+        assertFalse(rotatedFiles.isEmpty(), "Rotated log files with expected suffix were not found.");
+
+        // Check if rotated files contain the log message
+        boolean allFilesContainMessage = rotatedFiles.stream()
+                .allMatch(rotatedFile -> {
+                    String fileContent = FileUtils.loadFile(rotatedFile.toFile());
+                    return fileContent.contains(EXPECTED_LOG_MESSAGE);
+                });
+
+        assertTrue(allFilesContainMessage, "Rotated log files do not contain the expected message " + EXPECTED_LOG_MESSAGE);
     }
 }
--- a/logging/jboss/src/main/java/io/quarkus/ts/logging/jboss/LogResource.java
+++ b/logging/jboss/src/main/java/io/quarkus/ts/logging/jboss/LogResource.java
@@ -51,6 +51,10 @@ public class LogResource {
         LOG.info("Info log example");
         LOG.debug("Debug log example");
         LOG.trace("Trace log example");
+
+        for (int i = 0; i < 10; i++) {
+            LOG.info("Example log message: " + i);
+        }
     }
 
     private void addLogMessage(Logger logger, String level, String message) {


jedla97:
@gtroitsk It's look good to me! Just would be better also checking if the file is created with that suffix?

michalvavrik:
> @gtroitsk It's look good to me! Just would be better also checking if the file is created with that suffix?

Let's go further, let's check content as it adds better coverage while this is very much like unit test. IMHO we don't want to test isolated bugs if we can have something similar to what users do. And if they enable logging to file, they are likely to desire there is a content.

michalvavrik (dismissed):
In addition to my previous comment, also please rebase on current main to fix CI.

logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/DefaultMinLogLevelIT.java
```diff
@@ -9,6 +23,11 @@
 @QuarkusScenario
 public class DefaultMinLogLevelIT {
 
+    private static final String LOG_FILE_NAME = "quarkus.log";
+    private static final DateTimeFormatter DATE_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM");
+    private static final String EXPECTED_LOG_MESSAGE = "Example log message";
+    private final String LOG_DIR = "target/" + this.getClass().getSimpleName() + "/app";
```
michalvavrik, line 29: this won't do, please use app service folder on `app` static property

logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/DefaultMinLogLevelIT.java
```diff
@@ -25,4 +44,37 @@ public void checkDefaultLogMinLevel() {
...
+            File logDir = new File(LOG_DIR);
+            File[] rotatedFiles = logDir.listFiles((dir, name) -> name.startsWith(LOG_FILE_NAME));
+            return rotatedFiles != null && rotatedFiles.length > 1;
+        });
+
+        // Verify that the rotated file with the expected suffix exists
+        String expectedSuffix = "." + DATE_FORMATTER.format(LocalDate.now()) + "\\.\\d+$";
+        Path logDirPath = Path.of(LOG_DIR);
+        List<Path> rotatedFiles = Files.list(logDirPath)
+                .filter(path -> path.getFileName().toString().matches(LOG_FILE_NAME + expectedSuffix)).toList();
```
michalvavrik, line 63: I don't understand where is difference between your code and https://github.com/quarkus-qe/quarkus-test-framework/blob/main/quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/LocalhostQuarkusApplicationManagedResource.java#L47 that you can access with `new RestService().logs().assertContains()`. Can't you avoid all of this code or maybe leverage our framework somehow?

gtroitsk: What you address is our custom logging from FW. I test quarkus logging rotation to the file, that is by default named "quarkus.log"

logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/DefaultMinLogLevelIT.java
gtroitsk: I can change it, np

logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/DefaultMinLogLevelIT.java
michalvavrik: Isn't it in the same folder?

michalvavrik: I mean, if it is the same folder, same format, just differs in the file name, you can easily do what our FW do and you have log asserter. So `io.quarkus.test.utils.LogsVerifier` doesn't allow you to specify concrete filename, which makes it impossible. Ok.

gtroitsk: I made some simplification in code but the principle remains the same

logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/DefaultMinLogLevelIT.java
gtroitsk: changed

logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/DefaultMinLogLevelIT.java
```diff
@@ -25,4 +43,37 @@ public void checkDefaultLogMinLevel() {
...
+        // Check if any rotated file contains the log message
+        boolean messageFound = rotatedFiles.stream()
+                .flatMap(rotatedFile -> {
+                    try {
+                        return Files.lines(rotatedFile);
+                    } catch (IOException e) {
+                        throw new UncheckedIOException(e);
+                    }
+                })
+                .anyMatch(line -> line.contains(EXPECTED_LOG_MESSAGE));
```
michalvavrik, line 75: this succeeds if at least one of the files has expected line, I don't know how many files you expect. We have `io.quarkus.test.utils.FileUtils`, have you considered using them to cut the verbosity (try catches etc.)? I'd expect using `io.quarkus.test.utils.FileUtils#findTargetFile(java.nio.file.Path, java.lang.String)` and `io.quarkus.test.utils.FileUtils#loadFile(java.lang.String)` would cut your code to three lines.

michalvavrik:
@gtroitsk it looks good, I still think you code refactor it bit, but please consider it rather unimportant.

logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/DefaultMinLogLevelIT.java
gtroitsk: I expect 3 files according to configuration. Check if each of rotated files contain the log message now use `FileUtils.loadFile`

michalvavrik: okay, 3? in that case previous check was incorrect because you didn't check that **each** file contains that line. You did test that **any** did. I'll comment elsewhere on this expectation.

logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/DefaultMinLogLevelIT.java
```diff
@@ -25,4 +43,33 @@ public void checkDefaultLogMinLevel() {
...
+    // Verifies https://github.com/quarkusio/quarkus/issues/40016
+    @Test
+    public void checkLogRotationContent() throws IOException {
+        Path logDir = app.getServiceFolder();
+
+        app.given().when().get("/log").then().statusCode(204);
+
+        Awaitility.await().atMost(10, TimeUnit.SECONDS).until(() -> {
+            File[] rotatedFiles = logDir.toFile().listFiles((dir, name) -> name.startsWith(LOG_FILE_NAME));
+            return rotatedFiles != null && rotatedFiles.length > 1;
```
michalvavrik, line 56: yeah, so, you expect 3 files but are content with finding one? I understood your comment that you want to find this line in 3 files, so I expect you need 3 files. Won't the test be flaky if you wait just for one and hope others are stored in the meanwhile?

logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/DefaultMinLogLevelIT.java
```diff
@@ -25,4 +43,33 @@ public void checkDefaultLogMinLevel() {
...
+        });
+
+        // Verify that the rotated file with the expected suffix exists
+        String expectedSuffix = "." + DATE_FORMATTER.format(LocalDate.now()) + "\\.\\d+$";
+        List<Path> rotatedFiles = Files.list(logDir)
+                .filter(path -> path.getFileName().toString().matches(LOG_FILE_NAME + expectedSuffix)).toList();
+
+        assertFalse(rotatedFiles.isEmpty(), "Rotated log file with expected suffix not found.");
```
michalvavrik, line 64: this looks like you are looking for single file...

logging/jboss/src/main/resources/application.properties
```diff
@@ -8,3 +8,8 @@
 # By default min-level is set to DEBUG
 #quarkus.log.min-level=DEBUG
 quarkus.log.level=TRACE
+
+quarkus.log.file.enable=true
+quarkus.log.file.rotation.file-suffix=.yyyy-MM
+quarkus.log.file.rotation.max-file-size=100
```
jedla97, line 14: Can you add some comment that this size is for test populate both backup logs.

As someone can try to set bigger value or when adding some tests, the `checkLogRotationContent` can fail

I play with the value and hit the failing `checkLogRotationContent`

jedla97 (changes requested):
I just have two minor comments in this last update.

logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/DefaultMinLogLevelIT.java
jedla97: Tbh I don't think it was wrong as with backup logs you can't know how many will be there. Someone can update `quarkus.log.file.rotation.max-file-size` and this test can have only one backup log.

logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/DefaultMinLogLevelIT.java
```diff
@@ -25,4 +43,33 @@ public void checkDefaultLogMinLevel() {
         // the value of minimum logging level overrides the logging level
         app.logs().assertDoesNotContain("Trace log example");
     }
+
+    // Verifies https://github.com/quarkusio/quarkus/issues/40016
```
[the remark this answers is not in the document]
gtroitsk: updated

logging/jboss/src/main/resources/application.properties
gtroitsk: updated

logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/DefaultMinLogLevelIT.java
michalvavrik: sorry, that is not acceptable; someone cannot just update `max-file-size` because this is part of the test; if he updates it, the test must fail

michalvavrik: @jedla97 configuration properties are either required for test in their current value to test a feature, or we should drop them

jedla97 (approved):
Thanks for changes

logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/DefaultMinLogLevelIT.java
michalvavrik: Sure, I have no idea whether this test is stable. I commented on changing property for no reason. by looking at this test I think it is weird I don't understand why it is separated to 3 files and max-file-size refers to memory size, not lines. I somehow expected @gtroitsk have it worked out :-)

logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/DefaultMinLogLevelIT.java
```diff
@@ -25,4 +44,35 @@ public void checkDefaultLogMinLevel() {
...
+            File[] rotatedFiles = logDir.toFile().listFiles((dir, name) -> name.startsWith(LOG_FILE_NAME));
+            return rotatedFiles != null && rotatedFiles.length == logFilesNumber;
+        });
+
+        // Verify that the rotated file with the expected suffix exists
+        String expectedSuffix = "." + DATE_FORMATTER.format(LocalDate.now()) + "\\.\\d+$";
+        List<Path> rotatedFiles = Files.list(logDir)
+                .filter(path -> path.getFileName().toString().matches(LOG_FILE_NAME + expectedSuffix)).toList();
+
+        assertFalse(rotatedFiles.isEmpty(), "Rotated log files with expected suffix were not found.");
```
michalvavrik, line 67: For record @jedla97 , I still think this condition doesn't make sense. First there is waiting for `max-backup-index` + 1=3 and now it's satisfied with 1

PR #1983, opened 2024-09-05
Fix GitHub changed-files issue occurring when fork branch is not rebased on the PR target branch resulting in inability to detect changes

michalvavrik:
### Summary

Sometimes we see that GitHub action `tj-actions/changed-files` fails with `Merge base is not in the local history, fetching remote target branch again..` and `Unable to find merge base between ...` etc.

We have seen this message (it is present twice in that script, but we hit the one I am pointing to): https://github.com/tj-actions/changed-files/blob/1d9fdda44c9da3604e9be858b3edb3e0501302af/src/commitSha.ts#L598

My thinking is that we compare it with base branch HEAD SHA from GH PR event. However:

- between PR event and time when `changed-files` action and `checkout` are run, there is some time so there can be changes on the target branch, therefore ideally we want to have event base HEAD SHA to be matching to the actual base HEAD sha that is fetched; action taken
  - moving checkout and change file detection to be the first thing that is done
  - refactoring - moving `Prepare Maven Params For Linux JVM and native Build` to the second place as they are alike
- I am making sure that base HEAD SHA is always present myself, using `git` instead of raising `changed-files` `depth`. This way, we know that when changed-files look for the base HEAD SHA locally, it will always be present.

There is one, hmm, discrepancy. Based on documentation here https://github.com/tj-actions/changed-files/blob/1d9fdda44c9da3604e9be858b3edb3e0501302af/README.md?plain=1#L654 I would say we should use `with: since_last_remote_commit: true`. However, the way I read their script, it doesn't work as it is described. We want to have diff between target branch HEAD SHA and fork branch HEAD SHA, however with that option it would result in `target = current` https://github.com/tj-actions/changed-files/blob/1d9fdda44c9da3604e9be858b3edb3e0501302af/src/commitSha.ts#L347. So I'd way say we don't want that option.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,72 @@ name: "Pull Request CI"
 on:
   - pull_request
 jobs:
+  detect-test-suite-modules:
+    name: Detect Modules in PR
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+      - id: fetch-latest-target-branch-commits # makes sure that latest target branch HEAD commit is available locally
+        # fetching the latest 2 commits, one should be enough but 2 could avoid race (diff between GH PR event base sha and actual remote base sha)
+        name: 'Fetch the latest 2 target (base) branch commits'
+        run: |
+          git remote add quarkus_qe_target_repo https://github.com/quarkus-qe/quarkus-test-suite.git
+          git fetch quarkus_qe_target_repo ${GITHUB_BASE_REF} --depth=2
+      - id: files
+        uses: tj-actions/changed-files@v45
+        continue-on-error: true
+      - id: detect-changes
+        run: |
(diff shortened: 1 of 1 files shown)


michalvavrik:
It just doesn't seem to trigger other linux runs, I have mentioned before but I thought there were just not enough runners. I'll look what is wrong.

jedla97 (approved):
Thanks for looking at this and fixing it! Changes make sense to me.

michalvavrik:
> Thanks for looking at this and fixing it! Changes make sense to me.

Yes, I agree, this fixes the issue (probably), thank you. But let's wait @jedla97 till I find why are not Linux tests not triggering. I just don't see it, YAML seems fine :-/. I'll check and let you know.

michalvavrik:
Ah, found it `[Pull Request CI](https://github.com/quarkus-qe/quarkus-test-suite/actions/runs/10729720585/workflow)
Error when evaluating 'strategy' for job 'linux-build-native-latest'. .github/workflows/ci.yml (Line: 186, Col: 26): Error parsing fromJson,.github`

jedla97 (approved):

PR #1996, opened 2024-09-10
[3.15] Turn openshift-arm profile into general aarch64

mjurc:
### Summary

* To avoid needing to define the same multiarch container versions for bare metal and OpenShift use cases, this commit turns openshift-arm into general aarch64 profile.
* Updating the readme file with examples on how to run this profile with both OpenShift and bare metal uses.

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:


michalvavrik (approved):
I don't know whether will `aarch64` profile win over `openshift` profile if one day there is conflict between merged system property variables, but ATM it seems they share set of properties and `aarch64` has some extra, which is Ok. Seems like a good idea.

PR #2003, opened 2024-09-12
Add coverage for quarkus-picocli

jcarranzan:
### Summary

JIRA: https://issues.redhat.com/browse/QUARKUS-4221
Currently, we are going to support Quarkus Picocli, so that's the reason for this new test coverage.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/quarkus-picocli/pom.xml
+++ b/quarkus-picocli/pom.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkus.ts.qe</groupId>
+        <artifactId>parent</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>quarkus-picocli</artifactId>
+    <name>Quarkus QE TS: Quarkus Picocli</name>
+    <packaging>jar</packaging>
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-picocli</artifactId>
+        </dependency>
+    </dependencies>
+</project>
--- a/quarkus-picocli/src/main/java/io/quarkus/ts/qe/command/AgeCommand.java
+++ b/quarkus-picocli/src/main/java/io/quarkus/ts/qe/command/AgeCommand.java
@@ -1,0 +1,25 @@
+package io.quarkus.ts.qe.command;
+
+import jakarta.enterprise.context.ApplicationScoped;
+
+import io.quarkus.logging.Log;
+
+import picocli.CommandLine;
+
+/**
+ * According picocli guide,
+ * Beans with @CommandLine.Command should not use proxied scopes (e.g. do not use @ApplicationScope)
+ * because Picocli will not be able to set field values in such beans.
+ * So this class creation is to test it, see testAgeCommandWithApplicationScoped.
+ */
+@CommandLine.Command(name = "age", mixinStandardHelpOptions = true)
+@ApplicationScoped
+public class AgeCommand implements Runnable {
+    @CommandLine.Option(names = { "-a", "--age" }, description = "your age", defaultValue = "0")
+    private int age;
+
+    @Override
+    public void run() {
+        Log.info("Your age is: " + age);
+    }
+}
--- a/quarkus-picocli/src/main/java/io/quarkus/ts/qe/configuration/Config.java
+++ b/quarkus-picocli/src/main/java/io/quarkus/ts/qe/configuration/Config.java
@@ -1,0 +1,27 @@
+package io.quarkus.ts.qe.configuration;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.inject.Produces;
+
+import io.quarkus.arc.profile.IfBuildProfile;
+import io.quarkus.picocli.runtime.annotations.TopCommand;
+import io.quarkus.ts.qe.command.EntryCommand;
+import io.quarkus.ts.qe.command.OtherEntryCommand;
+
+@ApplicationScoped
+public class Config {
+    @Produces
+    @TopCommand
+    @IfBuildProfile("dev")
+    public Object devCommand() {
+        return EntryCommand.class;
+    }
+
+    @Produces
+    @TopCommand
+    @IfBuildProfile("prod")
+    public Object prodCommand() {
+        return OtherEntryCommand.class;
+    }
+
+}
--- a/quarkus-picocli/src/test/java/PicocliDevIT.java
+++ b/quarkus-picocli/src/test/java/PicocliDevIT.java
@@ -1,0 +1,102 @@
+import static java.util.concurrent.CompletableFuture.runAsync;
+
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.QuarkusApplication;
+import io.quarkus.ts.qe.command.AgeCommand;
+import io.quarkus.ts.qe.command.EntryCommand;
+import io.quarkus.ts.qe.command.HelloCommand;
+import io.quarkus.ts.qe.configuration.Config;
+import io.quarkus.ts.qe.services.AgeService;
+import io.quarkus.ts.qe.services.HelloService;
+
+@QuarkusScenario
+public class PicocliDevIT {
+
+    @QuarkusApplication(classes = { EntryCommand.class, Config.class, HelloCommand.class, AgeCommand.class, AgeService.class,
+            HelloService.class }, properties = "dev.properties")
+    static final RestService greetingApp = new RestService()
+            .withProperty("quarkus.args", "greeting --name QE")
+            .setAutoStart(false);
+
+    @QuarkusApplication(classes = { EntryCommand.class, Config.class, HelloCommand.class, AgeCommand.class, AgeService.class,
+            HelloService.class }, properties = "dev.properties")
+    static final RestService ageApp = new RestService()
+            .withProperty("quarkus.args", "age --age 30")
+            .setAutoStart(false);
+
+    @QuarkusApplication(classes = { EntryCommand.class, Config.class, HelloCommand.class, AgeCommand.class, AgeService.class,
+            HelloService.class }, properties = "dev.properties")
+    static final RestService greetingBlankArgumentApp = new RestService()
+            .withProperty("quarkus.args", " --name QE")
+            .setAutoStart(false);
+
+    @QuarkusApplication(classes = { EntryCommand.class, Config.class, HelloCommand.class, AgeCommand.class, AgeService.class,
+            HelloService.class }, properties = "dev.properties")
+    static final RestService greetingInvalidArgumentApp = new RestService()
+            .withProperty("quarkus.args", "greeting -x QE")
+            .setAutoStart(false);
+
+    @QuarkusApplication(classes = { EntryCommand.class, Config.class, HelloCommand.class, AgeCommand.class, AgeService.class,
+            HelloService.class }, properties = "dev.properties")
+    static final RestService bothTopCommandApp = new RestService()
+            .setAutoStart(false);
+
+    @Test
+    public void verifyErrorForApplicationScopedBeanInPicocliCommand() {
+        try {
+            runAsync(ageApp::start);
+            ageApp.logs().assertContains("CDI: programmatic lookup problem detected");
+        } finally {
+            ageApp.stop();
+        }
+    }
+
+    @Test
+    public void verifyGreetingCommandOutputsExpectedMessage() {
+        try {
+            runAsync(greetingApp::start);
+            greetingApp.logs().assertContains("Hello QE!");
+        } finally {
+            greetingApp.stop();
+        }
+    }
+
+    @Test
+    void verifyErrorForBlankArgumentsInGreetingCommand() {
+        try {
+            runAsync(greetingBlankArgumentApp::start);
+            greetingBlankArgumentApp.logs().assertContains("Unmatched arguments from index 0: '', '--name', 'QE'");
+        } finally {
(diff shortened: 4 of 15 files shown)


rsvoboda:
@fedinskiy please take a look

quarkus-picocli/src/main/java/io/quarkus/ts/qe/configuration/Config.java
```diff
@@ -0,0 +1,27 @@
...
+import io.quarkus.picocli.runtime.annotations.TopCommand;
+import io.quarkus.ts.qe.command.EntryCommand;
+import io.quarkus.ts.qe.command.OtherEntryCommand;
+
+@ApplicationScoped
+public class Config {
+    @Produces
+    @TopCommand
+    @IfBuildProfile("dev")
+    public Object devCommand() {
```
fedinskiy, line 16: `Object`?

quarkus-picocli/src/test/java/PicocliIT.java
```diff
@@ -0,0 +1,118 @@
...
+        }
+    }
+
+    @Test
+    public void verifyGreetingCommandOutputsExpectedMessage() throws InterruptedException {
+        String expectedOutput = String.format("Hello %s!", GREETING_NAME_VALUE);
+        try {
+            runAsync(greetingApp::start);
+            greetingApp.logs().assertContains(expectedOutput);
+            Thread.sleep(3500);
```
fedinskiy, line 70: Why wait?

quarkus-picocli/src/test/java/PicocliIT.java
```diff
@@ -0,0 +1,118 @@
...
+        } finally {
+            greetingApp.stop();
+        }
+
+    }
+
+    @Test
+    void verifyErrorForBlankArgumentsInGreetingCommand() {
+        String expectedError = "Unmatched arguments from index 0: '', '--name', 'QE'";
```
fedinskiy, line 79: This variable is used in a single place only, Can it be inlined?

quarkus-picocli/src/test/java/ProdiIT.java
```diff
@@ -0,0 +1,29 @@
+import static java.util.concurrent.CompletableFuture.runAsync;
+
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.QuarkusApplication;
+
+@Disabled ("No possible to test 2 profiles with the same build")
```
fedinskiy, line 10: Why do we adding disabled test? Is there an existing issue?

quarkus-picocli/src/test/java/PicocliIT.java
```diff
@@ -0,0 +1,118 @@
...
+    static final String GREETING_NAME_VALUE = "QE";
+
+    static final String AGE_COMMAND_DESC = "age";
+    static final String AGE_OPTION = "--age";
+    static final String AGE_VALUE = "30";
+
+    @QuarkusApplication
+    static final RestService greetingApp = new RestService()
+            .withProperty("quarkus.args",
+                    GREETING_COMMAND_DESC + " " + GREETING_NAME_OPTION + " " + GREETING_NAME_VALUE)
```
fedinskiy, line 23: Here and everywhere in the file, using constants make the code both harder to read and longer. Maybe they should not be used?

quarkus-picocli/src/test/java/OpenShiftPicocliIT.java
```diff
@@ -0,0 +1,5 @@
+import io.quarkus.test.scenarios.OpenShiftScenario;
+
+@OpenShiftScenario
+public class OpenShiftPicocliIT extends PicocliIT {
```
fedinskiy, line 4: As far as I understand, we do not plan to run these tests on openshift . Do not forget to remove the file

fedinskiy (changes requested):

quarkus-picocli/src/main/java/io/quarkus/ts/qe/configuration/Config.java
jcarranzan: As in the guide example: https://quarkus.io/guides/picocli#different-entry-command-for-each-profile

quarkus-picocli/src/test/java/PicocliIT.java
jcarranzan: It was for local testing purposes, I forgot clean it, 'll remove it, thanks

quarkus-picocli/src/test/java/PicocliIT.java
jcarranzan: ok, I see your point here, I will change it for the commands Strings directly.

quarkus-picocli/src/test/java/ProdiIT.java
jcarranzan: I did an investigation and is not possible to have both tests classes enable because one of them will fail depending on the profile activated.
Because Quarkus uses build-time profiles to include or exclude beans and configurations during the build process, so with annotations with `@IfBuildProfile("prod")` are only included in the build if the 'prod' or 'dev' profile is active at build time. 
If we remove the application.porperties with quarkus.profile config, you can test through mvn commands` -Dquarkus.profile=dev` or pro, and the tests will pass.

michalvavrik: The `quarkus.profile` is runtime property, our FW is not aware that you needed new custom build due to the runtime property. IMO either improve FW or workaround it by setting some build-time property that will differ between test classes.

michalvavrik (dismissed):
There is absolutely no test class, but you only add new classes, new module. How this PR tests anything?

jcarranzan:
> There is absolutely no test class, but you only add new classes, new module. How this PR tests anything?

My src/folder wasn't upload, sorry about that, it's done now :)

fedinskiy:
@jcarranzan as Michal've already noticed, it looks like you've deleted all tests in the last update. Please recheck

quarkus-picocli/src/test/java/PicocliProdIT.java
```diff
@@ -0,0 +1,30 @@
+import static java.util.concurrent.CompletableFuture.runAsync;
+
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.QuarkusApplication;
+
+@Disabled("Requires 'prod' profile. Disabled because the test fails when the application is built and run with the default 'dev' profile. Enable by building with '-Dquarkus.profile=prod'.")
```
fedinskiy, line 10: Should we run the test suite manually? This defeats the whole idea of automated tests.

quarkus-picocli/src/test/java/PicocliProdIT.java
```diff
@@ -0,0 +1,30 @@
...
+            .withProperty("quarkus.profile", "prod")
+            .withProperty("quarkus.args", "start -t 60 -v")
+            .setAutoStart(false);
+
+    @Test
+    public void verifyCustomizedCommandLineBehavior() throws InterruptedException {
+        String expectedOutput = "Service started with timeout: 60 and verbosity";
+        try {
+            runAsync(customized::start);
+            Thread.sleep(3200);
```
fedinskiy, line 24: And I think we agreed to not use sleep in these tests?

quarkus-picocli/src/test/java/PicocliDevIT.java
```diff
@@ -0,0 +1,96 @@
...
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.QuarkusApplication;
+
+@QuarkusScenario
+public class PicocliDevIT {
+
+    @QuarkusApplication
+    static final RestService greetingApp = new RestService()
+            .withProperty("quarkus.profile", "dev")
```
fedinskiy, line 14: We have this property in both application.properties and here. Is it really needed?

fedinskiy (changes requested):

quarkus-picocli/src/test/java/PicocliProdIT.java
jcarranzan: I understand that disabling the test is not ideal, and I believe the reason is clearly explained in the code comments. I've tried several configuration properties, as I mentioned in my earlier comments—including those suggested by Michal—but unfortunately, I couldn't get them to work.

I'm open to any solutions you might have, or I can remove the test if you prefer. What are your suggestions?

quarkus-picocli/src/test/java/PicocliDevIT.java
jcarranzan: Yes, application.properties active the dev profile with the dev beans needed in the app build. If not beans specific to `dev` profile are not included and commands are unavailable then the tests fail.

quarkus-picocli/src/test/java/PicocliProdIT.java
michalvavrik: 1. only set `quarkus.profile` in `withProperty`
2. put `withProperty("quarkus.arc.remove-unused-beans", "none")` (which is build time property irrelevant for your test)
3. enable the test

and report back if it fixed your issues

quarkus-picocli/src/test/java/PicocliDevIT.java
michalvavrik: Proper explanation would be that framework doesn't know that you need the runtime property during the build time, so it only passes it as system property to a Java process.

quarkus-picocli/src/test/java/PicocliProdIT.java
michalvavrik: I kinda wonder if this can still work considering our FW don't know it is a build time property. There is another way to do this: instead of using `application.properties` you can specify the property using a test class profile. mmnt, I'll check if that can work.

michalvavrik: yeah, so here is the way:

- use `io.quarkus.test.services.QuarkusApplication#properties` instead of `application.properties`

Try just that for both test classes. If this alone isn't enough, you need to force custom build. The best way to do this is to adjust FW, but you can workaround this by specifying app classes, or forced dependency or the build time config property as I suggested before. All of that must work.

jcarranzan: I tried that approach on Wednesday but I forgot put a class in the @QuarkusApplication. 
Currently with my changes both tests works, thanks @michalvavrik

quarkus-picocli/src/test/java/PicocliDevIT.java
fedinskiy: We now have `properties = "dev.properties"` option. Do we still need to add properties explicitly?

jcarranzan: No, now we didn't need it, thanks @fedinskiy .

quarkus-picocli/src/test/java/PicocliDevIT.java
```diff
@@ -0,0 +1,102 @@
...
+
+    @QuarkusApplication(classes = { EntryCommand.class, Config.class, HelloCommand.class, AgeCommand.class, AgeService.class,
+            HelloService.class }, properties = "dev.properties")
+    static final RestService bothTopCommandApp = new RestService()
+            .setAutoStart(false);
+
+    @Test
+    public void verifyErrorForApplicationScopedBeanInPicocliCommand() {
+        try {
+            runAsync(ageApp::start);
```
fedinskiy, line 50: Test is called `verify error`, and test plans say, that there should be an error for application scoped commands, yet there is no error.

jcarranzan: good catch, I've just fixed.

fedinskiy (approved):

fedinskiy:
Thank you! Can be merged if CI is green

quarkus-picocli/pom.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkus.ts.qe</groupId>
+        <artifactId>parent</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>quarkus-picocli</artifactId>
```
michalvavrik, line 9: There is no test module name. You need name as all the other modules have. Like follow `Quarkus QE TS: Picocli`.

quarkus-picocli/src/main/java/io/quarkus/ts/qe/command/AgeCommand.java
```diff
@@ -0,0 +1,25 @@
...
+
+import io.quarkus.logging.Log;
+
+import picocli.CommandLine;
+
+/**
+ * According picocli guide,
+ * Beans with @CommandLine.Command should not use proxied scopes (e.g. do not use @ApplicationScope)
+ * because Picocli will not be able to set field values in such beans.
+ * So this class creation is to test it, see testAgeCommandWithApplicationScoped.
```
michalvavrik, line 13: Just an opinion for future test development, no action required:

Quarkus docs basically says "if you use proxied scope, define setter". And your test is "let's test it doesn't work without setter". IMO you should test that setter helps (if you want, in addition to what doesn't work). Because setter you don't test and I am not quite interested that something doesn't work if it is documented. Users shouldn't use it even if it worked considering it requires reflection and it's not considered unsupported.

michalvavrik: Maybe what you want to test is that `Unmatched arguments from index 3` is printed out when you try to set it? In that case it's ok, though I still think it's less important than testing what docs advise to do.

quarkus-picocli/pom.xml
jcarranzan: done

michalvavrik:
> Thank you! Can be merged if CI is green

Sadly I went through last GH CI run (before PR updated 'name' in pom) and it seems that this CI doesn't test this PR. I think these is some workflow bug, but won't look into now. I'll merge this when validation succeeds and we can see whether daily build pass or not.

PR #2038, opened 2024-09-24
Disable on Native

jcarranzan:
### Summary

As I described in the TP (https://github.com/quarkus-qe/quarkus-test-plans/blob/9c6b5a01882dd0a96fa0fad9019eeb479760e22e/QUARKUS-4221.md?plain=1#L11) , this is supported only for JVM. 

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/quarkus-picocli/src/test/java/PicocliDevIT.java
+++ b/quarkus-picocli/src/test/java/PicocliDevIT.java
@@ -4,6 +4,7 @@ import org.junit.jupiter.api.Test;
 
 import io.quarkus.test.bootstrap.RestService;
 import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.scenarios.annotations.DisabledOnNative;
 import io.quarkus.test.services.QuarkusApplication;
 import io.quarkus.ts.qe.command.AgeCommand;
 import io.quarkus.ts.qe.command.EntryCommand;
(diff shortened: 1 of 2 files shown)


michalvavrik (approved):
LGTM, I suspect imports are not sorted, but that's the CI problem. Thanks for the fix.

michalvavrik (approved):

rsvoboda:
I understand the disable in dev mode, but why for prod mode?

Keycloak team is the trigger, but we aim to support this for any customer. And having cli utilities compiled into native can be handy for them - one binary / no need for java runtime / faster boot

michalvavrik:
> that already mention some incompatibilities and workaround https://quarkus.io/guides/picocli#native-mode-support

just to be clear, I don't believe there are any incompatibilities. what the guide says is that "if we missed something with future bumps, here is trick that fixes native mode". but it will be a bug anyway, so if there is something not working, file the issue.

michalvavrik:
anyway, I suggest to merge this and give @jcarranzan time to investigate. there is no point in having failing daily builds.

michalvavrik:
because it follows current TP, so let's change TP first...

michalvavrik:
let's continue here https://github.com/quarkus-qe/quarkus-test-suite/pull/2043#pullrequestreview-2325551851

PR #1986, opened 2024-09-06
Test Mutual TLS with Management interface configured using TLS Registry extension and certificate reloading

michalvavrik:
### Summary

- tests TLS registry with management interface
- tests certificate reloading

Please select the relevant options.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport
- [x] New scenario (non-breaking change which adds functionality)
- [ ] This change requires a documentation update
- [ ] This change requires execution against OCP (use `run tests` phrase in comment)

### Checklist:
- [x] Methods and classes used in PR scenarios are meaningful
- [x] Commits are well encapsulated and follow [the best practices](https://cbea.ms/git-commit/)

PR #2060, opened 2024-10-02
Test Kafka TLS registry support and manual reloading of TLS registry configuration

michalvavrik:
### Summary

- adds Kafka TLS registry test coverage as https://github.com/quarkusio/quarkus/issues/43107 is fixed now; changes:
  - SASL_SSL client and producer were created and initialized (tried to consume topic) when only SSL was tested and vice versa, so I made sure we don't initialize it when we don't test it (as it would fail for TLS registry [missing named config])
  - newly we are taking SSL/SASL_SSL config from default broker as documented here https://quarkus.io/guides/kafka#kafka-bare-clients because it already contains what we set plus for TLS config scenario it also contains what Quarkus Kafka extension set (`ssl.engine.factory.class=io.quarkus.kafka.client.tls.QuarkusKafkaSslEngineFactory` among other non-TLS props) which is required otherwise SSL context is not created from the TLS registry config
  - I didn't cover message channel specific config like `mp.messaging.incoming.your-channel.tls-configuration-name=your-tls-config` I think it's tested upstream reasonably and don't want to add more tests
- adds test coverage for manual reloading of TLS registry config https://quarkus.io/guides/tls-registry-reference#reloading-certificates because I realized during TLS config reference review I missed that; we already have test coverage for periodic reloading of certs https://github.com/quarkus-qe/quarkus-test-suite/pull/1986

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/messaging/infinispan-grpc-kafka/src/main/java/io/quarkus/ts/messaging/infinispan/grpc/kafka/providers/SaslSslKafkaProvider.java
+++ b/messaging/infinispan-grpc-kafka/src/main/java/io/quarkus/ts/messaging/infinispan/grpc/kafka/providers/SaslSslKafkaProvider.java
@@ -1,47 +1,39 @@
 package io.quarkus.ts.messaging.infinispan.grpc.kafka.providers;
 
-import java.io.File;
 import java.util.Collections;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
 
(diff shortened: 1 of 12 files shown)


mjurc (approved):
Useful and nothing wrong I can see here, thanks for the coverage

PR #2083, opened 2024-10-11
Check kafka metrics independently from their order

fedinskiy:
### Summary

This test checks kafka metrics for being non-zero
At the moment, kafka-proudcer sends 3 metrics, and we only check first, but since they are ordered randomly, sometimes we check metric `kafka-consumer-dead-letter-topic-undeliverable-target` which is is always 0.0 (as expected).
To avoid that, now we only check metric for `alerts` topic (since we run other tests on it), and we check, that it is the only such metric, to avoid such kerfuffle in the future.

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:


fedinskiy:
run tests

michalvavrik (approved):
+1 for making this idempotent; thanks

PR #2091, opened 2024-10-14
Use periodic Quarkus snapshot build in PR workflows

michalvavrik:
### Summary

Spare cca 20 minuts in every PR. Not making change in the daily build as there no one is waiting and it might be smart to have independent build checked from time to time (in case upstream does something wrong?).

Big disadvantage is that if we want to enable some test after PR fix or if our CI is affected by something fixed in the main, we need to the next periodic build (every 24 hours) for the fix. We could have both (new approach and old workflow that builds Quarkus main) but it would add complexity.

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:


rsvoboda:
Nice, I didn't know there are Quarkus snapshots on sonatype. There are some unexpected missing artifacts - e.g. in CLI. Maybe we should do `cp .github/quarkus-snapshots-mvn-settings.xml ~/.m2/settings.xml` and drop `-s .github/quarkus-snapshots-mvn-settings.xml` from maven command

rsvoboda (approved):

Issue #2111, opened 2024-10-21
Don't connect to external domains in tests


mocenas:
Encountered issue when connecting to external domain `example.com` and the domain started to return HTTP code 499 (Token required) on some nodes. We probably hit some kind of threshold.  This was in test `io.quarkus.ts.http.restclient.reactive#sendRequestThroughProxy` - external domain is defined [here](https://github.com/quarkus-qe/quarkus-test-suite/blob/main/http/rest-client-reactive/src/main/resources/proxy.properties#L1).

It is IMHO generally a bad idea to use external services for testing - we don't have control over them and stuff like this might happen.

EDIT: the problem with access `example.com` was actually failure in DNS resolution in our infra. 
HTTP code 499 in nginx means something like "client closed request" nothing about token.

rsvoboda:
Please list other examples of other services so the scope is more clear.

mocenas:
Other examples + their tests counterparts

- https://github.com/quarkus-qe/quarkus-test-suite/blob/main/properties/src/main/resources/application.properties#L31 + https://github.com/quarkus-qe/quarkus-test-suite/blob/main/properties/src/test/java/io/quarkus/ts/properties/bulk/ConfigValueIT.java#L19
- https://github.com/quarkus-qe/quarkus-test-suite/blob/main/messaging/cloud-events/amqp-binary/src/main/java/io/quarkus/ts/messaging/cloudevents/amqpbinary/Producer.java#L31 + https://github.com/quarkus-qe/quarkus-test-suite/blob/main/messaging/cloud-events/amqp-binary/src/test/java/io/quarkus/ts/messaging/cloudevents/amqpbinary/BaseBinaryCloudEventsOverAmqpIT.java#L30
- https://github.com/quarkus-qe/quarkus-test-suite/blob/main/messaging/cloud-events/amqp-json/src/main/java/io/quarkus/ts/messaging/cloudevents/amqpjson/Producer.java#L39 + https://github.com/quarkus-qe/quarkus-test-suite/blob/main/messaging/cloud-events/amqp-json/src/test/java/io/quarkus/ts/messaging/cloudevents/amqpjson/BaseJSONCloudEventsOverAmqpIT.java#L30
[...]

PR #2112, opened 2024-10-21
Disable ReactiveRestClientProxyIT on aarch64

mocenas:
### Summary

Disable ReactiveRestClientProxyIT on aarch. The test tries to connect to domain `example.com` and this domain return 499 (token required) on arch. We are probably hitting some threshold. 

It is IMHO generally bad idea to use domain we don't own/control in our TS. But it is going OK for other nodes, AFAIK fails only on aarch so disabling it only on it.

Filed an issue for this - https://github.com/quarkus-qe/quarkus-test-suite/issues/2111

- [ ] Dependency update
- [X] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/http/rest-client-reactive/src/test/java/io/quarkus/ts/http/restclient/reactive/DisabledOnAarch64.java
+++ b/http/rest-client-reactive/src/test/java/io/quarkus/ts/http/restclient/reactive/DisabledOnAarch64.java
@@ -1,0 +1,21 @@
+package io.quarkus.ts.http.restclient.reactive;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@Inherited
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@ExtendWith(DisabledOnAarch64Conditions.class)
+public @interface DisabledOnAarch64 {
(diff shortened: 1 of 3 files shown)


mocenas:
cc @fedinskiy

mjurc:
I don't really want to disable the test without any RCA/improvement - why would this fail only on aarch64 machines, when they're in the same network as the other bare metal ones?

I don't see the example domain anywhere in test, but even so, cannot we use anything else?

rsvoboda:
Disabling should happen when there is no other way or alternative.

Please try the options mentions in your comment @mocenas

mocenas:
First of - I tried to change this test to use local service instead of public `example.com` and failed with that. In this test we use nginx proxy to proxy some http communication against some http server. This proxy is running in docker and it has a big problem connecting to localhost services. (127.0.0.1 inside docker is something different than on machine).


**Main thing** - I was wrong in the root source. After more debugging turned out problem is DNS. Nginx needs to have specifically configured DNS server. We have "8.8.8.8" in it's config. Turns out that our infra is blocking any outgoing DNS queries. Even manually I was unable to resolve anything against any public DNS server, from aarch machine.

Unfortunately nginx+docker doesn't have any default DNS resolver and must have one configured. Accessing local resolver is also problematic. Docker should have a resolver accessible on 127.0.0.11 but ends with `send() failed (111: Connection refused) while resolving, resolver: 127.0.0.11:53` from what I found online, you need to somehow setup additional docker network and configure it (I didn't analyze that too far).

IMHO best solution would be if we can enable DNS queries from aarch machines outside. WDYT @mjurc ?

mocenas:
I'm closing this one, since we had a successful run of this test.

PR #2124, opened 2024-10-23
Reference images, in the POM file, that do support ppc64le or s390x.

redmark-redhat:
There is no upstream Infinispan for ppc64le or s390x, we must use the downstream Data Grid. Also, needed a mariadb 10 multi-arch image.

### Summary

- [x] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -893,7 +893,6 @@
                                         <mysql.80.image>registry.redhat.io/rhel8/mysql-80</mysql.80.image>
                                         <mariadb.103.image>registry.redhat.io/rhel8/mariadb-103</mariadb.103.image>
                                         <mariadb.105.image>registry.redhat.io/rhel8/mariadb-105</mariadb.105.image>
-                                        <mariadb.10.image>registry.redhat.io/rhel8/mariadb-1011</mariadb.10.image>
                                         <amq-streams.image>registry.redhat.io/amq-streams/kafka-35-rhel8</amq-streams.image>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -893,9 +893,12 @@
                                         <mysql.80.image>registry.redhat.io/rhel8/mysql-80</mysql.80.image>
                                         <mariadb.103.image>registry.redhat.io/rhel8/mariadb-103</mariadb.103.image>
                                         <mariadb.105.image>registry.redhat.io/rhel8/mariadb-105</mariadb.105.image>
+                                        <mariadb.10.image>registry.redhat.io/rhel8/mariadb-1011</mariadb.10.image>
```
michalvavrik, line 896: On one hand, it's your decision what you use and I won't block this change. But on the other hand, I have hard time you really need this. You only test OpenShift, not baremetal, and I didn't find any test that is using this image. I suppose I just missed the one? Which OpenShift test is failing because of this image, please?

redmark-redhat: @michalvavrik Right, we only want to tests in the Openshift mode. Now that I understand what `-Dtest=!OpenShiftTodoDemoIT` does I suspect it was set on a run when we hit the mariadb.10.image image pull from docker. I can remove it from the change.

michalvavrik: It's small thing, but yeah, please remove it, I'll let formatting validation check pass and merge this.

michalvavrik (approved):

PR #2127, opened 2024-10-24
Overwrite images also in surefire plugin

mocenas:
### Summary

In aarch64 and openshift-ibm-z-p-containers profiles, we overwrite image for maven-failsafe plugin. But no images for surefire. Thus tests that are not IT will run with the default images and fail. (I found out about this with AgroalPoolTest on aarch64). This PR should fix that.

No sure what test ibm-z is running, if this is actually beneficial for it. But at least this shouldn't break it.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -774,6 +774,15 @@
             <build>
                 <plugins>
                     <plugin>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <systemPropertyVariables>
+                                <postgresql.latest.image>registry.redhat.io/rhel9/postgresql-16</postgresql.latest.image>
+                                <mysql.80.image>registry.redhat.io/rhel8/mysql-80</mysql.80.image>
+                            </systemPropertyVariables>
+                        </configuration>
+                    </plugin>
+                    <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-failsafe-plugin</artifactId>
                         <executions>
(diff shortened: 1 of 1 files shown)


michalvavrik:
> No sure what test ibm-z is running, if this is actually beneficial for it. But at least this shouldn't break it.

They **should** only test OpenShift, because they basically run their tests on Fedora so it's waste of their time that thy have baremetal results. We have seen during this release that they actually are running baremetal but it goes down to the wrong mvn command. So I think we should leave IBM Z be.

michalvavrik:
Anyway, I don't mind changes, it's just for your information.

michalvavrik (approved):
I think changes make sense.

michalvavrik:
CI JVM failures are expected due to the MP bump upstream.

pom.xml
```diff
@@ -773,6 +773,15 @@
             </activation>
             <build>
                 <plugins>
+                    <plugin>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <systemPropertyVariables>
+                                <postgresql.latest.image>registry.redhat.io/rhel9/postgresql-16</postgresql.latest.image>
```
jedla97, line 780: Just curious if this shouldn't define latest version as it was recently changed. Saw that when looking at https://github.com/quarkus-qe/quarkus-test-suite/pull/2132

michalvavrik: My understanding was that motivation had been our util class expecting tag and this was easiest fix. I also expect @mocenas tested it so if it is failing he will come back :-)

mocenas: > My understanding was that motivation had been our util class expecting tag and this was easiest fix.

That is true. 

I completely forgot that adding that. AFAIK missing ":latest" does not break anything in this particular case, but I will add it anyway, just to be sure.

PR #2129, opened 2024-10-24
Use RH postgres image and fix prepare statements usage

mocenas:
### Summary

Postgres image from docker.io has completely different startup mechanism than RH images. On RH images using `--max_prepared_transactions=100` causes it to crash.

I'm changing the default docker image to be the RH one, and changing the prepared transaction enablement to RH-image-supported approach.

Is there a reason to use docker.io image anyway? RH image for postgres is publicly available. 

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -50,7 +50,7 @@
         <exclude.quarkus.cli.tests>**/QuarkusCli*IT.java</exclude.quarkus.cli.tests>
         <exclude.quarkus.devmode.tests>no</exclude.quarkus.devmode.tests>
         <!-- Docker images used by both surefire and failsafe plugin -->
-        <postgresql.latest.image>docker.io/library/postgres:16</postgresql.latest.image>
+        <postgresql.latest.image>registry.access.redhat.com/rhel9/postgresql-16</postgresql.latest.image>
         <mysql.80.image>registry.access.redhat.com/rhscl/mysql-80-rhel7:latest</mysql.80.image>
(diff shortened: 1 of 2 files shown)


mocenas:
Failures in OCP JVM are not related - OpenShiftPostgreSql(Reactive)SbIT are already failing in main, other two failures are not related to postgres

michalvavrik (commented):
I don't think I can decide it myself. I'll add few reviewers and please don't merge till we have at least 2-3 people and noone categorically disagreeing.

BTW sadly CI is only testing narayana-transaction module but this change affect other baremetal tests. Could you maybe run them so that we know it works, please?

fedinskiy (approved):
I see no problem with this change.

michalvavrik:
Reason why I am careful is that I'd expect Quarkus users to use latest Postgres (at least some of them will) and we can catch things early. I don't mind fixing one tests, but general change... I'll wait for others to provide their view.

mjurc (changes requested):
> Is there a reason to use docker.io image anyway? RH image for postgres is publicly available.

Yes, this gives us chance to test the latest postgres image. The assumption that people are only going to run Quarkus with Red Hat built Postgres is also erroneous, or at least baseless.

I really don't like this. We need to test with latest on main, and with the oldest supported Red Hat version and latest greatest on product to ensure we're not doing anything obviously wrong with this smoke test. This change will make the smoke test even less helpful, so this is getting nack from me

mocenas:
OK, since this is not a good way, I'm replacing this with #2135

PR #2135, opened 2024-10-25
Configure postgres transaction separately for images

mocenas:
### Summary

Since #2129 is not going to fly, we need another way how to use both community and product psotgres image with prepared transaction. Configuring two versions for this.

Please select the relevant options.

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport
- [X] New scenario (non-breaking change which adds functionality)
- [ ] This change requires a documentation update
- [ ] This change requires execution against OCP (use `run tests` phrase in comment)

### Checklist:
- [ ] Methods and classes used in PR scenarios are meaningful
- [X] Commits are well encapsulated and follow [the best practices](https://cbea.ms/git-commit/)

PR #2136, opened 2024-10-25
[3.15] Disable SSL and SASL_SSL Kafka and Infinispan tests in FIPS and native mode as these scenarios are currently not supported

michalvavrik:
### Summary

FIPS in native are currently not supported, we have product trackers to alert us if anything changes, for now we need to disable tests due to failures experienced in the `rhbq-3.15-baremetal-ts-native` Jenkins job.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/messaging/infinispan-grpc-kafka/src/test/java/io/quarkus/ts/messaging/infinispan/grpc/kafka/InfinispanKafkaIT.java
+++ b/messaging/infinispan-grpc-kafka/src/test/java/io/quarkus/ts/messaging/infinispan/grpc/kafka/InfinispanKafkaIT.java
@@ -18,6 +18,7 @@ import io.quarkus.test.bootstrap.InfinispanService;
 import io.quarkus.test.bootstrap.KafkaService;
 import io.quarkus.test.bootstrap.RestService;
 import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.scenarios.annotations.DisabledOnFipsAndNative;
 import io.quarkus.test.services.Container;
 import io.quarkus.test.services.KafkaContainer;
 import io.quarkus.test.services.QuarkusApplication;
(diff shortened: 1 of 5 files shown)


fedinskiy (commented):
Thank you! But given the situation, I would suggest to merge to main with backport label (I'll do a backport together with https://github.com/quarkus-qe/quarkus-test-suite/pull/2135)

michalvavrik:
> Thank you! But given the situation, I would suggest to merge to main with backport label (I'll do a backport together with #2135)

I have no plan to disable it in main. Who knows what will be in half of the year? I cannot affected version to Jumanji until I know it will fail with it. And I don't know if this will be still case for the next Mandrel or what feedback we get - like what if feedback is we need to rewrite it?

michalvavrik:
Also if I remember correctly this annotation is not present in current main, so I can do that, but it will require FW release so you will need to wait few hours.

michalvavrik:
@fedinskiy I don't really mind, if that is what you prefer, ok, but I cannot provide fix before FW is released or I add some unnecessary repetition. Again, I don't mind either of that, so I'll let you decide.

fedinskiy:
@michalvavrik we are not in a hurry, so waiting for several hours (or even days) is not a problem. Since this is not fixed in any version of Quarkus/Mandrel (yet) and given a speed these bugs are resolved with[1], we can reasonably expect, that this error will affect Jumanji, and possibly even releases after that. I would like to save us some time from investigation of these failures and disable it now.
Ideally, also post a link to this PR to these issues, so when they are fixed, we can enable/adjust these tests as well

[1] eg https://issues.redhat.com/browse/QUARKUS-4612

michalvavrik:
sure @fedinskiy

PR #2119, opened 2024-10-23
Add access logging tests

mocenas:
### Summary

Add tests to cover https://issues.redhat.com/browse/QUARKUS-4663

Please select the relevant options.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport
- [X] New scenario (non-breaking change which adds functionality)
- [ ] This change requires a documentation update
- [ ] This change requires execution against OCP (use `run tests` phrase in comment)

### Checklist:
- [X] Methods and classes used in PR scenarios are meaningful
- [X] Commits are well encapsulated and follow [the best practices](https://cbea.ms/git-commit/)

PR #2167, opened 2024-11-06
Fix `AccessLoggingIT` when test is executed with long working directory path

michalvavrik:
### Summary

When `AccessLoggingIT` is executed in native mode in FIPS-enabled environment, working directory path in Jenkins can be long, which according to https://github.com/quarkusio/quarkus/issues/44346 means that maximum log file size can get bigger than required size. Therefore I am applying workaround - making sure that path to log directory is short.

This test was added recently https://github.com/quarkus-qe/quarkus-test-suite/pull/2119 and the TD was backported, therefore this fix needs to be backported to 3.15 as well.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/AccessLoggingIT.java
+++ b/logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/AccessLoggingIT.java
@@ -15,6 +15,7 @@ import java.util.zip.ZipFile;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.OS;
 
 import io.quarkus.test.bootstrap.RestService;
 import io.quarkus.test.scenarios.QuarkusScenario;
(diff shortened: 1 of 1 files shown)


jcarranzan (approved):
It looks good to me, thank you.

fedinskiy:
@jcarranzan there are no OCP tests in this module: https://github.com/quarkus-qe/quarkus-test-suite/tree/main/logging/jboss/src/test/java/io/quarkus/ts/logging/jboss

michalvavrik:
This PR was based on this https://github.com/quarkusio/quarkus/issues/44346#issuecomment-2459744128 suggestion. Now that I think about it I am not quite sure it makes sense because IMHO the issue is that this part: 
```
 {"timestamp":"2024-11-05T15:57:06.816135Z","sequence":517,"loggerClassName":"org.jboss.logging.Logger","loggerName":"io.quarkus","level":"INFO","message":"Profile prod activated. ","threadName":"main","threadId":1,"mdc":{},"ndc":"","hostName":"rhel8-fips-xlarge-263","processName":"/home/hudson/hudson_workspace/workspace/rhbq-3.15-baremetal-ts-native/420ae479/quarkus-test-suite/logging/jboss/target/logging-jboss-1.0.0-SNAPSHOT-runner","processId":27639}
```
is bigger when it is native (because path to JDK is shorter than to native executable). Also I wonder why is it longer in FIPS, only diffrence I can see is that `rhel8-fips-xlarge-263` contains extra `fips` :-D. Nevertheless, let's hope that expert is right and dummy (me) is wrong. I'll backport it and rerun.

TL;DR; if the route cause is same as the reproducer that I documented in https://github.com/quarkusio/quarkus/issues/44346, then I don't know how this fixes it. It can only be tested in Jenkins so the easiest way is to simply run it. Will do it later today.

michalvavrik:
I have tried it and I it fails. I think that suggestion was wrong, I'll comment on the issue. As far as fix goes, only workaround would be to place native executables to /tmp/something. I suppose there is easier way and it is to just avoid "processName" in there completely. I'll experiment with it and ask for new review.

PR #2186, opened 2024-11-12
Add coverage for Quarkus #43422

jedla97:
### Summary

Adding coverage for https://github.com/quarkusio/quarkus/issues/43422 issue.

The PR to fix the original issue is https://github.com/quarkusio/quarkus/pull/43440
 
- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/BaseHttpAdvancedReactiveIT.java
+++ b/http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/BaseHttpAdvancedReactiveIT.java
@@ -25,6 +25,8 @@ import static org.apache.http.HttpStatus.SC_INTERNAL_SERVER_ERROR;
 import static org.apache.http.HttpStatus.SC_NOT_FOUND;
 import static org.apache.http.HttpStatus.SC_OK;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.contains;
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.in;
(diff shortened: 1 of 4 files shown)


jedla97:
run tests

jedla97:
Native is related will look at it

http/http-advanced-reactive/src/main/java/io/quarkus/ts/http/advanced/reactive/GreetingResource.java
```diff
@@ -1,8 +1,18 @@
...
+
 public class GreetingResource extends GreetingAbstractResource {
     @Override
     public String hello() {
         return "Hello from Quarkus REST";
     }
+
+    @Path("/sub-resource")
+    public GreetingSubResources helloFromSubResource() {
+        return CDI.current()
```
jcarranzan, line 14: Why use CDI directly instead of creating a field @Inject annotation? Are there some benefits to doing it here?

jcarranzan: Ok I see that the issue reproducer use it, so I guess could be related to the issue and you had that motivation to use it.

jcarranzan (approved):
LGTM thanks

rsvoboda (changes requested):
* `OPTIONS` is not covered
 * what about other HTTP methods and sub-resource use case?
 * test failures need investigation

jedla97:
Native issue was caused by this https://github.com/quarkusio/quarkus/issues/42976 so I create different resource class where I tested it + testing it with interface. 

Added `OPTION` request, what I tried it worked same even before the upstream PR.
Added `POST` method to sub-resource to check if it works (Personaly think that using it like this is more like edgecase)

jedla97:
run tests

jcarranzan (commented):
Thanks @jedla97 the tests look good to me and now are covering all topics mentioned.
I have two small suggestions:
1. Could you consider creating a reusable helper method since most of tests have similar structure body?
2. On the other hand, I am wondering if it would be good having some negative test that returns for instance a 405 status code where be appropiated.

jedla97:
> Could you consider creating a reusable helper method since most of tests have similar structure body?

@jcarranzan I created the methods for HEAD and OPTION requests. With that I combined them into one test.

> On the other hand, I am wondering if it would be good having some negative test that returns for instance a 405 status code where be appropiated.

Tbh I don't see much added value in these negative tests. Also what should be testing? Not able to send `DELETE` request as it's not defined on endpoint? If you insist on adding something like this I can but as I said I don't see any adding value as I believe that this is tested by unit test in Quarkus/Vert.x/RestEasy

jedla97:
run tests

jcarranzan:
Ok thanks, let's wait for the checks.

rsvoboda (approved):

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/BaseHttpAdvancedReactiveIT.java
```diff
@@ -102,6 +104,64 @@ public void abstractResourceWithPath() {
...
+    @DisplayName("Test Quarkus REST interface resource with @Path using OPTION and HEAD request")
+    void optionAndHeadRequestUsingInterfaceResourceWithPath() {
+        testOptionRequest(GREETING_ENDPOINT + "/interface-greeting");
+        testHeadRequest(GREETING_ENDPOINT + "/interface-greeting");
+    }
+
+    @Test
+    @Tag("https://github.com/quarkusio/quarkus/issues/43422")
+    @DisplayName("Test Quarkus REST using CDI getting the sub-resource for OPTION and HEAD request")
+    void cdiSubResourceOptionAndHeadRequest() {
```
michalvavrik, line 160: This is only method that fails with 3.15.1. Why are other tests `optionAndHeadRequestUsingInterfaceResourceWithPath` and `optionAndHeadRequestUsingAbstractResourceWithPath` tagged with `https://github.com/quarkusio/quarkus/issues/43422` if they are passing with 3.15.1?

jedla97: I added the tag there as this issue was main motivation to create that coverage. If you things tha tag shouldn't be there. I can create PR for its removal.

michalvavrik: Clear answer :-D : 

I don't really care, keep it as it. I just thought that if you tag something with an issue, it means that issue is tested by this. It can still be true that changed code is not failing but testing changes because it can test code that we didn't have covered previously. What I do when I verify bug tickets is that I look for tests tagged with that issue. Now I'll keep in mind that failing test tagged with an issue may not mean that issue is not fixed. It probably won't happen anyway.

michalvavrik: Also, why are other 3 tests added in this PR without this tag, is that because they were not added because of this issue or you just forgot?

rsvoboda: Goodcatch @michalvavrik. @jedla97 can you prepare a followup PR with backport label?

jedla97: > Also, why are other 3 tests added in this PR without this tag, is that because they were not added because of this issue or you just forgot?

There are more like to ensure that endpoint work. The quarkus fix was to add `HEAD` and `OPTION`. At least this is my logic behind adding the tag here.

> @jedla97 can you prepare a followup PR with backport label?

@rsvoboda with adding or removing the tags?

michalvavrik: I think this kinda depends on how other QE members use `@Tag`, I always thought that it signals that some test tests the issue/ticket. But maybe that was just me.

michalvavrik:
@jedla97 I think we should try to have PR titled with a text that describes what PR does, you checked https://cbea.ms/git-commit/, `coverage for Quarkus #43422` is less informative then `Test support for @HEAD and @OPTIONS in sub-resources` or something like that. I realize sometimes it is really hard to write correct title (and I often fail).

http/http-advanced-reactive/src/test/java/io/quarkus/ts/http/advanced/reactive/BaseHttpAdvancedReactiveIT.java
rsvoboda: `@Tag` usage ... there is no rule really, some members started to use it, but we didn't have any discussion and  conclusion on its usage.

I would personally use `@Tag` only on tests testing the bug. I wouldn't tag the other related tests.

Issue #2025, opened 2024-09-18
No Spring Cloud Config server container to run as test service on aarch64


mjurc:
Currently, we run test services via containers. The only Spring Cloud Config server container  that exists and we use does not build containers for aarch64, only for x86-64. 

At the same time, the software itself is supported on aarch64, as it is Java application. Maybe the service could be ran out of container, or some other solution could be devised.

PR #2195, opened 2024-11-14
Run SpringCloudConfigIT on Aarch64

michalvavrik:
### Summary

- closes https://github.com/quarkus-qe/quarkus-test-suite/issues/2025

I have rebuild Quarkus QE image and created manifest for both Aarch64 and AMD64. I also updated instructions how to do it here https://quay.io/repository/quarkusqeteam/spring-cloud-config-server?tab=info. Then I tested this new image in beaker executor and the test is now passing on Aarch64.

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/spring/spring-cloud-config/src/test/java/io/quarkus/ts/spring/cloud/config/SpringCloudConfigIT.java
+++ b/spring/spring-cloud-config/src/test/java/io/quarkus/ts/spring/cloud/config/SpringCloudConfigIT.java
@@ -4,7 +4,6 @@ import static org.hamcrest.Matchers.is;
 
 import org.apache.http.HttpStatus;
 import org.junit.jupiter.api.Tag;
-import org.junit.jupiter.api.condition.DisabledIfSystemProperty;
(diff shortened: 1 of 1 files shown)


rsvoboda:
Wouldn't be better to run SpringCloud as a test service? it's a Java app and depending on image seems problematic when considering future platforms and IBM P/Z?

michalvavrik:
> Wouldn't be better to run SpringCloud as a test service? it's a Java app and depending on image seems problematic when considering future platforms and

I am fixing one problem at time. If we are going to run it as a test service, we need a way to do it. Probably to adjust FW bit for this edge case. I am not against it, but it is dead easy to create Aarch64 machine and build there image.

> IBM P/Z?

Spring Cloud is only tested on baremetal and IBM P/Z tests only OpenShift.

mjurc:
It's an okay way to unblock this, let's go with this until we have proper solution

mjurc (approved):

PR #2197, opened 2024-11-14
Use only fake domain in our tests

michalvavrik:
### Summary

- closes https://github.com/quarkus-qe/quarkus-test-suite/issues/2111

The `ReactiveRestClientProxyIT` test really talked to `http://example.com`, now no external communication to `example.com` is run, however we assert that correct host was proxied by passing `$host` via Nginx back to the test.

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:


michalvavrik:
run tests

michalvavrik:
run tests

michalvavrik:
There doesn't seem to be a sense to rerun OCP now, I'll try it tomorrow as the cluster is in a bad shape.

michalvavrik:
I thought about it and what we need to test is that proxy options are set correctly, not the redirect to external domain.

michalvavrik:
run tests

michalvavrik:
@mocenas OCP native run didn't fail, I canceled it manually because it was started when OCP cluster was down and now it is waiting in a queue. I won't merge this till it is green, so please review when you get a time. Thanks

mocenas:
> I thought about it and what we need to test is that proxy options are set correctly, not the redirect to external domain.

I thought about it too. It "would be nice" if we could do an E2E test to verify that communication though proxy works. But as you states, we want to test connection and options-passing to the proxy, not proxy's ability to forward it outside.

mocenas (approved):
LGTM

michalvavrik:
I have re-triggered OCP native manually in Jenkins several times, it seems like it fails randomly on OOM and failing tests are changing. I can see periodic OCP native CI is also failing over OOM (among other) so I dare say it is not related (also not sure how could changes here cause OOM). For that reason I think it should be safe to merge.

PR #2200, opened 2024-11-18
Lower super size xmx to 5g

mocenas:
### Summary

After https://github.com/quarkusio/quarkus/issues/44216 is fixed, memory requirements for super size is back to low. I was able to have a successful test run with xmx=3g. IMHO we can put it back to 5g (or at least try it).

WDYT @michalvavrik ?

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


mocenas:
run tests

mocenas:
cc @rsvoboda

michalvavrik:
I am not sure what good this bring us. If we lower memory, I'd expect things will be slower. I'd use available resources to have test execution time as low as necessary. But I agree it makes sense to test it. I am neutral to this proposal, let me assign this to @rsvoboda instead.

mocenas:
IMHO it would be good to have a test that would catch another issue, when memory requirements sky rockets to huge amounts. Now it will not do that, since there is a lot of GB in reserve.

About time - on my PC it makes no difference in time, if it has xmx=10g or 5g.  (as I stated above, it works even with 3g, but I prefer not to risk lowering it that much).

michalvavrik (approved):
Alright, it's strange to me that available memory makes little time difference, but I trust you. Let's catch problems early.

super-size/many-extensions/pom.xml
```diff
@@ -10,8 +10,7 @@
     <artifactId>many-extensions</artifactId>
     <name>Quarkus QE TS: Many extensions</name>
     <properties>
-        <!-- set to 7g after we experienced https://github.com/quarkusio/quarkus/issues/44216 -->
-        <quarkus.native.native-image-xmx>10g</quarkus.native.native-image-xmx>
+        <quarkus.native.native-image-xmx>5g</quarkus.native.native-image-xmx>
```
michalvavrik, line 13: This is default value already https://github.com/quarkus-qe/quarkus-test-suite/blob/7a5a184040487cc031a1166e7c8cf600fbe5b236/pom.xml#L703 so setting it here makes no difference?

mocenas: Yes it makes difference, so we can probably remove it entirely.

mocenas: I removed it.

michalvavrik (approved):

super-size/many-extensions/pom.xml
michalvavrik: Thanks

PR #2201, opened 2024-11-18
Add git rebase to PR CI and fix not running full CI when pom.xml is changed

jedla97:
### Summary

This adding the rebase for us to test on latest version of branch, it help only detect the changed file in the PR. Also the rebase fail if the contributor have outdated branch in this case 16 or more commits (in my POV it should be enough)

To test the rebase I used the f66d24d805480689a62b4f863bb3c64ca23ae3e8 commit as base for this branch

This PR also fix the problem when main `pom.xml` was updated but the PR also contains other changes (updating tests etc.) so only selected modules were run. Wth this fix all the modules will be executed.

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,11 @@ jobs:
 
           for module in $MODULES
           do
+            # Check if files contains main pom.xml as we want to run all modules here.
+            if [[ $CHANGED_FILE =~ [^\/]pom\.xml|^pom\.xml ]]; then
+                CHANGED=""
+                break
+            fi
             if [[ $CHANGED_FILE =~ ("$module") ]] ; then
                 CHANGED=$(echo $CHANGED" "$module)
             fi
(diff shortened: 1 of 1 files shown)


jedla97:
@michalvavrik this should adress both issues which we hit this week. 

The fix for the pom should be definitely fixed

The second fix with running more modules I'll decision up to you as I think it's not "must have" thing. So if you don't agree with the solution I'll remove the commit.

michalvavrik:
Thanks, I'll probably look in the evening or early tomorrow, I am bit engaged ATM.

michalvavrik:
I am going to remove the `triage/flaky-test` label because the `PR - Linux - JVM build - Latest Version` workflow has started before https://github.com/quarkus-qe/quarkus-test-suite/pull/2211 was merged and I don't want to get confused when I analyze new flaky results.

michalvavrik (approved):
LGTM, thanks!

michalvavrik:
> This PR also fix the problem when main pom.xml was updated but the PR also contains other changes (updating tests etc.) so only selected modules were run. Wth this fix all the modules will be executed.

No, I think it is fine. It could be improved, like if you change `.github/*` file and something in specific module there will still be an issue, but we don't need perfect, this is much more frequent situation; thx

PR #2203, opened 2024-11-19
Disable TodoDemoIT on windows

mocenas:
### Summary

TodoDemoIT is failing on windows with issue https://github.com/quarkusio/todo-demo-app/issues/36, where is fails to build into uber-jar with quarkus since 3.16.1. This problem is not caused by our TS, it also happens with plain TodoDemoApp.

I tried to package some other apps from quickstart the same way, and encountered no problems. I don't know what is the root source of this problem, but the TodoDemoApp is using quarkus 3.11.0 by default and seems to not be updated much since.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


michalvavrik (approved):
Thanks.

PR #2205, opened 2024-11-19
Fix flaky AgroalPoolTest

michalvavrik:
### Summary

The `io.quarkus.ts.sqldb.panacheflyway.dbpool.AgroalPoolTest#idleTimeoutTest` test is flaky, in this PR I provide my best guess.

First, the reasno why is it flaky. Strictly speaking I wasn't able to reproduce it, but I found a trick to reproduce it, just drop the `getIdleBackgroundValidationMs` timeout. I'll explain later why that internal doesn't make sense anyway. When you drop this, you fill find out that `idleTimeoutTest` fails the first time and suceeds on rerun. The reason why it succeeds the second time is that only this one failed test is repeated.

- If you run `AgroalPoolTest` is run, `connectionConcurrencyTest` leaves connections opened and in use, so that when `idleTimeoutTest` is executed, it has extra connections.
- publisher created in the `connectionConcurrencyTest` continues querying database long after this test method as concluded and respective `afterEach` invocation has been invoked
- cancelling publishes inside `connectionConcurrencyTest` ensures that no queries are issued once this async cancellation concludes
- I moved the `connectionConcurrencyTest` to the end because only advantage would be to add yet new sleep till everything is cancelled

That alone fixes the issue, but I didn't stop there. More changes:

- we don't test validation, so I dropped it
  - before this PR, we wait for test method `idleTimeoutTest` we waited after last query (see ref. here https://agroal.github.io/docs.html): validationTimeout + reapTimeout + 200 millis, however:
  - reap timeout is standalone task not related to validation: https://github.com/agroal/agroal/blob/3cde889cd9c90cfed00ba494ebde9689e4837fb9/agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java#L780
  - validation added here https://github.com/quarkusio/quarkus/blob/d2ba09f1f1c82b15e93167556ff0b5c4f2302400/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/DataSources.java#L327 performed here https://github.com/agroal/agroal/blob/3cde889cd9c90cfed00ba494ebde9689e4837fb9/agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java#L395 simply checks time https://github.com/quarkus-qe/quarkus-test-suite/blob/9a671a90697ad681cbb5dbd44aa35d0d56bc8caf/sql-db/panache-flyway/src/main/resources/application.properties#L31 which is ALWAYS TRUE
  - we are niot testing this "background validation"
  - it has no function, we are just waiting extra period for no time
  - this point I am not sure about, but it seems to me that it is extra query, and whether connection is idle is determined by last use of connection https://github.com/agroal/agroal/blob/master/agroal-pool/src/main/java/io/agroal/pool/ConnectionHandler.java#L227 which I think also happens for https://github.com/agroal/agroal/blob/master/agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java#L423 whose interval was set also to 1 second. I am not quite sure if there isn't potential for race. nevertheless, I could not detect any so who knows...
- I canceled subscribers, because number of expected items doesn't necessarily match number of published events; but I am not waiting (simply because tests do not require it)
- `concurrentLoadTest` seems to have async queries so quick that the end pretty much with the method, but I have added latch to ensure test methods are independent

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


sql-db/panache-flyway/src/test/java/io/quarkus/ts/sqldb/panacheflyway/dbpool/AgroalPoolTest.java
```diff
@@ -137,23 +149,36 @@ public void borderConditionBetweenIdleAndGetConnectionTest() {
...
         for (int i = 0; i < events; i++) {
             Multi.createFrom()
                     .range(0, CONCURRENCY_LEVEL).subscribe()
-                    .with(n -> assertEquals(2, users.count(), "UnexpectedUser Amount"));
+                    .with(n -> {
+                        assertEquals(2, users.count(), "UnexpectedUser Amount");
+                        latch.countDown();
+                    });
         }
+        var allUserCountsDone = latch.await(2, TimeUnit.SECONDS);
```
mocenas, line 172: I would prefer to have a boolean here, instead of var. It's IMHO easier to read.

mocenas (approved):
Thanks for getting to the bottom of this. Waiting for threads to finnish and closing subscribers are IMHO good steps. I trust you background validation.

I have a small comment about variable types, but I don't insist on it. Feel free to leave it this way.

LGTM

sql-db/panache-flyway/src/test/java/io/quarkus/ts/sqldb/panacheflyway/dbpool/AgroalPoolTest.java
michalvavrik: Yes, you are right. And even more, IIRC if you assign boolean to `var` it becomes boxed. I think I just typed it by habit. I'll fix it.

PR #2209, opened 2024-11-21
Add test for coverage for scnerario using same ports for http and https

jcarranzan:
### Summary
This PR forms part of 3.15.2 backports set 1 and adds the test coverage for https://github.com/quarkusio/quarkus/issues/43373 that was fixed here https://github.com/quarkusio/quarkus/pull/43985/commits/e24f7a3170b846b6205ff5dbb12bf0c31ac9e025

Note: Before Quarkus 3.15.2 this scenario fails and the app hangs, currently fails.

- [ ] Dependency update
- [ ] Refactoring
- [] Backport

### Checklist:


http/vertx/src/test/resources/server.crt
```diff
@@ -0,0 +1,29 @@
+-----BEGIN CERTIFICATE-----
```
michalvavrik, line 1: No :-) You are adding some hardcoded certificates without any information how you created it and I have no idea if it works in FIPS. Please configure HTTPS using FW. It should be enough to do something like `@QuarkusApplication(ssl = true, certificates = @Certificate(format = PEM, configureHttpServer = true, configureKeystore = true, configureTruststore = true))` and you can drop 

```
.withProperty("quarkus.http.ssl.certificate.files", "server.crt")
.withProperty("quarkus.http.ssl.certificate.key-files", "server.key");
````

You can also set `useTlsRegistry = false` if you insist on avoiding TLS registry but I expect it is irrelevant for your test so don't do it.

jcarranzan: Ok thanks @michalvavrik , I will change to use FW.

jedla97 (changes requested):
I have one comment. 
Also tested it with 3.15.1 and the quarkus is not terminated so the proces stay open and blocking port 8080 for any usage. I don't like as it can couse problems on other test which possibly using 8080 port. When I thinking about it I see `WARNING [app] Service didn't start in PT5M minutes` in log so maybe it should be done on side of framework, can you look at it?

michalvavrik:
> Also tested it with 3.15.1 and the quarkus is not terminated so the proces stay open and blocking port 8080 for any usage.

Kudos!

> Also tested it with 3.15.1 and the quarkus is not terminated so the proces stay open and blocking port 8080 for any usage.

Is this ever going to be run with 3.15.1? If there is not an easy fix on the FW side, you can go the other way around, exclude this from execution in 3.15.1.

PR #2215, opened 2024-11-25
Add coverage for HTTP 1.0 request with no host header

jedla97:
### Summary

Adding test to check http 1.0 request work correctly without headers. Checked request with wireshart and the request is same as expected.

Motivated by https://github.com/quarkusio/quarkus/discussions/42636

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


michalvavrik (approved):
I think this is fine. I'd prefer to also check response body (which I assume is `ok` string) because response headers and status doesn't really mean everything went fine. but this reproduces issue and verifies that fix is in place, thanks for the creative solution, I didn't even know how to work with `Socket`.

PR #2216, opened 2024-11-25
Detect changed long-running tagged tests to run in CI

gtroitsk:
### Summary

Our PR CI was configured to skip tests annotated `@Tag("long-running")` even if they were changed. This PR detects changed long running test classes and runs them separately

Example run: https://github.com/gtroitsk/quarkus-test-suite/pull/4/checks

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,18 +29,29 @@ jobs:
           MODULES=$(find -name pom.xml | sed -e 's|pom.xml| |' | sed -e 's|./| |' | grep -v " quarkus/" | grep -v resources)
           CHANGED=""
           MODULES_ARG=""
+          EXCLUDE_LONG_RUNS="-DexcludedGroups=long-running"
 
           # If changed file have some special character, its path is surrounded with quotes which causing the if statement fail
           CHANGED_FILE=$(echo ${{ steps.files.outputs.all_changed_and_modified_files }} | sed 's/\"/\\"/')
 
           for module in $MODULES
           do
-            # Check if files contains main pom.xml as we want to run all modules here.
-            if [[ $CHANGED_FILE =~ [^\/]pom\.xml|^pom\.xml ]]; then
+            # Check if files contains main pom.xml or changes github workflows as we want to run all modules here.
+            if [[ $CHANGED_FILE =~ [^\/]pom\.xml|^pom\.xml|^\.github/ ]]; then
                 CHANGED=""
+                EXCLUDE_LONG_RUNS=""
                 break
             fi
+          
             if [[ $CHANGED_FILE =~ ("$module") ]] ; then
+                # Check if module contains tests with @Tag("long-running")
+                for file in $(find "$module" -type f -name "*.java"); do
+                  if grep -q '@Tag("long-running")' "$file"; then
+                    EXCLUDE_LONG_RUNS=""
+                    break
+                  fi
+                done
+          
                 CHANGED=$(echo $CHANGED" "$module)
             fi
           done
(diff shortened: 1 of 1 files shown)


michalvavrik (changes requested):
Daily build is green, so it's reasonable assumption that failures are related, please have a look. Regarding changes, jeez, I'll have to think about it, I don't like that it took so many changes, but as said, I'll review when the CI is green and think :-)

.github/workflows/ci.yml
```diff
@@ -232,13 +271,16 @@ jobs:
...
           if [[ -n ${MODULES_ARG} ]]; then
             echo "Running modules: ${MODULES_ARG}"
-            MODULES_MAVEN_PARAM="-pl ${MODULES_ARG}"
+          mvn -B --no-transfer-progress -fae -s .github/quarkus-snapshots-mvn-settings.xml clean verify -Dall-modules -pl ${MODULES_ARG} -am -DexcludedGroups=long-running
+          fi
+          
+          # Long running tests
+          if [[ -n ${MODULES_LONG_RUNS_ARG} ]]; then
+            echo "Running modules: ${MODULES_LONG_RUNS_ARG}"
+            mvn -B --no-transfer-progress -fae -s .github/quarkus-snapshots-mvn-settings.xml verify -Dall-modules -pl ${MODULES_LONG_RUNS_ARG} -am
```
michalvavrik, line 282: Hey @gtroitsk , please patience, but I'll need to understand what you are trying to do. What I see:

- you detect long-running tests
- if they are among changed files, you run them

but if the same module (project) is changed, but not the long-running tests, how exactly do you know that long-running test wasn't affected by changes? I doubt that long-running tests are isolated, so if they can be affected by changed files around (like the application.properties that long-running test use), you don't want to run them?

gtroitsk: Script scans all changed files in module and if module contains long-running annotated file adds this module to `MODULES_LONG_RUNS_ARG` var. You can see printed modules for each category in first jobs from https://github.com/gtroitsk/quarkus-test-suite/actions/runs/12012099361/job/33482565059?pr=4

michalvavrik: You did not answer @gtroitsk . Please read my comment and answer why would you ignore module changes that can affect long-running test. Thanks

.github/workflows/ci.yml
```diff
@@ -22,31 +22,51 @@ jobs:
...
+                # Check if changed file contains @Tag("long-running")
+                LONG_RUNNING_TAG_FOUND=false
+                for file in $(find "$module" -type f -name "*.java"); do
+                  if grep -q '@Tag("long-running")' "$file"; then
+                    LONG_RUNNING_TAG_FOUND=true
+                    break
+                  fi
+                done
+          
+                if $LONG_RUNNING_TAG_FOUND; then
```
jedla97, line 44: Looking at this and I think you work with outdated version of action file. We merged  https://github.com/quarkus-qe/quarkus-test-suite/pull/2201 last week so you probably need to update it.

.github/workflows/ci.yml
gtroitsk: @michalvavrik [This change](https://github.com/gtroitsk/quarkus-test-suite/pull/4/files#diff-48d661ee020bbd17564ba9adaa63e9f5291940c08d716a76385c3640f2d3943a) in module that has annotated test(but not changed even class) address situation that you mentioned, because I don't understand

.github/workflows/ci.yml
gtroitsk: Thank you, I will rebase PR

.github/workflows/ci.yml
```diff
@@ -40,24 +42,43 @@ jobs:
                 CHANGED=""
```
michalvavrik, line 42: If test with `@Tag("long-running")` has been modified together with a root project POM file, this long running test will not be tested, right?

.github/workflows/ci.yml
```diff
@@ -130,7 +157,11 @@ jobs:
           ./quarkus-dev-cli version
       - name: Build with Maven
         run: |
-          mvn -fae -V -B --no-transfer-progress clean verify -Dinclude.quarkus-cli-tests -Dts.quarkus.cli.cmd="${PWD}/quarkus-dev-cli"${{ matrix.module-mvn-args }} -am -DexcludedGroups=long-running
+          mvn -fae -V -B --no-transfer-progress clean verify -Dinclude.quarkus-cli-tests -Dts.quarkus.cli.cmd="${PWD}/quarkus-dev-cli" ${{ matrix.module-mvn-args }} -am -DexcludedGroups=long-running
+          # Long running tests
+          if [[ -n "${{ matrix.module-long-run-mvn-args }}" ]]; then
+            mvn -fae -V -B --no-transfer-progress verify -Dinclude.quarkus-cli-tests -Dts.quarkus.cli.cmd="${PWD}/quarkus-dev-cli" ${{ matrix.module-long-run-mvn-args }} -am
```
michalvavrik, line 163: useless thoughts: This mean you will have 2 test executions runs in one workflow, so it's fine in regards of failsafe reports, we can ignore loss of a flaky test report in such a rare case when this is run, ok, but there is still a readability. I always scroll down and read results, now you will need to go through long logs to find first results or have a search term like `build success`.

questions:

1.  wouldn't it be clearer if you created next step that is only executed when `matrix.module-long-run-mvn-args`  is set?
2. are there any long running CLI tests or do you think it is likely there will be? it would make this command shorter if you dropped CLI stuff

I'll leave you to decide on both topics, just want to make sure you thought about that and current state is what you prefer.

.github/workflows/ci.yml
gtroitsk: Yes, for this version. I am working on adding support to run all modules include long-running in [my forked project](https://github.com/gtroitsk/quarkus-test-suite/pull/4/files) - you can expect something like that

michalvavrik: ok, just request review again when you are done.

for record, I think this is PR is fine, I think we don't need perfect version because IMHO it is super edge case. personally, I'd much more prefer less verbose version that drops `excludedGroups` when it detects changes in at least one module with a long running tag and that would be it. IMO bash in the workflow is hard to maintain, debug and keep consistent. anyway, I'll go with whatever you bring as long as it works.

.github/workflows/ci.yml
gtroitsk: 1. I think it can be possible solution - single step for long running tests and all modules run
2. I don't expect CLI tests will have long waiting condition, so I think we can drop it for now

jedla97:
Thinking aloud, wouldn't be better in this case if we detect long running test, just remove the `-DexcludedGroups=long-running` ?

Like by default we would have variable (LONG_RUNNING_ARGUMENT) which contains `-DexcludedGroups=long-running` and if the long running tag is detected this variable will be emptied.

Maybe you think about it but there were some problems.

gtroitsk:
@michalvavrik I think I have end nicer solution

Example runs:

- Include changes in log running test module
 https://github.com/gtroitsk/quarkus-test-suite/actions/runs/12048806018

- Without changes in log running test module
https://github.com/gtroitsk/quarkus-test-suite/actions/runs/12048698402

michalvavrik:
> Looking at it and don't see anything wrong. Thanks for improvement! I'll monitor PR CI for few day if I spot something unexpected.

Thanks, I'll merge it now and we can torture @gtroitsk if something goes wrong later.

PR #2220, opened 2024-11-26
Add coverage for ArC fix when a framework bean uses the decorator

jedla97:
### Summary

This adding the coverage for https://issues.redhat.com/browse/QUARKUS-5178. It's tested with Quarkiverse extension as the reproducer attached in issue and it was only scenario which I was able to reproduce.

To test this you need external lib which contains class (lib_class) implementing the interface. This interface is also aplied on class anotated as decorator in Quarkus app when the bean of lib_class is injected at the same time.

This error happen randomly in dev/test mode.

The Quarkus PR is https://github.com/quarkusio/quarkus/pull/43245

I tried to use same test which they using but wasn't able to reproduce the original issue.

**Note for 3.8 backport: the relevant issue is** https://issues.redhat.com/browse/QUARKUS-5292

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/messaging/kafka-processor/pom.xml
+++ b/messaging/kafka-processor/pom.xml
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkus.ts.qe</groupId>
+        <artifactId>parent</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+        <relativePath>../..</relativePath>
+    </parent>
+    <properties>
+        <!-- Need to be this version, I wasn't able to reproduce QUARKUS-5178 on newer versions-->
+        <quarkus-kafka-streams-processor.version>2.0.1</quarkus-kafka-streams-processor.version>
+    </properties>
+    <artifactId>kafka-processor</artifactId>
+    <name>Quarkus QE TS: Messaging: Reactive Processor Quarkiverse</name>
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkiverse.kafkastreamsprocessor</groupId>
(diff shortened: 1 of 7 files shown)


mocenas (approved):
IMHO this PR is OK on technical part. 

I don't like that we are adding tests for quarkiverse extensions (which are AFAIK unsupported). But if that is the only way how to reproduce the issue, guess we'll go with it.

jedla97:
> I don't like that we are adding tests for quarkiverse extensions (which are AFAIK unsupported). But if that is the only way how to reproduce the issue, guess we'll go with it.

I would love it other way but I was not able to reproduce it another way (spend few days on it). It's fixing the race condition in dev/test mode, so upstream coverage can also work but at the same time it can occur once in 50 runs.

Personally think that there is small number of cases where this happen.

Maybe cc @rsvoboda for info if he agree with this coverage of the [QUARKUS-5292](https://issues.redhat.com/browse/QUARKUS-5292)

rsvoboda:
Let's have it for 3.15 branch. I wouldn't backport it to 3.8. 

If the fixed version of https://github.com/quarkiverse/quarkus-kafka-streams-processor/ makes troubles in the future, we can drop it.

messaging/kafka-processor/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkus.ts.qe</groupId>
+        <artifactId>parent</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+        <relativePath>../..</relativePath>
+    </parent>
+    <properties>
+        <!-- Need to be this version, I wasn't able to reproduce QUARKUS-5178 on newer versions-->
+        <quarkus-kafka-streams-processor>2.0.1</quarkus-kafka-streams-processor>
```
rsvoboda, line 12:
```suggestion
        <quarkus-kafka-streams-processor.version>2.0.1</quarkus-kafka-streams-processor.version>
```
Plus where the property is used ...
Motivation: to be in sync with naming in root pom.xml

rsvoboda (changes requested):

messaging/kafka-processor/pom.xml
jedla97: Good point, fixed

rsvoboda (approved):

PR #2275, opened 2025-01-17
Tests for rest-client-reactive URL override feature

mocenas:
### Summary

Implement test coverage for https://github.com/quarkusio/quarkus/pull/43331

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/UrlOverrideableClient.java
+++ b/http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/UrlOverrideableClient.java
@@ -1,0 +1,37 @@
+package io.quarkus.ts.http.restclient.reactive;
+
+import java.net.URI;
+import java.net.URL;
+
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.Path;
+
+import org.eclipse.microprofile.rest.client.inject.RegisterRestClient;
+
+import io.quarkus.rest.client.reactive.Url;
+
+/**
+ * RestClient which provides capability to override base URL.
+ * URL might be overridden to access different service.
+ * Base URL can be overridden by String, URL or URI parameter type.
+ */
+@RegisterRestClient
+@Path("urlOverride")
+public interface UrlOverrideableClient {
+
+    @GET
+    @Path("basic")
+    String getByString(@Url String uri);
+
+    @GET
+    @Path("basic")
+    String getByUrl(@Url URL uri);
+
+    @GET
+    @Path("basic")
+    String getByUri(@Url URI uri);
+
+    @GET
+    @Path("basic")
+    String getNotOverridden();
+}
--- a/http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/resources/UrlOverrideResource.java
+++ b/http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/resources/UrlOverrideResource.java
@@ -1,0 +1,19 @@
+package io.quarkus.ts.http.restclient.reactive.resources;
+
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.Path;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+@Path("urlOverride")
+public class UrlOverrideResource {
+
+    @ConfigProperty(name = "ts.quarkus.urlOverride.response", defaultValue = "empty")
+    String response;
+
+    @GET
+    @Path("basic")
+    public String basic() {
+        return response;
+    }
+}
--- a/http/rest-client-reactive/src/test/java/io/quarkus/ts/http/restclient/reactive/UrlOverrideClientIT.java
+++ b/http/rest-client-reactive/src/test/java/io/quarkus/ts/http/restclient/reactive/UrlOverrideClientIT.java
@@ -1,0 +1,127 @@
+package io.quarkus.ts.http.restclient.reactive;
+
+import static io.restassured.RestAssured.given;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.bootstrap.JaegerService;
+import io.quarkus.test.bootstrap.Protocol;
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.JaegerContainer;
+import io.quarkus.test.services.QuarkusApplication;
+import io.restassured.response.Response;
+
+@QuarkusScenario
+@Tag("QUARKUS-5612")
+public class UrlOverrideClientIT {
+    private static final String MAIN_SERVICE_NAME = "overridingService";
+
+    // used for jaeger filtering, to only query traces relevant for current test
+    private long testStartTime = 0;
+
+    /*
+     * Start two application. By default, restClient in app will do HTTP requests against app.
+     * But if we override the URL, it will do requests against app2.
+     * Both apps will be distinguished by value of "ts.quarkus.urlOverride.response" which they return as HTTP response.
+     * The URL override feature overrides base URL, but not HTTP path. So we cannot easily test it using just one app.
+     */
+    @JaegerContainer(expectedLog = "\"Health Check state change\",\"status\":\"ready\"")
+    static final JaegerService jaeger = new JaegerService();
+
+    @QuarkusApplication()
+    static RestService app = new RestService()
+            .withProperties("urlOverride.properties")
+            .withProperty("quarkus.otel.enabled", "true")
+            .withProperty("quarkus.otel.simple", "true")
+            .withProperty("quarkus.application.name", MAIN_SERVICE_NAME)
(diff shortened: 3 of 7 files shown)


http/rest-client-reactive/src/test/java/io/quarkus/ts/http/restclient/reactive/UrlOverrideClientIT.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.bootstrap.Protocol;
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.QuarkusApplication;
+
+@QuarkusScenario
+public class UrlOverrideClientIT {
```
fedinskiy, line 13: 1. What if we use non-existent url for override? How good/terrible is the error?
2. Does app writes overridden URL into log?

mocenas: 1. Non-existing URL causes Exception on the service. It is possible to catch it and pretend it is OK, but that might cover some actual failures going on.
2. No it does not not log rewriting URL.

fedinskiy: 1. I thought, that maybe we should check that error, and that it really happens, to have some non-happy-path tests. Do not know, whether it is a good idea, but please think about it.
2. And what about tracing? Is there any way for an observer to distinguish these calls?

fedinskiy (commented):
The existing tests are good, but please consider advanced topics suggested in the comment. If you they do not make much sense in this context — you're free to merge this issue.

rsvoboda:
Is there a TP?

mocenas:
> Is there a TP?

Yes there is https://github.com/quarkus-qe/quarkus-test-plans/pull/191

http/rest-client-reactive/src/test/java/io/quarkus/ts/http/restclient/reactive/UrlOverrideClientIT.java
mocenas: 1. I'm not against having tests for non-happy-path. In this case, redirecting it to non-existing URL. But what would we test exactly? That something, that AFAIK cannot do anything else than fail, actually fails? Should we check that exception thrown is actually `HostNotFoundException` or something similar?  I don't see much value in that.
2. I didn't try tracing. Possibly it would work. But IMHO the module we adding it to `rest-client-reactive` is already complicated enough and I'm not sure what consequences might it have, to adding tracing to it.

http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/UrlOverrideableClient.java
```diff
@@ -0,0 +1,33 @@
...
+    @Path("basic")
+    String getByString(@Url String uri);
+
+    @GET
+    @Path("basic")
+    String getByUrl(@Url URL uri);
+
+    @GET
+    @Path("basic")
+    String getByUri(@Url URI uri);
```
michalvavrik, line 32: You override `@Url` in all the three client methods, how do you know that non-overriden method still works? Maybe consider testing it.

http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/resources/UrlOverrideResource.java
```diff
@@ -0,0 +1,21 @@
+package io.quarkus.ts.http.restclient.reactive.resources;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.Path;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+@ApplicationScoped
```
michalvavrik, line 9: Kinda curious why you specify `@ApplicationScoped`, it is not wrong if there is a point and it cause no harm, but I don't get it from the context. Thanks

http/rest-client-reactive/src/test/java/io/quarkus/ts/http/restclient/reactive/UrlOverrideClientIT.java
michalvavrik: I agree, the tracing test is relevant here. Let's get OTel involved. I have vague memory that I saw some related Quarkus issue, but maybe I just remember it wrong.

michalvavrik: If OTel shouldn't be here, let's check if we already have a module with OTel that can be used.

http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/UrlOverrideableClient.java
mocenas: I'm testing it. If you pass "null" as an argument, it leaves it non-overriden. (I took this approach from unit tests for the feature itself).

http/rest-client-reactive/src/test/java/io/quarkus/ts/http/restclient/reactive/UrlOverrideClientIT.java
mocenas: OK I agree that tracing test would be a good thing to have here. But since this looks like a big change, I will postpone it until I have time for it.

http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/UrlOverrideableClient.java
michalvavrik: That's not a same as if you have a REST client method without `@Url` though, it depends on the implementation. Nevertheless, I leave it to you to decide if you want to test it...

fedinskiy:
@mocenas there was no activity on this issue for three weeks. Are you planning any changes around it?

http/rest-client-reactive/src/main/java/io/quarkus/ts/http/restclient/reactive/UrlOverrideableClient.java
mocenas: I've added another method and test to cover non-override case.

http/rest-client-reactive/src/test/java/io/quarkus/ts/http/restclient/reactive/UrlOverrideClientIT.java
mocenas: I've added OTEL to the entire module and disabled it in default application.properties. For my local testing that caused no trouble for other tests. Also I've added testing the OTEL traces and test for failing the redirection on unknown host.

mocenas:
I've implemented the suggestion please re-review @michalvavrik  @fedinskiy.

http/rest-client-reactive/src/test/java/io/quarkus/ts/http/restclient/reactive/UrlOverrideClientIT.java
```diff
@@ -0,0 +1,127 @@
...
+            .withProperties("urlOverride.properties")
+            .withProperty("quarkus.otel.enabled", "true")
+            .withProperty("quarkus.otel.simple", "true")
+            .withProperty("quarkus.application.name", MAIN_SERVICE_NAME)
+            .withProperty("quarkus.otel.exporter.otlp.traces.endpoint", jaeger::getCollectorUrl);
+
+    @QuarkusApplication()
+    static RestService app2 = new RestService()
+            .withProperties("urlOverride.properties")
+            .withProperty("quarkus.http.ssl-port", "8444")
```
michalvavrik, line 50: I don't understand why are you configuring HTTPS port but never use it.

http/rest-client-reactive/src/test/java/io/quarkus/ts/http/restclient/reactive/UrlOverrideClientIT.java
```diff
@@ -0,0 +1,127 @@
...
+
+        String defaultResponse = app.given().get("/testUrlOverride/" + defaultEndpoint).asString();
+        assertEquals("default", defaultResponse);
+
+        String overriddenResponse = app.given()
+                .get("/testUrlOverride/" + overrideEndpoint + "/?port=" + secondAppPort)
+                .body().asString();
+        assertEquals("overridden", overriddenResponse);
+
+        verifyOtelTraces();
```
michalvavrik, line 99: Just because you set `quarkus.otel.simple` it doesn't mean that sending of traces _must_ be quicker than Quarkus respond arriving here. I think span export is still async. We can keep this and wait if this ever become flaky.

michalvavrik (approved):

http/rest-client-reactive/src/test/java/io/quarkus/ts/http/restclient/reactive/UrlOverrideClientIT.java
mocenas: App requires it. I'm not entirely sure where it is coming from, but at the start both apps try co claim port 8443 (if not changed) and one fails.

michalvavrik: You don't use HTTPS in your tests, if it comes from some global module application.properties, you could have disabled it or add comment because it feels confusing. Thanks for the explanation though, I think it is OK.

PR #2293, opened 2025-02-03
Fix build-time analytics module execution failure as we should only set QE FW test properties in the test properties

michalvavrik:
### Summary

As result of the https://github.com/quarkus-qe/quarkus-test-framework/pull/791 we expect that all QE FW properties are listed as enum constants in `Property`, however in past it was possible to set also Quarkus properties as `ts.global`, which now results in failure. Due to this issue, daily build fails https://github.com/quarkus-qe/quarkus-test-suite/actions/runs/13111089359/job/36575004262 over `[ERROR] java.util.ServiceConfigurationError: org.junit.platform.launcher.TestExecutionListener: Provider io.quarkus.test.listener.QuarkusTestResourceCleaningFilter could not be instantiated`. I have debugged it and found out that everytime `new PropertyLookup` constructor is instantiated in this module, loading of the global configuration fails as no such element.

This is expected to fail until https://github.com/quarkus-qe/quarkus-test-framework/pull/1485 is merged and this PR CI is re-run.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/build-time-analytics/src/test/java/io/quarkus/ts/buildtimeanalytics/AbstractAnalyticsIT.java
+++ b/build-time-analytics/src/test/java/io/quarkus/ts/buildtimeanalytics/AbstractAnalyticsIT.java
@@ -64,7 +64,9 @@ public abstract class AbstractAnalyticsIT {
                 .withStream(null)
                 .withExtraArgs("--no-code", "--no-wrapper")
                 .withExtensions(extensions);
-        return cliClient.createApplication(appName, createApplicationRequest);
+        QuarkusCliRestService cliRestService = cliClient.createApplication(appName, createApplicationRequest);
+        cliRestService.withProperty("quarkus.analytics.disabled", "false");
+        return cliRestService;
     }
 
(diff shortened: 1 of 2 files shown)


michalvavrik:
Sorry, I haven't seen failures as I had cached page result, I need to inspect them first.

michalvavrik:
I think the property has to be global, so service scope won't cut it. I added `application.properties` file which should be most precise solution.

fedinskiy (approved):
Thank you!

michalvavrik:
> Thank you!

thanks, yeah, I think CLI takes application properties from a different file or source. I had to go so I pushed it to see CI, I'll just find where CI is taking app properties from and fix it.

michalvavrik:
I think either we stop relying on global scope and use service context for the property lookup, or we need to set system property and it still will be just a trick. I'll provide a FW PR and once we agree on solution, I'll update this one.

michalvavrik:
hey @fedinskiy , thanks for the first review. Now CI is green, let's have a look. Thanks

fedinskiy (approved):

PR #2303, opened 2025-02-13
Move the start of TS CI to be executed 1 hour after Quarkus snapshot build

jedla97:
### Summary

The current CI is set to start at 0:30 UTC, but it's starting around 1:55 UTC (I don't know the reason), but as we now consuming the artifacts from snapshot repository, it causing sometimes problems. The main reason is that the Quarkus snapshot build start around 2:22 UTC, so some of the artifact which are used when testing are day old and some of them are latest. So this PR set the daily execution 1 hour after the Quarkus snapshot build which is defined in [deploy-snapshots.yml](https://github.com/quarkusio/quarkus/blob/main/.github/workflows/deploy-snapshots.yml#L5) 

I run it for 3 days now on my CI and the time when the daily start was around 3:12 UTC. The average build time of Quarkus 
is 30 min so the latest version of Quarkus should be always prepared before the start of our daily.

Link for Quarkus snapshot build CI: https://github.com/quarkusio/quarkus/actions/workflows/deploy-snapshots.yml

Note 1: The similar change was also done for our jenkins periodic runs.

Note 2: Maybe this is also reason why we sometimes saw the not know class failiure but wasn't able to reproduce it localy. And it's probably causing today failiures

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -2,7 +2,7 @@ name: "Daily Build"
 on:
   workflow_dispatch:
   schedule:
-    - cron: '30 0 * * *'
+    - cron: '0 3 * * *'
 jobs:
   linux-build-jvm-latest:
     name: Daily - Linux - JVM build - Latest Version


michalvavrik (approved):
lgtm

PR #2331, opened 2025-03-13
Add otel logging tests

mocenas:
### Summary

Add tests to cover OpenTelemetry log collection, according to the test plan in https://github.com/quarkus-qe/quarkus-test-plans/pull/196

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


michalvavrik (approved):
Perfect job. This PR complies https://github.com/quarkus-qe/quarkus-test-plans/blob/main/QUARKUS-5623.md.

One thing that is not in the TP and intrigues me is if there is a way to link OTel logs to _then_ current span so that you can get details about what was happening. And also about MDC, if you had MDC with trace id / span id, you could parse if of course, but you don't test MDC either. https://opentelemetry.io/docs/specs/otel/logs/#opentelemetry-solution speaks about _ recorded logs contained trace context identifiers (such as trace and span ids or user-defined baggage)_ and it feels like that is one of motivations?

michalvavrik:
run tests

monitoring/opentelemetry/src/test/java/io/quarkus/ts/opentelemetry/OpenTelemetryLoggingIT.java
```diff
@@ -0,0 +1,180 @@
...
+
+        // all messages should arrive in at most one sending period
+        await().atMost(7, TimeUnit.SECONDS).untilAsserted(() -> {
+            assertEquals(initLogLines + 22, getWarningLogCount(SERVICE_NAME),
+                    "All logged lines should arrive withing sending interval");
+        });
+    }
+
+    protected void validateHostname(LogEntry logEntry) throws UnknownHostException {
+        String hostname = InetAddress.getLocalHost().getHostName();
```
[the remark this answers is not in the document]
mocenas: Hostname is not localhost, it is "canonical name" or how is it called (e.g. rhel-large-123). Tests on OCP have this methods overriden to use pod-name.

AFAIK we have only two use cases - either tests are executed on same machine as application, or application is in the OCP.

For reason that it runs also on OCP I don't want to use anything like `localHostname`. And IMHO computer hostname is well defined term so I don't feel the need to comment this much. But if you think so I can add some comment to it.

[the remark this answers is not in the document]
michalvavrik: As you wish.

mocenas:
> Perfect job. This PR complies https://github.com/quarkus-qe/quarkus-test-plans/blob/main/QUARKUS-5623.md.
> 
> One thing that is not in the TP and intrigues me is if there is a way to link OTel logs to _then_ current span so that you can get details about what was happening. And also about MDC, if you had MDC with trace id / span id, you could parse if of course, but you don't test MDC either. https://opentelemetry.io/docs/specs/otel/logs/#opentelemetry-solution speaks about _recorded logs contained trace context identifiers (such as trace and span ids or user-defined baggage)_ and it feels like that is one of motivations?

I'm not sure about MDC, I didn't study that. But AFAIK logs and traces are independent on each other and often might not be even created by the same event (quarkus app is producing some logs even without any rest request). And even on the collector side logs and traces are collected via different app. AFAIK otel-LGTM (which we use) uses "loki" for logs and "tempo" for traces.

monitoring/opentelemetry/src/test/java/io/quarkus/ts/opentelemetry/OpenShiftOpenTelemetryLoggingIT.java
```diff
@@ -0,0 +1,20 @@
...
+
+@OpenShiftScenario
+public class OpenShiftOpenTelemetryLoggingIT extends OpenTelemetryLoggingIT {
+    @Inject
+    static OpenShiftClient ocClient;
+
+    @Override
+    protected void validateHostname(LogEntry logEntry) {
+        String hostname = ocClient.podsInService(app).get(0).getMetadata().getName();
+        assertEquals(hostname, logEntry.hostname(), "Hostname in the logs should be local machine hostname");
```
michalvavrik, line 18: Note, you don't need to take any action:

I am actually bit surprised this works. I think it means they take this information from incoming request, which raise interesting point what happen when there is no active incoming request like when scheduled task is executed...

michalvavrik: probably no hostname as there is no HTTP server in action...

mocenas: This is not a hostname of the http client, but hostname of the quarkus app's host. So hostname will always be filled.

michalvavrik: > This is not a hostname of the http client, but hostname of the quarkus app's host. So hostname will always be filled.

I am not sure what you mean by `hostname of the quarkus app's host` @mocenas, I don't know these terms. Quarkus application - `InetAddress.getLocalHost().getHostName()` returns `app-69469c4759-7s7h9`. Value of the configuration property `quarkus.http.host` is `0.0.0.0`. I inspected incoming request headers and it only contains `accept=*/*`, authority is `app-ts-lgveljyrzd.apps.ocp4-18.quarkus-qe.eng.redhat.com`.

That means that they use `InetAddress.getLocalHost().getHostName()` which I think resolves host address (e.g. `10.128.3.199` to the hostname). I think it means that it will work when there is no active incoming http request. This is using blocking DNS to resolve name (blocking operation cached for time configured in JVM). If we ever see occasional warnings about blocking, we should raise that. I don't know how it works internally anyway, so maybe it is sooo quick it won't be an issue.

mocenas: > I am not sure what you mean by hostname of the quarkus app's host

What system CLI command `hostname` will output on the server.

In this case hostname is purely server side thing, nothing to do with HTTP requests, client, etc. And it is also what `InetAddress.getLocalHost().getHostName()` should return (AFAIK java has no native way to access hostname without using it's network interface).
In OCP hostname is the same as pod name.


AFAIK config property `quarkus.http.host` basically select which network interface should the app listen on. While selecting 0.0.0.0 is basically wildcard to listen on all. Nothing to do with hostname.

michalvavrik: Alright, thanks for the explanation.

PR #2409, opened 2025-04-17
Fix Keycloak OpenShift scenarios on IBM Power Z by using Red Hat Build of Keycloak in OpenShift

michalvavrik:
### Summary

IBM Power/Z teams reported failures of:

- `io.quarkus.ts.security.keycloak.oidcclient.reactive.extended.OpenShiftOidcResponseFilterIT`
- `io.quarkus.ts.security.keycloak.OpenShiftMultiMechanismAuthenticationIT`

They can't use community Keycloak image, however the Red Hat Build of Quarkus is available for their system architecture. This change also enables previously disabled tests `OpenShiftOidcSinglePageAppLogoutFlowIT` and `OpenShiftOidcRestClientIT` test execution by IBM.

Please select the relevant options.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport
- [ ] New scenario (non-breaking change which adds functionality)
- [ ] This change requires a documentation update
- [ ] This change requires execution against OCP (use `run tests` phrase in comment)
- [x] This change requires execution with OCP on Aarch64 (use `run arm tests` phrase in comment)

### Checklist:
- [x] Methods and classes used in PR scenarios are meaningful
- [x] Commits are well encapsulated and follow [the best practices](https://cbea.ms/git-commit/)

PR #2419, opened 2025-04-22
Use Red Hat Build of Keycloak for OpenShift tests and thus enable IBM partners to run more Keycloak tests

michalvavrik:
### Summary

@ramdr suggested that if we start using the Keycloak image provided  by Red Hat, our IBM partners will be able to run more tests. This PR makes changes similar to what has been done in #2409. 

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:


michalvavrik:
run tests

michalvavrik:
run arm tests

jedla97 (approved):
Changes looks fine to me.

Just I forgot in last PR that RHBK is not supported on arm (they have one arm image, probably testing one). The one released is  updated 5 month ago (https://catalog.redhat.com/software/containers/rhbk/keycloak-rhel9/64f0add883a29ec473d40906?architecture=arm64&image=672b8dbcf734f2ece6d20eef)

I don't think it's that a big of the problem to test with this image. Only concern could be that with new RHBK the arm image won't be available, but we can overwrite in pom for arm profile.

michalvavrik:
@jedla97 valid point, but as long as ARM tests are passing, no action is required and I remember I tested the last PR with it. I think just because they don't declare support for that architecture, it doesn't mean it will fail, because it is the same team that creates both images. If I am wrong, we can follow your suggestion.

jedla97:
Yes, if test pass it's ok, I just mention it more like a note for awareness

michalvavrik:
run arm tests

PR #1962, opened 2024-09-02
Ensure abstract jakarta.ws.rs.core.Application is ignored

rsvoboda:
Ensure abstract jakarta.ws.rs.core.Application is ignored

Trigger: https://github.com/quarkusio/quarkus/pull/41465

Identified issue: https://github.com/quarkusio/quarkus/issues/42963

PASS:  `mvn clean verify -f http/http-minimum -Dreruns=0 -Dquarkus.platform.version=3.12.1`
FAIL: `mvn clean verify -f http/http-minimum -Dreruns=0 -Dquarkus.platform.version=3.12.0`
Reactive - `mvn clean verify -f http/http-minimum-reactive -Dreruns=0 -Dquarkus.platform.version=3.12.0`

Please select the relevant options.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport
- [x] New scenario (non-breaking change which adds functionality)
- [ ] This change requires a documentation update
- [ ] This change requires execution against OCP (use `run tests` phrase in comment)

### Checklist:
- [x] Methods and classes used in PR scenarios are meaningful
- [x] Commits are well encapsulated and follow [the best practices](https://cbea.ms/git-commit/)

PR #2457, opened 2025-05-13
Coverage for Native compilation failure - issue https://github.com/quarkusio/quarkus/issues/47033

jcarranzan:
### Summary

Coverage for native compilation failure - issue https://github.com/quarkusio/quarkus/issues/47033
Command used:
` mvn clean clean verify -Dit.test=PostgresqlDatabaseHibernateReactiveIT#nativeBuildCompilationWithValidList -Dnative -Dquarkus.native.container-build=true`

Command used with the Quarkus version that still fails:
` mvn clean clean verify -Dit.test=PostgresqlDatabaseHibernateReactiveIT#nativeBuildCompilationWithValidList -Dnative -Dquarkus.native.container-build=true -Dquarkus.platform.version=3.19.4`

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:


sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/hibernate/reactive/AbstractDatabaseHibernateReactiveIT.java
```diff
@@ -297,5 +301,21 @@ public void newLineInQuery() {
         assertThat(author.body().asString(), containsString("Slovník"));
     }
 
+    @Tag("https://github.com/quarkusio/quarkus/issues/47033")
+    @Test
+    public void nativeBuildCompilationWithValidList() {
```
jedla97, line 306: Thinking about this and I don't think it's needed. The coverage in my POV is that this module/quarkus app compile in native. Similar to https://github.com/quarkus-qe/quarkus-test-suite/pull/1962/files where it's just tested compilation. But maybe I'm mistaken

sql-db/hibernate-reactive/pom.xml
```diff
@@ -10,6 +10,17 @@
     <artifactId>hibernate-reactive</artifactId>
     <packaging>jar</packaging>
     <name>Quarkus QE TS: SQL Database: Hibernate Reactive</name>
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>io.quarkus.platform</groupId>
```
jedla97, line 16: Afaik when this will be tested with RHBQ I think it will fail as it won't find the correct version. Can you check it's possible to run this with 3.20.0 RHBQ?

Also I'm not sure if it's good idea to add cxf. I know without it it's hard/impossible to reproduce. It would be nice to have some thoughts about someone else if we want to cover it when it's this specific usecase containing cxf.

jedla97 (changes requested):

sql-db/hibernate-reactive/src/test/java/io/quarkus/ts/hibernate/reactive/AbstractDatabaseHibernateReactiveIT.java
jcarranzan: Yeah, you are right, I think I could keep the test to a healthy post-compilation flag, but I will remove the test eventually

sql-db/hibernate-reactive/pom.xml
jcarranzan: Yep thanks, the original issue involves a native compilation when these dependencies are present : 
```
- quarkus-hibernate-validator
- quarkus-cxf
- quarkus-mongodb-client
```

I've tried with `3.15.3.redhat-00004` : 
BOM: io.quarkiverse.cxf:quarkus-cxf-bom:pom:3.15.3.redhat-00008
JAR: io.quarkiverse.cxf:quarkus-cxf:jar:3.15.3.redhat-00008
Modifying the specific hibernate-reactive POM: 
`<properties>
    <rhbq.cxf-bom.version>3.15.3.redhat-00008</rhbq.cxf-bom.version>
</properties>
<dependencyManagement>
    <dependencies>
        <dependency>
            <groupId>io.quarkiverse.cxf</groupId>
            <artifactId>quarkus-cxf-bom</artifactId>
            <version>${rhbq.cxf-bom.version}</version>
            <type>pom</type>
            <scope>import</scope>
        </dependency>
    </dependencies>
</dependencyManagement>
<dependencies>
    <dependency>
        <groupId>io.quarkiverse.cxf</groupId>
        <artifactId>quarkus-cxf</artifactId>
    </dependency>
    </dependencies>`
    
    However, it's not possible to configure it like that, I'm encountering Non-resolvable import errors...
    Maybe @michalvavrik could shed some light or guidance here...

michalvavrik (changes requested):
Judging by comments in https://github.com/quarkusio/quarkus/issues/47033, reproducing this should be as easy as declaring a class required at runtime (just use it somewhere) that is also `org.graalvm.nativeimage.hosted.Feature`. Using `quarkus-cxf-bom` is no go.

sql-db/hibernate-reactive/pom.xml
michalvavrik: left comment separately of this comment thread because it's not just about this dep, but also about the way you reproduce it. @jedla97 is right in my eyes

jcarranzan:
After several attempts to reproduce the issue without the dependencies needed in the reproducer (`quarkus-cxf` and `quarkus-mongodb-client`) I couldn't reproduce it.
I tried reproducing the failure with a custom minimal `org.graalvm.nativeimage.hosted.Feature` .
But it seems the specific conditions created by `org.eclipse.angus.activation.nativeimage.AngusActivationFeature`  when `quarkus-mongodb-client` and `quarkus-cxf`  are also present and are necessary to trigger the issue that persists in 3.19.4.

michalvavrik:
> After several attempts to reproduce the issue without the dependencies needed in the reproducer (`quarkus-cxf` and `quarkus-mongodb-client`) I couldn't reproduce it. I tried reproducing the failure with a custom minimal `org.graalvm.nativeimage.hosted.Feature` . But it seems the specific conditions created by `org.eclipse.angus.activation.nativeimage.AngusActivationFeature` when `quarkus-mongodb-client` and `quarkus-cxf` are also present and are necessary to trigger the issue that persists in 3.19.4.

I would like to verify this statement myself before this is merged. Please don't merge until we check,  I think it is strange, but of course you are the one who invested your time into this, so you are more likely to be true.

michalvavrik:
Probably early next week...

PR #2476, opened 2025-05-26
Test OIDC token revocation using OidcProviderClient 

gtroitsk:
### Summary

TP: https://github.com/quarkus-qe/quarkus-test-plans/blob/main/QUARKUS-5653.md

`Security event-driven token revocation` cannot be really tested with Keycloak as it automatically revoke tokens during logout process and there is no way to distinguish between KC internal event and manual token revocation.
At least there is a [mock test](https://github.com/quarkusio/quarkus/blob/main/integration-tests/oidc-wiremock-logout/src/test/java/io/quarkus/it/keycloak/CodeFlowAuthorizationTest.java) on upstream side

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/security/keycloak/src/main/java/io/quarkus/ts/security/keycloak/OidcProviderClientResource.java
+++ b/security/keycloak/src/main/java/io/quarkus/ts/security/keycloak/OidcProviderClientResource.java
@@ -1,0 +1,46 @@
+package io.quarkus.ts.security.keycloak;
+
+import jakarta.inject.Inject;
+import jakarta.ws.rs.POST;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.core.Response;
+
+import io.quarkus.oidc.AccessTokenCredential;
+import io.quarkus.oidc.OidcProviderClient;
+
+@Path("/oidc-provider-client")
+public class OidcProviderClientResource {
+
+    @Inject
+    OidcProviderClient oidcProviderClient;
+
+    @Inject
+    AccessTokenCredential accessToken;
+
+    @POST
+    @Path("/token/revoke")
+    public String revokeAccessTokens(String refreshToken) {
+        oidcProviderClient.revokeAccessToken(accessToken.getToken()).await().indefinitely();
+        oidcProviderClient.revokeRefreshToken(refreshToken).await().indefinitely();
+        return "Logout, tokens revoked";
+    }
+
+    @POST
+    @Path("/username/from-token")
+    public Response getUsernameFromToken(String accessToken) {
+        return oidcProviderClient.getUserInfo(accessToken)
+                .onItem().transform(userInfo -> Response.ok(userInfo.getPreferredUserName()).build())
+                .onFailure().recoverWithItem(throwable -> Response.status(Response.Status.UNAUTHORIZED)
+                        .entity("Failed to resolve user info: " + throwable.getMessage())
+                        .build())
+                .await().indefinitely();
+    }
+
+    @POST
+    @Path("/token/is-active")
+    public String introspectToken(String passedAccessToken) {
+        return oidcProviderClient.introspectAccessToken(passedAccessToken).await().indefinitely().isActive()
+                ? "active"
+                : "inactive";
+    }
+}
--- a/security/keycloak/src/test/java/io/quarkus/ts/security/keycloak/BaseOidcSecurityIT.java
+++ b/security/keycloak/src/test/java/io/quarkus/ts/security/keycloak/BaseOidcSecurityIT.java
@@ -3,23 +3,30 @@ package io.quarkus.ts.security.keycloak;
 import static io.restassured.RestAssured.given;
 import static org.awaitility.Awaitility.await;
 import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.startsWith;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.http.HttpStatus;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.keycloak.authorization.client.AuthzClient;
+import org.keycloak.representations.AccessTokenResponse;
 
 import io.quarkus.test.bootstrap.KeycloakService;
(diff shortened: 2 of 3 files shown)


security/keycloak-webapp/src/main/java/io/quarkus/ts/security/keycloak/webapp/SecurityEventListener.java
```diff
@@ -0,0 +1,24 @@
...
+import io.quarkus.oidc.OidcProviderClient;
+import io.quarkus.oidc.RefreshToken;
+import io.quarkus.oidc.SecurityEvent;
+
+@ApplicationScoped
+public class SecurityEventListener {
+
+    public void event(@ObservesAsync SecurityEvent event) {
+        if (SecurityEvent.Type.OIDC_LOGOUT_RP_INITIATED == event.getEventType()) {
+            OidcProviderClient oidcProvider = event.getSecurityIdentity().getAttribute(OidcProviderClient.class.getName());
```
michalvavrik, line 16: I have commented out this whole listener and your tests are still passing. This tells me that you are not really testing this.

security/keycloak/src/main/java/io/quarkus/ts/security/keycloak/OidcProviderClientResource.java
```diff
@@ -0,0 +1,43 @@
...
+    @Inject
+    AccessTokenCredential accessToken;
+
+    @Inject
+    UserInfo userInfo;
+
+    @POST
+    @Path("/token/revoke")
+    public String revokeAccessTokens(String refreshToken) {
+        return oidcSession.logout()
```
michalvavrik, line 31: calling `oidcSession.logout()` may give someone idea that it actually does something for you, but if you drop it and leave just:

```
oidcProviderClient.revokeAccessToken(accessToken.getToken()).await().indefinitely();
oidcProviderClient.revokeRefreshToken(refreshToken).await().indefinitely();
return "Logout, tokens revoked";
```

pass anyway, AFAICT you are using bearer token auth, not code flow.

michalvavrik: and yes, I read docs and I know you copied it from there so docs suggest that, but that's only relevant when you have a web-app, you are doing a different thing

security/keycloak/src/main/java/io/quarkus/ts/security/keycloak/OidcProviderClientResource.java
```diff
@@ -0,0 +1,43 @@
...
+                .chain(() -> oidcProviderClient.revokeAccessToken(accessToken.getToken()))
+                .chain(() -> oidcProviderClient.revokeRefreshToken(refreshToken))
+                .map((result) -> "Logout, tokens revoked")
+                .await().indefinitely();
+    }
+
+    @GET
+    @Path("/username")
+    public String getUsername() {
+        return userInfo.getPreferredUserName();
```
michalvavrik, line 41: Quarkus OIDC internally uses `io.quarkus.oidc.OidcProviderClientImpl#getUserInfo`, however it never internally injects it (doesn't take it from CDI). However that was always the case. The fact that UserInfo is resolved correctly we know, it is omnipresent (it is very often used by Quarkus OIDC itself etc.).

What has changed with https://github.com/quarkusio/quarkus/pull/44993 is that `OidcProviderClient#getUserInfo` (apart of the fact  some user info code was moved internally from provider to provider client impl) is exposed to users, so that they can use it.

Using this new `Uni<UserInfo> getUserInfo(String accessToken);` method makes sense when request to user info didn't already happen (e.g. you have it disabled by default) or when you have access token that wasn't used to authenticate current `SecurityIdentity` (AKA the one that you didn't send as authorization header).

Based on this section https://github.com/quarkus-qe/quarkus-test-plans/blob/main/QUARKUS-5653.md#scope-of-testing I think that is what you wanted to set, not the `UserInfo` bean itself.

michalvavrik (changes requested):
I left couple of comments that have to be addressed. Additionally, there is `Uni<TokenIntrospection> introspectAccessToken(String accessToken);` on `OidcProviderClient` that you pledged to test in https://github.com/quarkus-qe/quarkus-test-plans/blob/main/QUARKUS-5653.md#scope-of-testing, but it is not tested here.

security/keycloak-webapp/src/main/java/io/quarkus/ts/security/keycloak/webapp/SecurityEventListener.java
michalvavrik: > The problem is that web app store tokens in session and they cannot be changed manually, only after creating new session for a client.

You want to test that access token / refresh tokens are actually revoked when this happens. You can access them, so you can use them. The web app can use the access token to communicate with other microservices using REST client. You can just use these tokens inside the Quarkus app after they were revoked (try to use the token, try to refresh the access token. There is no real difference to what you do in `security/keycloak/src/test/java/io/quarkus/ts/security/keycloak/BaseOidcSecurityIT.java`, but maybe I am missing something... 

> I don't know a way how to revoke tokens on web app, not just check that those methods were called as upstream test did.

I think if the OP like Keycloak notifies Quarkus that user logged out and the `OIDC_LOGOUT_RP_INITIATED` even is fired, tokens are already invoked unless... I don't remember that "unless" something, I suppose that Quarkus example only makes sense for certain scenarios. Maybe you need to actually test if these tokens are revoked or you can successfully use them. If they are already revoked, I can't see realistic user scenario with KC. However if you call the logout path directly, like via RestAssured, I expect they won't be revoked and then your test inside Quarkus app (before/after) revocation will make sense.

gtroitsk:
@jedla97 web-app scenario token revocation triggered by security event call is not working and I don't know how to really test it with Keycloak, I asked KC team to provide me some guidance, no answer yet. Please see comments in this PR and QQE ticket for more details

gtroitsk:
@jedla97 I updated PR and its description why we can't test Security event-driven token revocation with Keycloak. CI is failing due to Maven Central snapshot outage, I will retrigger CI as they fix it

gtroitsk:
@fedinskiy I'm reassigning the review to you since both previous reviewers are currently on PTO

fedinskiy:
From what I understand, `Security event-driven token revocation` is now out of scope for the PR, since we cannot test it. Please remove it from the test plan, then, or add an explanation there.

fedinskiy (commented):
And the last thing: is this tested in openshift? If yes, we need to reflect that in the Test plan.

PR #2514, opened 2025-07-15
Disable REST Client http tests made synchronous calls affected by issue #48927

gtroitsk:
### Summary

Our synchronous REST Client calls fail with `IllegalStateException: Response has been closed` due to upstream regressions - https://github.com/quarkusio/quarkus/issues/48927

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:


michalvavrik:
I think this PR is a bad idea @gtroitsk @jedla97 , why are we loosing tests for such a trivial reasons? Can you just create a new response instead? FWIW I don't think reusing the client response is required feature, it sounds like "optional feature", we should just insist on clear message and ideally validation. It certainly doesn't seem to be point of https://github.com/quarkus-qe/quarkus-test-suite/commit/fa479aa651057e5e487fda9939c05e9a189a2a54, I am against disabling.

jedla97:
For my POV the reason to disable tests was to wait before the issue is closed as fixed. But looking at latest response on issue it should be probably modified.

> FWIW I don't think reusing the client response is required feature, it sounds like "optional feature"

Yeah I think it was more like easier for me than creating. So I don't see big deal of creating new response or modify it in different way. The tests were created to cover the http2 version for rest client so there return new response shouldn't be problem.

PR #2657, opened 2025-09-25
[DRAFT] Ensure annotation processor respect depmgmt #2

mjurc:
* Adding configuration to Maven compiler plugin to ensure that annotation processor dependency resolution respects dependency management of the project rather than dependencies of the processor in each project declaring the processor since we run the test suite in different ways and the generated pom and build root changes depending on whether the test is ran from parent module or the individual child test modules.

Follows up on https://github.com/quarkus-qe/quarkus-test-suite/pull/2644/

### Summary

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/hibernate/hibernate-fulltext-search/pom.xml
+++ b/hibernate/hibernate-fulltext-search/pom.xml
@@ -57,6 +57,7 @@
                     <execution>
                         <id>default-compile</id>
                         <configuration>
+                            <annotationProcessorPathsUseDepMgmt>true</annotationProcessorPathsUseDepMgmt>
                             <annotationProcessors>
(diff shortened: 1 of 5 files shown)


rsvoboda (commented):
Ok, so having things in root pom.xml doesn't solve it properly because there are always different configurations in the modules?

rsvoboda:
About https://github.com/quarkus-qe/quarkus-test-suite/pull/2644, why wasn't `maven-compiler-plugin` definition added into `pluginManagement` part?

Please check if that isn't simpler solution instead of putting the property into every module that uses annotation processor

PR #2658, opened 2025-09-25
Disable tests due to quarkus#50247 (Unable to find datasource '<default>')

jedla97:
### Summary

Disabling tests/module due to https://github.com/quarkusio/quarkus/issues/50247. When the fix is available I'll revert this commit.

The security/jpa module needs to disabled,
otherwise there would need to modification to remove panache.

The super-size/many extension have ony disabled minimal number of extension to preserve maximal matrix for test.

In build-time-analytics only some extensions in set A and B were disabled.

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/build-time-analytics/src/test/java/io/quarkus/ts/buildtimeanalytics/AnalyticsUtils.java
+++ b/build-time-analytics/src/test/java/io/quarkus/ts/buildtimeanalytics/AnalyticsUtils.java
@@ -36,7 +36,8 @@ public class AnalyticsUtils {
             "quarkus-core",
             "quarkus-grpc",
             "quarkus-hibernate-orm",
-            "quarkus-hibernate-orm-panache",
+            // Enable when https://github.com/quarkusio/quarkus/issues/50247 is fixed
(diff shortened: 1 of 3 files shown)


jedla97:
> why is the whole security/jpa module disabled?

@rsvoboda it using the `quarkus-hibernate-orm-panache` which should be problematic extension causing it (not looked at that much) and we using some classes/interfaces from it. So I needed to disabled it.

I just checked the jpa again and realize that when setting `hibernate-orm.enable` to false, I should switch it back to true when starting the app.



I play with `quarkus.hibernate-orm.enabled=false` on `build-time-analytics` and I still got that error, but looking at it now it's probably because we creating the app using our cli wrapper. So we would need to pass the property somehow to it (I prefer to temporary remove few extension then looking at it and try to figure it out, when the upstream issue should be fixed).

For `many-extension` I did not try it, but I try it now and the `quarkus.hibernate-orm.enabled=false` seems work there.

PR #2662, opened 2025-09-29
Coverage for QUARKUS-6243 - execute HQL queries using DevUI

gtroitsk:
### Summary

Test HQL in Quarkus dev mode according to the [test plan](https://github.com/quarkus-qe/quarkus-test-plans/blob/main/QUARKUS-6243.md)

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/hibernate/hibernate-orm/src/main/java/io/quarkus/qe/hibernate/hql/DevUIJsonRpcClient.java
+++ b/hibernate/hibernate-orm/src/main/java/io/quarkus/qe/hibernate/hql/DevUIJsonRpcClient.java
@@ -1,0 +1,254 @@
+package io.quarkus.qe.hibernate.hql;
+
+import java.io.IOException;
+import java.net.URI;
+import java.util.Map;
+import java.util.Random;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import io.vertx.core.Vertx;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.http.HttpClient;
+import io.vertx.core.http.HttpClientOptions;
+import io.vertx.core.http.WebSocket;
+import io.vertx.core.http.WebSocketConnectOptions;
+
+// This class is copied from a Quarkus project with minor modifications
+// https://github.com/quarkusio/quarkus/blob/main/extensions/devui/test-spi/src/main/java/io/quarkus/devui/tests/DevUIJsonRPCTest.java
+public class DevUIJsonRpcClient {
+
+    protected URI uri;
+
+    private final ObjectMapper mapper = new ObjectMapper();
+    private final String namespace;
+    private final String DOT = ".";
+    private final Vertx vertx;
+    private final HttpClient client;
+
+    public DevUIJsonRpcClient(String namespace, String testUrl) {
+        // The namespace changed to be compatible with MCP. We add some code here to be backward compatible
+        if (namespace.contains(DOT)) {
+            namespace = namespace.substring(namespace.lastIndexOf(DOT) + 1);
+        }
+
+        this.namespace = namespace;
+        this.uri = URI.create(testUrl + "/q/dev-ui/json-rpc-ws");
+
+        this.vertx = Vertx.vertx();
+        this.client = vertx.createHttpClient(new HttpClientOptions()
+                .setDefaultHost(this.uri.getHost())
+                .setDefaultPort(this.uri.getPort()));
+    }
+
+    public <T> T executeJsonRPCMethod(TypeReference typeReference, String methodName) throws Exception {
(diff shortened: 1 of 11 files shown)


gtroitsk:
> @gtroitsk please fix the formatting
> 
> Any reason for having this as Draft?
> 
> Also PR description should have some content.

I want to add some tests with more complex scenarios. PR will be ready for review till EOD

hibernate/hibernate-orm/pom.xml
```diff
@@ -27,6 +27,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-jdbc-postgresql</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-jdbc-mysql</artifactId>
```
michalvavrik, line 32: Please consider dropping this from here and set `quarkus-jdbc-mysql` on `@DevModeQuarkusApplication` in `DevModePostgresqlHQLConsoleIT`, which is the only place where you use MySQL. It should allow you to avoid these new properties like:

```
quarkus.hibernate-orm.named.datasource=mysql
quarkus.hibernate-orm.named-2.datasource=mysql
quarkus.hibernate-orm.test-hql-pu.datasource=mysql
quarkus.datasource.postgres.db-kind=postgresql
quarkus.datasource.mysql.db-kind=mysql
quarkus.hibernate-orm.named.datasource=postgres
quarkus.hibernate-orm.named-2.datasource=postgres
```

and switch `quarkus.hibernate-orm.test-hql-pu.datasource=postgres` to the default.

This is not "hard no" on what you do, but we test extensively multiple drivers and named PUs and I am (literally) guessing that still normal applications simply use default datasource and have one drive. It is worth it. Plus the only place where MySQL is used is a DEV mode test, in which case there is literally no test execution increase downside.

michalvavrik (changes requested):
I am going to review within hour or three, but there is small change I'd prefer if you don't mind.

hibernate/hibernate-orm/src/main/java/io/quarkus/qe/hibernate/hql/DevUIJsonRpcClient.java
```diff
@@ -0,0 +1,243 @@
...
+                });
+            } else {
+                RESPONSES.put(id, new WebSocketResponse(ar.cause()));
+                vertx.close();
+            }
+        });
+        return id;
+    }
+
+    private static final ConcurrentHashMap<Integer, WebSocketResponse> RESPONSES = new ConcurrentHashMap<>();
```
michalvavrik, line 220: I am not fan of this class, but we can keep it while it works, it makes no sense to waste more time on it. But this part doesn't make sense to me and I think it can confuse results. It is a static class, but the `HQLConsoleClient` creates `new DevUIJsonRpcClient("quarkus-hibernate-orm", testUrl)`, so why is it not a member property instead?

The issue is that I know that at least in some cases, JUnit initialize test classes together (I don't know if it is every case or whatever) because I was fixing in a framework issue that one test class had stored state in static variables from other test class. can you try to keep the `RESPONSES` without `static` and discover why is it needed?

hibernate/hibernate-orm/src/test/java/io/quarkus/qe/hibernate/AbstractHQLConsoleIT.java
```diff
@@ -0,0 +1,131 @@
...
+@Tag("QUARKUS-6243")
+public abstract class AbstractHQLConsoleIT {
+
+    private static HQLConsoleClient hqlClient;
+    private static final int DEFAULT_PAGE_SIZE = 15;
+    private static final String PERSISTENCE_UNIT = "test-hql-pu";
+
+    @BeforeEach
+    public void beforeEach() {
+        hqlClient = new HQLConsoleClient(getApp().getURI(Protocol.HTTP).toString(), PERSISTENCE_UNIT);
```
michalvavrik, line 29: `HQLConsoleClient` has test class scope, so I don't understand why you create it per each test method. But if you want to do that for example to clear previous results, it is fine. It feels like a strange pattern:

1. store `hqlClient` in static variable
2. only use it in a member methods
3. create the `hqlClient` in a `beforeEach` but it only is used within one test method

I think the proper pattern would be that inside one method you use try-with-resources and close Vert.x/HttpClient with it. I don't wan to be annoying, so just drop the `static` and I can live with it.

michalvavrik (commented):
I left few comments, but most of them are just to you my thinking - coding. You don't need to follow them. I have 2 things that bother be though:

1. you only use "retrieve", there is no `create/update/insert/delete`, shouldn't you test it? such a commands have different requirements, like Quarkus generally require that you have active transaction to delete entity instance, so isn't it relevant to test?
2. You only test DEV UI backend, but users will not do that, they will use Quarkus DEV UI frontend. So how do we know it works? What you do here  makes sense, but have you considered adding a smoke check for FE?

gtroitsk:
> I left few comments, but most of them are just to show you my thinking - coding - in case it helps you. You don't need to follow them. I have 2 things that bother me though:
> 
> 1. you only use "retrieve", there is no `create/update/insert/delete`, shouldn't you test it? such a commands have different requirements, like Quarkus generally require that you have active transaction to delete entity instance, so isn't it relevant to test?
> 2. You only test DEV UI backend, but users will not do that, they will use Quarkus DEV UI frontend. So how do we know it works? What you do here  makes sense, but have you considered adding a smoke check for FE?

Michal, thank you for review. Answering your questions:
1. That definitely should be tested, I will prepare new scenarios.
2. I did a manual verification. I don’t think it is necessary to add automated FE checks for a dev-support tool, and parsing returned HTML tables would be complex and hard to maintain.

michalvavrik (approved):
LGTM

PR #2966, opened 2026-04-13
Enable testJDBCMetrics test for CLI update

jedla97:
### Summary

The issue is closed so we can enable this test

Also needed to refactor the tests as current flow not getting the properties directly from `application.properties` + the `quarkus update` not working in `target` directory 

Now it's same as other test for update property (like https://github.com/quarkus-qe/quarkus-test-suite/blob/main/quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/update/QuarkusCli327UpdatesIT.java#L132)

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/update/QuarkusCli333UpdatesIT.java
+++ b/quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/update/QuarkusCli333UpdatesIT.java
@@ -2,14 +2,12 @@ package io.quarkus.ts.quarkus.cli.update;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
-import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.List;
-import java.util.Map;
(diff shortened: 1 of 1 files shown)


jedla97:
Changing to draft, it seems that the issue is not fixed or there is something problematic. Need to look more into it when I have time

fedinskiy:
@jedla97 beware, that the fix probably was not backported to 3.33: https://github.com/quarkusio/quarkus-updates/pull/452

jedla97:
Good point @fedinskiy but from what I remember when testing the 3.27.0 we encounter that the quarkus update was updated and it break our tests. So it should matter on Quarkus version and backports.

But thanks as I was go to look at that issue for 3.27 and it was the issue what causing the update problem https://github.com/quarkusio/quarkus-updates/issues/394

I refactor it to be same as https://github.com/quarkus-qe/quarkus-test-suite/blob/main/quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/update/QuarkusCli327UpdatesIT.java#L132

The problem which I found when just adding `cliClient.useTemporaryDirectory()` was that `app.getProperties()` returned 0 properties and from my understanding it's because it's returning only properties which are add using `withProperties` and not ones which are in `applications.properties` files (https://github.com/quarkus-qe/quarkus-test-framework/blob/main/quarkus-test-core/src/main/java/io/quarkus/test/bootstrap/ServiceContext.java#L24)

fedinskiy (approved):

PR #2976, opened 2026-04-22
Fix DevModeInfinispanIT on Podman

SlimaneAbzar:
**Summary**

DevModeInfinispanIT fails on CI (quarkus-main-rhel8-jdk17-baremetal-ts-jvm-podman-nightly) because the Infinispan client tries to connect to the container's internal IP instead of localhost, which doesn't work with rootless Podman. This causes the books cache to be created without transaction support, resulting in CacheNotTransactionalException (ISPN004084). Setting client-intelligence=BASIC keeps the client on localhost and fixes the issue.

The other tests in this module (`InfinispanIT`, `TransactionalInfinispanIT`) already set `client-intelligence=BASIC` via `infinispan-it.properties`, but `DevModeInfinispanIT` only uses `application.properties` which was missing this setting.

I reproduced the failure locally using Podman and confirmed the fix resolves the error.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

**Checklist:**

- [x] Commits are well encapsulated and follow the best practices

### DIFF
--- a/nosql-db/infinispan/src/main/resources/application.properties
+++ b/nosql-db/infinispan/src/main/resources/application.properties
@@ -1,3 +1,4 @@
+quarkus.infinispan-client.client-intelligence=BASIC
 quarkus.infinispan-client.cache."cache".configuration-uri=cache-configuration.xml
 
 quarkus.infinispan-client.cache.books.configuration-resource=books-cache-config.xml


fedinskiy (approved):

fedinskiy:
@SlimaneAbzar could you also check, whether 3.33 and 3.27 are affected? If they are, then we may want to backport this patch.

SlimaneAbzar:
> @SlimaneAbzar could you also check, whether 3.33 and 3.27 are affected? If they are, then we may want to backport this patch.

Both 3.33 and 3.27 are affected, they have the same application.properties without client-intelligence=BASIC.

fedinskiy:
@SlimaneAbzar and does the test fail there?

if it does, then please mark the PR with `triage/backport-3.27` and/or `triage/backport-3.33`

SlimaneAbzar:
> @SlimaneAbzar and does the test fail there?
> 
> if it does, then please mark the PR with `triage/backport-3.27` and/or `triage/backport-3.33`

The test doesn't fail in 3.27 but fails in 3.33. I dont have the permissions to add the triage/backport-3.33 label.

PR #2977, opened 2026-04-24
Decrease nuber of started Dev mode apps at same time

jedla97:
### Summary

Running 5 devmode apps at the same time need lot of resources. Running this tests on 8GB system cause some unstability when the system needs some amount of ddle memory. For comparison our RHEL8 VM needs using around 300ory and RHEL 9 using 650MB.

The VMs running on the edge of out of memory when running this single test. When runnign the full module I belive there is some allocated memory which won't be freed and causing OOM for system.

For measuring I used the `watch -n 5 free -m`

These are the results of used memory:
```
| OS           | RHEL 8 | RHEL 9 |
| idle memory  | 274    | 646    |
| Before tests | 1200   | 1500   | 
| During tests | 6748   | 7398   |
```
These are the results of free and cached memory:
```
| OS            | RHEL 8 | RHEL 9 |
| free memory   | 149   | 140    |
| Cached memory | 683   | 436   | 
```

Note these are results with 3.33.1, It's possible that the 3.33 have also bigger memory footprint then 3.27. I'll check it and see, but I believe that this test was on edge of OOM even with 3.27.

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/sql-db/sql-app/src/test/java/io/quarkus/ts/sqldb/sqlapp/DevModeComposeLocationsIT.java
+++ b/sql-db/sql-app/src/test/java/io/quarkus/ts/sqldb/sqlapp/DevModeComposeLocationsIT.java
@@ -1,5 +1,7 @@
 package io.quarkus.ts.sqldb.sqlapp;
 
+import java.util.List;
+
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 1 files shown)


jedla97:
Hmm it seems that it's using the public bits instead of build bits. Probably because the https://github.com/quarkus-qe/quarkus-test-suite/blob/main/pom.xml#L169 is forcing the new bits every day. It's probably because of `updatePolicy` defaults to daily so it's probably forcing this update.

rsvoboda:
So the settings need something like https://github.com/quarkus-qe/quarkus-test-framework/pull/1886 ?

fedinskiy:
In case you need it, here is a PR: https://github.com/quarkus-qe/quarkus-test-suite/pull/2978

rsvoboda (approved):

jedla97:
CI green, merging

PR #2978, opened 2026-04-24
Build and deploy Quarkus platform as a part of daily CI

fedinskiy:
### Summary

Unlike upstream platform build, ours use Quarkus 999-SNAPSHOT

Similar to https://github.com/quarkus-qe/quarkus-test-framework/pull/1886

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/build-time-analytics/src/test/java/io/quarkus/ts/buildtimeanalytics/utils/AnalyticsUtils.java
+++ b/build-time-analytics/src/test/java/io/quarkus/ts/buildtimeanalytics/utils/AnalyticsUtils.java
@@ -6,10 +6,8 @@ import java.util.regex.Pattern;
 import io.quarkus.test.services.quarkus.model.QuarkusProperties;
 
 public class AnalyticsUtils {
-    private static final String QUARKUS_SNAPSHOT_VERSION = "3.999-SNAPSHOT";
-    // Quarkus reports extensions' versions in the analytics payload as full artifact version.
-    // So for 999-SNAPSHOT, it can look like e.g. 999-20230718.203351-1091
-    private static final String QUARKUS_EXTENSION_SNAPSHOT_VERSION_REGEX = "3\\.999-.*";
+    private static final String QUARKUS_SNAPSHOT_SUFFIX = "-SNAPSHOT";
+    private static final String QUARKUS_EXTENSION_SNAPSHOT_VERSION_REGEX = ".*999-SNAPSHOT";
     // RHBQ artifacts may differ in the number suffix from a platform version.
     // E.g.: 3.8.0.redhat-00002 vs. 3.8.5.redhat-00003
     private static final String RHBQ_VERSION_REGEX_FORMAT = "%s\\.redhat-\\d{5}";
@@ -108,7 +106,7 @@ public class AnalyticsUtils {
         if (rhbqMatcher.matches()) {
             return String.format(RHBQ_VERSION_REGEX_FORMAT, rhbqMatcher.group(1));
         }
-        if (version.equalsIgnoreCase(QUARKUS_SNAPSHOT_VERSION)) {
+        if (version.endsWith(QUARKUS_SNAPSHOT_SUFFIX)) {
             return QUARKUS_EXTENSION_SNAPSHOT_VERSION_REGEX;
         }
         return version;
--- a/pom.xml
+++ b/pom.xml
@@ -17,8 +17,8 @@
         <surefire-plugin.version>3.5.6</surefire-plugin.version>
         <failsafe-plugin.version>3.5.6</failsafe-plugin.version>
         <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
-        <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
-        <quarkus.platform.version>3.999-SNAPSHOT</quarkus.platform.version>
+        <quarkus.platform.group-id>io.quarkus.platform</quarkus.platform.group-id>
+        <quarkus.platform.version>999-core-3.x-SNAPSHOT</quarkus.platform.version>
         <quarkus.ide.version>3.30.0</quarkus.ide.version>
         <quarkus.qe.framework.version>999-SNAPSHOT</quarkus.qe.framework.version>
         <quarkus-qpid-jms.version>2.12.0</quarkus-qpid-jms.version>
@@ -180,12 +180,12 @@
         <pluginRepository>
             <id>quarkus-snapshots-plugin-repository</id>
             <url>https://central.sonatype.com/repository/maven-snapshots/</url>
-            <releases>
-                <enabled>false</enabled>
-            </releases>
             <snapshots>
                 <enabled>true</enabled>
             </snapshots>
+            <releases>
+                <enabled>false</enabled>
+            </releases>
         </pluginRepository>
     </pluginRepositories>
     <build>
@@ -324,6 +324,8 @@
                                 <keycloak.image>quay.io/keycloak/keycloak:26.6</keycloak.image>
                                 <bouncycastle.bctls-fips.version>1.0.12.2</bouncycastle.bctls-fips.version>
                                 <rhbk.image>${rhbk.image}</rhbk.image>
+                                <quarkus.platform.group-id>${quarkus.platform.group-id}</quarkus.platform.group-id>
+                                <quarkus.platform.version>${quarkus.platform.version}</quarkus.platform.version>
                             </systemPropertyVariables>
                         </configuration>
                     </execution>
@@ -433,6 +435,13 @@
     </build>
     <profiles>
         <profile>
+            <id>core-only</id>
+            <properties>
+                <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
+                <quarkus.platform.version>3.999-SNAPSHOT</quarkus.platform.version>
+            </properties>
+        </profile>
+        <profile>
             <id>env-info</id>
             <build>
                 <defaultGoal>clean compile</defaultGoal>
@@ -451,11 +460,8 @@
             </activation>
             <modules>
                 <module>env-info</module>
-                <!-- TODO: re-enable once we have aligned builds for LC4J/MCP server extensions
-                     See also https://github.com/quarkus-qe/quarkus-test-suite/pull/2978
                 <module>ai/langchain4j</module>
                 <module>ai/mcp</module>
-                -->
             </modules>
         </profile>
         <profile>
--- a/ai/langchain4j/pom.xml
+++ b/ai/langchain4j/pom.xml
@@ -15,7 +15,6 @@
         <quarkus.build.skip>true</quarkus.build.skip>
         <skipTests>true</skipTests>
         <skipITs>true</skipITs>
-        <quarkus.platform.group-id>io.quarkus.platform</quarkus.platform.group-id>
     </properties>
     <dependencyManagement>
         <dependencies>
@@ -75,9 +74,6 @@
                                     <quarkus.s2i.base-native-image>${quarkus.s2i.base-native-image}</quarkus.s2i.base-native-image>
                                     <!-- Services used in test suite -->
                                     <pgvector.image>docker.io/pgvector/pgvector:pg18</pgvector.image>
-                                    <!-- Passing values to cloned applications -->
-                                    <!-- Langchain4j extensions are not in core, so use platform prefix by default -->
-                                    <quarkus.platform.group-id>io.quarkus.platform</quarkus.platform.group-id>
                                 </systemPropertyVariables>
                             </configuration>
                         </execution>
--- a/ai/mcp/pom.xml
+++ b/ai/mcp/pom.xml
@@ -13,7 +13,6 @@
     <properties>
         <!-- The line below is required because we have empty src/main: https://github.com/quarkusio/quarkus/issues/33061 -->
         <quarkus.build.skip>true</quarkus.build.skip>
-        <quarkus.platform.group-id>io.quarkus.platform</quarkus.platform.group-id>
     </properties>
     <dependencyManagement>
         <dependencies>
@@ -82,9 +81,6 @@
                                     <!-- S2i configuration for OpenShift -->
                                     <quarkus.s2i.base-jvm-image>${quarkus.s2i.base-jvm-image}</quarkus.s2i.base-jvm-image>
                                     <quarkus.s2i.base-native-image>${quarkus.s2i.base-native-image}</quarkus.s2i.base-native-image>
-                                    <!-- Passing values to cloned applications -->
-                                    <!-- Langchain4j extensions are not in core, so use platform prefix by default -->
-                                    <quarkus.platform.group-id>${quarkus.platform.group-id}</quarkus.platform.group-id>
                                 </systemPropertyVariables>
                             </configuration>
                         </execution>
--- a/quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/QuarkusCliCreateJvmApplicationIT.java
+++ b/quarkus-cli/src/test/java/io/quarkus/ts/quarkus/cli/QuarkusCliCreateJvmApplicationIT.java
@@ -36,6 +36,7 @@ import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.OS;
 
+import io.quarkus.builder.Version;
 import io.quarkus.logging.Log;
 import io.quarkus.test.bootstrap.QuarkusCliClient;
 import io.quarkus.test.bootstrap.QuarkusCliRestService;
@@ -258,8 +259,9 @@ public class QuarkusCliCreateJvmApplicationIT {
         // The snapshot not providing platform-bom so it use the latest stream in registry
         // otherwise when the quarkus version is defined (3.27.0) or if it's snapshot of some quarkus stream (3.27.999-SNAPSHOT)
         // it will set the correct stream to use
-        String quarkusStream = QuarkusProperties.getVersion().equals("3.999-SNAPSHOT") ? null
-                : QuarkusProperties.getVersion().replaceAll("^(\\d+\\.\\d+).*", "$1");
+        String version = Version.getVersion(); // We are interested in core version only
+        String quarkusStream = version.equals("3.999-SNAPSHOT") ? null
+                : version.replaceAll("^(\\d+\\.\\d+).*", "$1");
         // This can't use `--platform-bom` as it contain quarkiverse extension
         QuarkusCliRestService app = cliClient.createApplication("app",
                 defaults().withPlatformBom(null).withStream(quarkusStream).withExtensions(camelExtension,


jedla97:
My thoughts on the disable the update of snapshot

The FW is also in that snapshot repository, so when developing locally, we would need to build the FW (or delete it from local repo). I don't know workflow of each of us, and how often we build the FW locally (I mostly don't do that if I don't fixing something related to FW)

Also on Jenkins it's possible that it will still try to download the upstream snapshot instead of the one we build (I don't know which have higher priority, the repo defined or settings defined)

I'm not opposed to this but I think it's good topic for Monday meeting to get other people thoughts.

fedinskiy:
run tests

pom.xml
```diff
@@ -179,12 +180,13 @@
         <pluginRepository>
             <id>quarkus-snapshots-plugin-repository</id>
             <url>https://central.sonatype.com/repository/maven-snapshots/</url>
-            <releases>
```
mjurc, line 183: To be honest, I don't understand this removal. We shouldn't be trying to pull releases from snapshot repository, that will just slow down the build.

pom.xml
```diff
@@ -169,6 +169,7 @@
             <url>https://central.sonatype.com/repository/maven-snapshots/</url>
             <snapshots>
                 <enabled>true</enabled>
+                <updatePolicy>never</updatePolicy>
```
mjurc: Like I said in other PR, maybe it is worth considering just scratching Sonatype Snapshot repository altogether if we cannot depend on its contents, otherwise what is the point of defining it here if we build everything that we want from it (Quarkus) in our actions?

.github/workflows/daily.yaml
```diff
@@ -21,11 +21,19 @@ jobs:
...
       - name: Build Quarkus 999-SNAPSHOT
         run: |
-          mkdir maven-repository
-          LOCAL_REPO="$(pwd)/maven-repository"
-          git clone https://github.com/quarkusio/quarkus.git && cd quarkus && ./mvnw -B --no-transfer-progress -s .github/mvn-settings.xml clean install -Dquickly -Dno-test-modules -Prelocations -Dmaven.repo.local=$LOCAL_REPO
+          git clone https://github.com/quarkusio/quarkus.git && cd quarkus
+          ./mvnw -B --no-transfer-progress -s .github/mvn-settings.xml clean install -Dquickly -Dno-test-modules -Prelocations -Dmaven.repo.local=${{ env.LOCAL_REPO }}
+      - name: Build Quarkus Platform 999-SNAPSHOT
+        run: |
+          git clone https://github.com/quarkusio/quarkus-platform.git && cd quarkus-platform
```
mjurc, line 53: I have just one thing here - like I mentioned in the request for enhancement, I think we should rename the version of io.quarkus.platform:platform bom to something else than 999-SNAPSHOT because even if people pull and build the default version of platform themselves, they're gonna get something based on released Quarkus, and we're doing something else here.

I'd propose going with quarkus-main-999-SNAPSHOT or something like that, just like in my RFE

mjurc (changes requested):
A few notes, also I'll not be available to review this so I'm reassigning this to @rsvoboda

pom.xml
fedinskiy: This is not a removal, I just reordered lines (so both repos have sections in the same order)

.github/workflows/daily.yaml
fedinskiy: @mjurc could you elaborate your position?

From the comment[1] to the previous PR I got an impression, that you are strictly opposed to adding new versions and that is why I implemented Jakubs suggestion to disable snapshots. If you are fine with building new versions, when should not we also use them in Quarkus Framework?

[1] https://github.com/quarkus-qe/quarkus-test-framework/pull/1886#issuecomment-4312521525

rsvoboda (changes requested):
different platform version - I agree with Michal, this build differs from what's deployed in https://github.com/quarkusio/quarkus-platform/blob/main/.github/workflows/deploy-snapshots.yml
We already had a case when people spent time investigating a wrong setup (https://github.com/quarkiverse/quarkus-langchain4j/issues/2340#issuecomment-4267252665) 

`<updatePolicy>never</updatePolicy>` - this breaks usability for end users. Imagine having an old 999-SNAPSHOT version of the framework in the local repository.

Design of AI module for snapshot builds - you expect to have the same core and platform version. Based on the details above, this is not an ideal approach.

To summarize, this PR proposes changes to (finally) run Quarkus main and Quarkus platform combination for AI modules, but introduces regression in user experience and overloads the platform 999-SNAPSHOT version with a different dependency chain.

This needs a new round of thinking. And please discuss the implementation options before investing in a new round of changes.

fedinskiy:
run tests

fedinskiy:
run tests

fedinskiy:
CI failures:
- GH failures are expected, since daily build of platform is a part of the PR
- OCP JVM failure seemed to be temporary (I was able to reproduce them today at 13:30, but not at 14:30). Will retrigger the check
- OCP Native failures are due to https://github.com/quarkus-qe/quarkus-test-suite/pull/3011, should be solved by rebasing. Will retrigger the check

.github/workflows/daily.yaml
mjurc: This has been explained in messages so I'll resolve it

mjurc:
@fedinskiy can you rebase so the CI can run?

pom.xml
```diff
@@ -16,8 +16,8 @@
         <surefire-plugin.version>3.5.5</surefire-plugin.version>
         <failsafe-plugin.version>3.5.5</failsafe-plugin.version>
         <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
-        <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
-        <quarkus.platform.version>999-SNAPSHOT</quarkus.platform.version>
+        <quarkus.platform.group-id>io.quarkus.platform</quarkus.platform.group-id>
+        <quarkus.platform.version>999-core-main-SNAPSHOT</quarkus.platform.version>
```
jedla97: I'm thinking out loud but wouldn't be better to have 999-SNAPSHOT as default version and for CI to have profile which would overwrite this. Note I'm not sure that it would overwrite the value in AI tests as it was problem from what I remember.

btw this is mainly my laziness speaking as I will need to remember overwrite the value each time when running the tests locally.

.github/workflows/daily.yaml
```diff
@@ -21,11 +21,23 @@ jobs:
...
+          mkdir -p maven-repository
+          echo "LOCAL_REPO=$(pwd)/maven-repository" >> $GITHUB_ENV
       - name: Build Quarkus 999-SNAPSHOT
         run: |
-          mkdir maven-repository
-          LOCAL_REPO="$(pwd)/maven-repository"
-          git clone https://github.com/quarkusio/quarkus.git && cd quarkus && ./mvnw -B --no-transfer-progress -s .github/mvn-settings.xml clean install -Dquickly -Dno-test-modules -Prelocations -Dmaven.repo.local=$LOCAL_REPO
+          git clone https://github.com/quarkusio/quarkus.git && cd quarkus
+          ./mvnw -B --no-transfer-progress -s .github/mvn-settings.xml clean install -Dquickly -Dno-test-modules -Prelocations -Dmaven.repo.local=${{ env.LOCAL_REPO }}
+      - name: Build Quarkus Platform 999-SNAPSHOT with Quarkus core 999-SNAPSHOT
```
jedla97:
```suggestion
      - name: Build Quarkus Platform 999-core-main-SNAPSHOT with Quarkus core 999-SNAPSHOT
```

pom.xml
mjurc: I endorse this idea.

fedinskiy: > I will need to remember overwrite the value each time when running the tests locally.

You do not need to do this, the built platform will be deployed to our nexus every day

jedla97: So now it will be mandatory to either overwrite it (when looking for commit which caused failure) or have access to our nexus (I don't know if on ibm side it will be available as it is now + for RH and IBM side connection to vpn will be mandatory).

\+ note it's possible that the our nexus will have slightly different version of Quarkus then gh action as they are build at different time.

mjurc: The goal of this pull request is to have ability to run tests with platform based on Quarkus main. There's no such builds available anywhere, sadly, but this is something we need in our actions due to having have to test platform participants that are not in Quarkus core.

The timing issue is there already - 999-SNAPSHOT is built at different times between the lab and GH actions too. The burden here is that the platform build that we need is not available anywhere, so people will need to either produce it or depend on one of our Maven repositories with their builds. We have options for all the internal lab environments.

So yeah, to be able to run the tests with versions of platform and Quarkus from main, the people running locally will have to supply both the 999-SNAPSHOT build of Quarkus main and the platform based on it. I would like the profile idea if there was any solution like that if it worked, but... It's hard to say how it would work ahead of time without trying it

jedla97:
I have one question which is kinda related. If we build our platform to tests with snapshot, the platform still setting the `quarkus-langchain4j` and `quarkus-mcp` to released version which depends on released Quarkus so we will tests Quarkus snapshot with released extension which is based on some released Quarkus (currently it seems on LTS) + this extension not contains latest changes.

So if something change in Quarkus core which broke the compatibility, we will need to wait until the new version of langchain4j is fixed and released.

So my question is shouldn't we build also the `quarkus-langchain4j` and `quarkus-mcp` and use them in platform as we want to tests for future releases? Maybe this is not question for now but for future iteration.

mjurc:
@jedla97 Right you are. I actually wonder - do the MCP/LC4J tests need the Quarkus platform? Wouldn't it be enough just to build the MCP/LC4J platform participants? But I'll leave the analysis answering those questions to @fedinskiy since he's on this effort and he knows what the tests need :)

pom.xml
fedinskiy: Implemented (in a separate commit, so it will be easier to revert/ change it later)

fedinskiy:
Regarding AI _tests_ modules — I do not think, that building them is a good idea:
- They have their own CI, so they are able to detect errors and incompatibilities early
- Their latest released versions are added to the platform regularly (there is even an automatic reminder about that)
- We care about things which are already in the platform, since they are to be productised. We do not need to monitor all the changes in every component, otherwise we will have to build such things as Hibernate and Vertx.

fedinskiy:
run tests

mjurc:
@fedinskiy The question posed was not about building AI tests. One question was how this change ensures that the LC4J/MCP server platform participant boms that the tests use ( https://github.com/quarkus-qe/quarkus-test-suite/blob/main/ai/mcp/pom.xml#L23 + https://github.com/quarkus-qe/quarkus-test-suite/blob/main/ai/langchain4j/pom.xml#L25 ), because the platform by default has version of lc4j/mcp server extensions that have been released. Is this correct?

My other question was whether the new tests in langchain4j/mcp server module do really need Quarkus platform bom for upstream testing if they already depend on lc4j/mcp server platform boms. Instinctively here I'd say the answer is yes, cos that's where the lc4j/mcp server platform boms come from, but I'm just making sure.

> We care about things which are already in the platform, since they are to be productised. We do not need to monitor all the changes in every component, otherwise we will have to build such things as Hibernate and Vertx.

Hibernate and Vertx are part of Quarkus core. We override version of Quarkus core for building the platform in this PR to one built from main.  But this does not happen for lc4j/mcp server core. I'm not saying this is how it should be done (because I do not know the correct answer), but if they're not overriden in some way, both the lc4j and mcp server platforms will manage released versions of their extensions:

- https://github.com/quarkusio/quarkus-platform/blob/main/generated-platform-project/quarkus-mcp-server/bom/pom.xml#L65 
- https://github.com/quarkusio/quarkus-platform/blob/main/generated-platform-project/quarkus-langchain4j/bom/pom.xml#L65

fedinskiy:
@mjurc in my previous message I meant building AI modules/platforms. Sorry for misunderstanding.

Regarding your questions:
- Quarkus Platform uses released versions of mcp and langchain4j (I will call them AI for brevity) quarkiverse modules, but wrap them inside boms, which use the same version as the rest of the platform (eg https://github.com/quarkusio/quarkus-platform/blob/main/generated-platform-project/quarkus-mcp-server/bom/pom.xml)
- AI boms and AI quarkiverse modules manage their own versions of dependencies, and I do no think this is a problem
- AI quarkiverse modules also depend on core Quarkus BOM (eg https://github.com/quarkiverse/quarkus-langchain4j/blob/main/pom.xml#L58-L63) of some earlier released version. This may lead to CI failures, when there are incompatibilities between quarkus core main branch and the released version. I suppose, this is fine, and it is intended purpose of this test suit to detect issues like those.
- We could depend on AI BOMs instead of platform bom. We could build AI quarkiverse modules in addition to Quarkus core. I see no reason why we would want to do any of these changes, especially given that they will complicate CI and testing.

pom.xml
fedinskiy: This change requires explicit platform version in the CI, which leads to some unexpected consequences. Now we also require this fix: https://github.com/quarkus-qe/quarkus-test-framework/pull/1915

mjurc: I've reviewed https://github.com/quarkus-qe/quarkus-test-framework/pull/1915

I also am not sure if https://github.com/quarkus-qe/quarkus-test-suite/pull/2978/changes/3bf28c0a75433a45f83bd8e0be7bc245b9c09683 addresses what @jedla97 mentioned. This just changes the default platform version, but not groupId, so it still uses `io.quarkus.quarkus-platform:quarkus-bom:999-SNAPSHOT` (which doesn't exist) rather than `io.quarkus:quarkus-bom:999-SNAPSHOT` by default

mjurc (commented):
Let's go point by point:

1. > Quarkus Platform uses released versions of mcp and langchain4j (I will call them AI for brevity) quarkiverse modules, but wrap them inside boms, which use the same version as the rest of the platform (eg https://github.com/quarkusio/quarkus-platform/blob/main/generated-platform-project/quarkus-mcp-server/bom/pom.xml)

Right, that's the point. Platform build (at least on release) runs ecosystem CI verifying there's no incompatibilities between platform members, generates all platform participant boms with the same version, but the platform boms manage a  list of extension modules at a version for the same platform version. On product, some dependency management overrides are added, but this does not apply for scenario on main.

The point I was trying to make is I am not sure if this is what we want to be doing. The test suite runs on nightly with Quarkus built from Quarkus main. Is the goal to test whether released versions of LC4J modules work with Quarkus main? I guess that has some use, but it also implies a missing spot in coverage. The other issue is I am not entirely convinced that just having a platform build would solve the issue we have with CI now, e.g. https://github.com/quarkus-qe/quarkus-test-suite/actions/runs/27223398358/job/80405963863?pr=3027 because I don't really see the mechanism by which this would happen. But I could be msising something :)

2. > AI boms and AI quarkiverse modules manage their own versions of dependencies, and I do no think this is a problem

Not a problem. All the extensions have to depend on Quarkus core

3. > AI quarkiverse modules also depend on core Quarkus BOM (eg https://github.com/quarkiverse/quarkus-langchain4j/blob/main/pom.xml#L58-L63) of some earlier released version. This may lead to CI failures, when there are incompatibilities between quarkus core main branch and the released version. I suppose, this is fine, and it is intended purpose of this test suit to detect issues like those.

Not a problem. All the extensions have to depend on Quarkus core but the side effect is they have to be rebuilt with new Quarkus core versions

4. We could depend on AI BOMs instead of platform bom. We could build AI quarkiverse modules in addition to Quarkus core. I see no reason why we would want to do any of these changes, especially given that they will complicate CI and testing.

Again, this really depends on what we want to achieve with lc4j and mcp modules on main CI. There is some value in testing whether Quarkus main works with latest released LC4J or MCP server extensions, but the goal of this CI is to detect changes early so we don't have to report bugs and rebuild dependencies at the time of product build or late in LTS schedule which would make testing and rebuilding quite stressful. If we test just Quarkus main against latest released version of MCP server and LC4J extensions, we just simply don't have an idea of what's happening in development of those extensions

Also, this will need to be rebased once all the other answers are figured out ;)

pom.xml
fedinskiy: Now I am lost.

- Jakub explicitly asked about keeping the original version ("wouldn't be better to have 999-SNAPSHOT").
- Changing the groupId, and switching from `io.quarkus` to `io.quarkus.platform` for the whole TS is the main purpose of this PR, one which we discussed earlier and agreed on.

Given the above, I concluded you want to switch to io.quarkus.platform:quarkus-bom:999-SNAPSHOT as a default BOM. Do you prefer some other option?

jedla97:
Hi, as I mention previously, would you consider some profile which would keep the `io.quarkus:quarkus-bom:999-SNAPSHOT` so when developing or debugging tests we don't need to build custom platform as for most of the use cases the platform is not needed.

Or 2nd option would be keep the pom.xml values as they were and have profile for CI which would overwrite it to platform.

I don't have strong opinion which to use, but it would be easier then overwriting 2 values each time

fedinskiy:
run tests

fedinskiy:
run tests

fedinskiy:
The following CI failures are expected and not related to the content of the PR:
- Linux JVM 21( cache-modules,spring-modules, etc )
- Linux  JVM 25 ( cache-modules,spring-modules, etc )
- Windows JVM 21 ( cache-modules,spring-modules, etc )
- Openshift JVM
I also expect a failure in Linux Native (21, -P security-modules,spring-modules )

All of them are due to compilation failure in spring-data module and should be fixed, when https://github.com/quarkusio/quarkus/pull/55227 is backported to 3.x.

mjurc (approved):
I think we can go ahead with this, @jedla97 wdyt?

build-time-analytics/src/test/java/io/quarkus/ts/buildtimeanalytics/utils/AnalyticsUtils.java
```diff
@@ -6,10 +6,8 @@
 import io.quarkus.test.services.quarkus.model.QuarkusProperties;
 
 public class AnalyticsUtils {
-    private static final String QUARKUS_SNAPSHOT_VERSION = "3.999-SNAPSHOT";
-    // Quarkus reports extensions' versions in the analytics payload as full artifact version.
-    // So for 999-SNAPSHOT, it can look like e.g. 999-20230718.203351-1091
-    private static final String QUARKUS_EXTENSION_SNAPSHOT_VERSION_REGEX = "3\\.999-.*";
+    private static final String QUARKUS_SNAPSHOT_SUFFIX = "-SNAPSHOT";
+    private static final String QUARKUS_EXTENSION_SNAPSHOT_VERSION_REGEX = ".*999-SNAPSHOT";
```
jedla97, line 10: I didn't look closely where this is used but should it also end with `.*`? The code currently contains it so I think maybe it should remain.

jedla97 (approved):
I think this looks fine I have just one question, but we probably can go with this and merge this.

build-time-analytics/src/test/java/io/quarkus/ts/buildtimeanalytics/utils/AnalyticsUtils.java
fedinskiy: According to the comment above (which I also removed) Quarkus CLI used to use build version of jar file as a CLI version for SNAPSHOT. I checked this behavior, and it doesn't happen anymore.

jedla97: Thanks for checking!

fedinskiy:
@rsvoboda github rules require your approval before merging :)

PR #2993, opened 2026-05-14
Disable check of version in update test for testcontainers

jedla97:
### Summary

I disabling the version check in `QuarkusCli333UpdatesIT#testContainers` because of https://redhat.atlassian.net/browse/QUARKUS-7802. 

Personally I think maybe the version should not be there at all as the version is set to latest available, not the platform specific, which is managed by platform. I'm adding it first in main as it's possible that this won't be fixed fast.

For end user the impact is also minimal as the plaform bom manage the version, so it can be problematic only for users which manage theirs version for some specific reason.  

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:


fedinskiy (approved):
Thank you!

PR #2996, opened 2026-05-18
Fix flaky OpenShiftToolsIT.askAI test

gastaldi:
## Summary
- Use an explicit prompt that names the `readFileContent` tool and provides the full `playground/longword.txt` path, reducing ambiguity for the LLM
- Increase the `waitForAnswers` timeout from ~5s to ~20s to accommodate tool invocation round-trips on OpenShift
- Needed for QQE-2544: the LLM sometimes responds with conversational filler instead of invoking the MCP tool
- Companion PR: https://github.com/quarkiverse/quarkus-langchain4j/pull/2469 (strengthens the upstream system prompt)
- Also get rid of the warning message:
```
15:02:31 18:02:31,648 INFO  mvn: [WARNING] [io.smallrye.config] SRCFG01008: The value false #otherwise we run into problems while parsing CLI logs on Windows has been converted by a Boolean Converter to "false"
```

## Test plan
- [x] Run `OpenShiftToolsIT` on OpenShift to verify the test passes consistently
- [x] Verify `ToolsIT` still passes locally

### DIFF
--- a/ai/langchain4j/src/test/java/io/quarkus/ts/langchain4j/AbstractToolsIT.java
+++ b/ai/langchain4j/src/test/java/io/quarkus/ts/langchain4j/AbstractToolsIT.java
@@ -64,7 +64,7 @@ public abstract class AbstractToolsIT {
         waitForAnswers(answers);
         Assertions.assertEquals("Hello, I am a filesystem robot, how can I help?", answers.get(0));
         answers.clear();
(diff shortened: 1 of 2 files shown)


fedinskiy (approved):
Thank you!

jcarranzan:
Adding triage label for 3.33

PR #3003, opened 2026-05-25
Retry askAI prompt to handle non-deterministic LLM tool invocation

gastaldi:
The LLM sometimes responds with conversational filler instead of invoking the MCP tool. Retry the prompt up to 3 times within the same WebSocket session before failing, which makes the test resilient to non-deterministic LLM behavior without requiring upstream changes.


fedinskiy (approved):

rsvoboda:
> without requiring upstream changes.

@gastaldi so there is possible fix on LangChain4J side?

gastaldi:
@rsvoboda yes, see https://github.com/quarkiverse/quarkus-langchain4j/pull/2477

Issue #118, opened 2021-07-23
Enhance readme to add more info about the framework and how to run native mode / against OpenShift, Kubernetes


rsvoboda:
Enhance readme of the TS to add more info about the framework and how to run native mode

https://github.com/quarkus-qe/quarkus-test-framework#readme
https://github.com/quarkus-qe/quarkus-test-framework#native

The details need to be at the top

Current one-liner (https://github.com/quarkus-qe/quarkus-test-suite#test-framework) is not sufficiently visible and strong on the message.

Add details how to run against OpenShift, Kubernetes.

rsvoboda:
One thing I miss even in the framework readme is how to tune the config for native execution

 - How to use GraalVM instead of Mandrel
 - How to use native-image on my local path, not the docker on
   - I'm on macOS and I can't execute Linux binaries

rsvoboda:
@pjgg / @kshpak I've hit this again. we need simple how to run stuff directly in the readme

fedinskiy:
@rsvoboda did https://github.com/quarkus-qe/quarkus-test-suite/pull/389 solve your problems? If not, then which topics should be covered?

rsvoboda:
Yes, I'm happy with https://github.com/quarkus-qe/quarkus-test-suite/pull/389 changes, closing.

PR #3006, opened 2026-05-25
Drop -Dplatform-deps from langchain4j sample builds to fix Agroal BOM conflict

gastaldi:
## Summary

- Fixes `OpenShiftMoviesIT` native build failure (failing since build #118, 5 consecutive failures)
- The `quarkus-langchain4j-bom` from the Quarkus Platform is aligned to Quarkus 3.27.2, which manages `agroal-api:2.8`. When testing against Quarkus `999-SNAPSHOT` (which uses `agroal-api:3.1.2`), the platform BOM causes the older Agroal version to end up on the classpath. This breaks native image compilation because `DataSources.networkTimeout(Duration)` only exists in `agroal-api:3.1.2`
- By dropping `-Dplatform-deps`, the sample apps use the `default-project-deps` profile which imports **only** `quarkus-bom` (no conflicting langchain4j platform BOM) and declares langchain4j extension versions explicitly

## Test plan

- [ ] Verify `OpenShiftMoviesIT` passes on OpenShift native build with Quarkus `999-SNAPSHOT`
- [ ] Verify `OpenShiftChatbotIT` and `OpenShiftToolsIT` still pass
- [ ] Verify local `MoviesIT`, `ChatbotIT`, `ToolsIT` still pass

### DIFF
--- a/ai/langchain4j/src/test/java/io/quarkus/ts/langchain4j/auxiliary/CommonTools.java
+++ b/ai/langchain4j/src/test/java/io/quarkus/ts/langchain4j/auxiliary/CommonTools.java
@@ -3,7 +3,7 @@ package io.quarkus.ts.langchain4j.auxiliary;
 import org.eclipse.microprofile.config.ConfigProvider;
 
 public final class CommonTools {
-    public static final String DEFAULT_ARGS = "--no-transfer-progress -DskipTests=true -DskipITs=true -Dquarkus.platform.version=${QUARKUS_PLATFORM_VERSION} -Dquarkus.platform.group-id=${QUARKUS_PLATFORM_GROUP-ID}";
+    public static final String DEFAULT_ARGS = "--no-transfer-progress -DskipTests=true -DskipITs=true -Dsamples -Dquarkus.platform.version=${QUARKUS_PLATFORM_VERSION} -Dquarkus.platform.group-id=${QUARKUS_PLATFORM_GROUP-ID}";
     public static final String SAMPLE_BRANCH = "1.7";
 
     public static String getKey() {
--- a/ai/langchain4j/src/test/java/io/quarkus/ts/langchain4j/ChatbotIT.java
+++ b/ai/langchain4j/src/test/java/io/quarkus/ts/langchain4j/ChatbotIT.java
@@ -16,8 +16,7 @@ public class ChatbotIT extends AbstractChatbotIT {
     @Container(image = "${redis.image}", port = 6379, expectedLog = "Ready to accept connections")
     static DefaultService redis = new DefaultService().withProperty("ALLOW_EMPTY_PASSWORD", "YES");
 
-    @GitRepositoryQuarkusApplication(repo = "https://github.com/quarkiverse/quarkus-langchain4j.git", branch = SAMPLE_BRANCH, contextDir = "samples/chatbot", mavenArgs = DEFAULT_ARGS
-            + " -Dsamples -Dplatform-deps")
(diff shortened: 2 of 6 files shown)


gastaldi:
This should fix the build failure in https://jenkins-csb-quarkusqe-main.dno.corp.redhat.com/view/periodic-builds/job/quarkus-main-rhel8-jdk17-openshift-ts-native-ocp-4-eus-weekly/SCENARIO=ai,jdk=openjdk-17,label=RHEL8%20&&%20xlarge/122/consoleFull

rsvoboda:
I don't think this is the right approach, we need "-Dplatform-deps" for testing with released version.

For testing with main @fedinskiy is looking into customized platform main build with core main.
So we would need LangChain4J Quarkiverse extension main too? I will let Fedor comment on this.

> quarkus-langchain4j-bom from the Quarkus Platform is aligned to Quarkus 3.27.2

I'm wondering why is that, there is already 3.33.1

gastaldi:
Quarkus main uses Agroal 3.1.2 which introduces a new method. Any attempt to build with Quarkus 999-SNAPSHOT without quarkus-bom 999-SNAPSHOT is wrong.

fedinskiy:
The main purpose of this test suite is to run with productised platform, so dropping the `platform-deps` profile is not an option, unfortunately.

I am making progress with switching our whole TS to 999 platform a default, so, hopefully, this issue will be solved soon.

Regarding building the ql4j main: I do not think this is needed, new langchain4j (and mcp) versions are usually added to quarkus platform soon after release.

ai/langchain4j/src/test/java/io/quarkus/ts/langchain4j/auxiliary/CommonTools.java
```diff
@@ -3,7 +3,7 @@
 import org.eclipse.microprofile.config.ConfigProvider;
 
 public final class CommonTools {
-    public static final String DEFAULT_ARGS = "--no-transfer-progress -DskipTests=true -DskipITs=true -Dquarkus.platform.version=${QUARKUS_PLATFORM_VERSION} -Dquarkus.platform.group-id=${QUARKUS_PLATFORM_GROUP-ID}";
+    public static final String DEFAULT_ARGS = "--no-transfer-progress -DskipTests=true -DskipITs=true -Dsamples -Dquarkus.platform.version=${QUARKUS_PLATFORM_VERSION} -Dquarkus.platform.group-id=${QUARKUS_PLATFORM_GROUP-ID}";
     public static final String SAMPLE_BRANCH = "1.7";
```
jedla97, line 7: Afaik this is just thought but I think we should test with main branch not 1.7 as if I remember correctly the 1.7 is for 3.33 Quarkus. Would be just switching `SAMPLE_BRANCH` to main enough?

gastaldi: Not really AFAIR, the problem is that the quarkus-bom 999-SNAPSHOT declared does not take precedence while resolving the transitive dependencies.

gastaldi:
> I am making progress with switching our whole TS to 999 platform a default, so, hopefully, this issue will be solved soon.

@fedinskiy Thanks, is there a JIRA with your work I can use to link to the existing Jenkins investigation issue for reference?

gastaldi:
Closing PR in favor of @fedinskiy's work in switching the whole TS to 999 platform a default

fedinskiy:
@gastaldi the Jira is this one: https://redhat.atlassian.net/browse/QQE-2465

PR #3032, opened 2026-06-09
Add -nsu to CI and daily workflows to skip remote snapshot resolution

gastaldi:
## Summary
- Add `-nsu` (`--no-snapshot-updates`) to all `mvn` test commands in both `daily.yaml` and `ci.yml` workflows
- Both workflows restore a pre-built Quarkus `999-SNAPSHOT` from cache into the local Maven repository. Without `-nsu`, Maven still checks the Sonatype Central SNAPSHOT repository for updates, which is unnecessary and can pull unwanted artifacts.

## Test plan
- [ ] Verify CI and daily builds still resolve all Quarkus artifacts from the local repository
- [ ] Confirm no remote snapshot resolution attempts in Maven logs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,7 +131,7 @@ jobs:
           wait-before-restore: false
       - name: Build with Maven
         run: |
-          mvn -V -B --no-transfer-progress -s .github/mvn-settings.xml verify -Dall-modules -Dvalidate-format -DskipTests -DskipITs -Dquarkus.container-image.build=false -Dquarkus.container-image.push=false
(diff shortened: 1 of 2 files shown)


rsvoboda (commented):
I think `-nsu` should be added also into https://github.com/quarkus-qe/quarkus-test-suite/blob/main/.github/workflows/ci.yml

jedla97 (approved):
I think this make sense
* Quarkus which we build won't be downloaded
* FW which is stored in sonatype will be downloaded as it's not present
* ATM Quarkus platform will be downloaded as it's not build currently, but it will in few days hopefully

PR #3039, opened 2026-06-13
Adapt step-up auth test to upstream code flow redirect change

gastaldi:
Fixes https://redhat.atlassian.net/browse/QQE-2617

## Summary

- Adapt `testBrowserStepUpAuthenticationFlow` in `AbstractOidcStepUpAuthenticationIT` to match the new step-up authentication behavior introduced by [quarkusio/quarkus#54785](https://github.com/quarkusio/quarkus/pull/54785)

## What changed in Quarkus

Quarkus commit [`8a5058c6e8c`](https://github.com/quarkusio/quarkus/commit/8a5058c6e8c) ("Carry acr_values and max_age in the code flow authorization request") changed how step-up authentication works for web-app (authorization code flow) applications:

- **Before**: When a session did not have the required ACR level (e.g. user had "silver" but endpoint required "gold"), Quarkus returned **401 Unauthorized** with a `WWW-Authenticate` header containing `insufficient_user_authentication`
- **After**: Quarkus now **redirects the user (302)** to the OIDC provider's authorization endpoint with `acr_values` and `max_age` parameters in the request, allowing the provider to run a stronger authentication flow (step-up)

This aligns with [RFC 9470](https://www.rfc-editor.org/rfc/rfc9470) and enables seamless step-up authentication in the browser without requiring the user to manually re-authenticate.

## What changed in this PR

The test previously expected a `FailingHttpStatusCodeException` (triggered by a 401 response) when accessing a "gold" ACR endpoint with "silver" credentials:

```java
// Old: expected 401
Assertions.assertThrows(FailingHttpStatusCodeException.class,
    () -> webClient.getPage(...).getWebResponse().getContentAsString());
```

Now it verifies the redirect behavior instead:

```java
// New: expect 302 redirect to Keycloak with acr_values=gold
webClient.getOptions().setRedirectEnabled(false);
WebResponse stepUpResponse = webClient.loadWebResponse(new WebRequest(...));
Assertions.assertEquals(302, stepUpResponse.getStatusCode());
assertThat(location, containsString("acr_values=gold"));
webClient.getOptions().setRedirectEnabled(true);
```

This matches the pattern used in the upstream Quarkus test `CodeFlowStepUpAuthenticationTest.testStepUpFromExistingSession`.

## Test plan

- [x] Verify `OidcStepUpAuthenticationIT` passes locally in JVM mode
- [x] Verify `OpenShiftOidcStepUpAuthenticationIT` passes in OpenShift native mode

### DIFF
--- a/security/keycloak-oidc-client-extended/src/test/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/AbstractOidcStepUpAuthenticationIT.java
+++ b/security/keycloak-oidc-client-extended/src/test/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/AbstractOidcStepUpAuthenticationIT.java
@@ -10,6 +10,7 @@ import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.not;
 
 import java.io.IOException;
+import java.net.URI;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
@@ -19,10 +20,10 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import org.apache.http.HttpHeaders;
-import org.htmlunit.FailingHttpStatusCodeException;
 import org.htmlunit.SilentCssErrorHandler;
 import org.htmlunit.WebClient;
 import org.htmlunit.WebClientOptions;
+import org.htmlunit.WebRequest;
 import org.htmlunit.WebResponse;
 import org.htmlunit.html.HtmlForm;
 import org.htmlunit.html.HtmlPage;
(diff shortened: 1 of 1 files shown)


gastaldi:
The second commit refactors the assertion to test the **outcome** (silver credentials cannot access gold content) rather than the **HTTP mechanism** (401 vs 302 redirect).

This makes the test resilient to future upstream changes in how step-up authentication is enforced. For example, the original test broke when Quarkus switched from returning 401 to issuing a 302 redirect — but the *contract* never changed: a user without sufficient ACR still cannot access the protected content. By asserting `not(containsString("Single ACR gold validated"))` regardless of status code, the test survives either mechanism without modification.

security/keycloak-oidc-client-extended/src/test/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/AbstractOidcStepUpAuthenticationIT.java
```diff
@@ -135,10 +134,13 @@ public void testBrowserStepUpAuthenticationFlow() throws IOException {
...
-            // "gold" ACR needs step-up authentication, we expect 401
-            Assertions.assertThrows(FailingHttpStatusCodeException.class,
-                    () -> webClient.getPage(app.getURI(Protocol.HTTP).withPath("/step-up/single-acr-gold-web-app").toString())
-                            .getWebResponse().getContentAsString());
+            // "gold" ACR needs step-up authentication, silver credentials are not sufficient
+            webClient.getOptions().setThrowExceptionOnFailingStatusCode(false);
+            content = webClient.getPage(
+                    app.getURI(Protocol.HTTP).withPath("/step-up/single-acr-gold-web-app").toString())
+                    .getWebResponse().getContentAsString();
+            assertThat(content, not(containsString("Single ACR gold validated")));
```
jcarranzan: I'd prefer to verify the step-up redirect mechanism (for instance the status 302, also acr_values=gold is present), not just the absence of success content

jcarranzan:
I understand your reasoning, but in this case (https://github.com/quarkusio/quarkus/commit/8a5058c6e8c) the 302 redirect with acr_values is the feature introduced for [RFC](https://www.rfc-editor.org/rfc/rfc9470) as described your summary PR...

If the test passes when the feature is changed/broken will be a problem...

jcarranzan (changes requested):

gastaldi:
@jcarranzan Good point — you're right that the 302 redirect with `acr_values` is the feature itself (RFC 9470), not just an implementation detail. If we only asserted the outcome, the test would pass even if Quarkus reverted to returning 401, which would mask a regression.

I've updated the PR to keep the explicit redirect assertion (`assertEquals(302, ...)` + `containsString("acr_values=gold")`), so we verify both the mechanism (redirect with acr_values per RFC 9470) and the outcome (gold access after step-up re-authentication).

rsvoboda:
I removed myself from the review, I have no knowledge of this test coverage. Jose's review should be enough.

jcarranzan (approved):

PR #3046, opened 2026-06-17
Upgrade PostgreSQL 13 to 15 in OpenShift tests

gastaldi:
## Summary

Fixes OpenShift test failures in the `hibernate-modules` scenario caused by Quarkus now requiring PostgreSQL 14+ by default. Tests using PostgreSQL 13 crash-loop at startup with:

```
Persistence unit '<default>' was configured to run with a database version of at least '14.0.0',
but the actual version is '13.23.0'.
```

Observed in [build #164 (hibernate-modules)](https://jenkins-csb-quarkusqe-main.dno.corp.redhat.com/view/periodic-builds/job/quarkus-main-rhel8-jdk17-openshift-ts-jvm-ocp-4-stable-weekly/SCENARIO=hibernate-modules,jdk=openjdk-17,label=RHEL8%20&&%20large/).

### Changes

PostgreSQL 14 is not available on the Red Hat registry, so we upgrade to PostgreSQL 15 as the oldest supported version:

- **`postgresql.13.image`** property replaced with **`postgresql.15.image`** (`registry.redhat.io/rhel9/postgresql-15:latest`) in all 3 Maven profiles in `pom.xml`
- **`OpenShiftPostgresql13HibernateReactiveIT`** renamed to **`OpenShiftPostgresql15HibernateReactiveIT`**
- **`OpenShiftPostgresql13DatabaseIT`** renamed to **`OpenShiftPostgresql15DatabaseIT`**
- **`OpenShiftPostgresqlMultitenantHibernateSearchIT`** switched from `${postgresql.13.image}` to `${postgresql.latest.image}` (PostgreSQL 16), matching its non-OpenShift sibling `PostgresqlMultitenantHibernateSearchIT`

## Test plan

- [ ] Verify `OpenShiftPostgresql15HibernateReactiveIT` passes on OpenShift with PostgreSQL 15
- [ ] Verify `OpenShiftPostgresqlMultitenantHibernateSearchIT` passes on OpenShift with PostgreSQL 16
- [ ] Verify `OpenShiftPostgresql15DatabaseIT` passes on OpenShift with PostgreSQL 15

### DIFF
--- a/hibernate/hibernate-fulltext-search/src/test/java/io/quarkus/ts/hibernate/search/OpenShiftPostgresqlMultitenantHibernateSearchIT.java
+++ b/hibernate/hibernate-fulltext-search/src/test/java/io/quarkus/ts/hibernate/search/OpenShiftPostgresqlMultitenantHibernateSearchIT.java
@@ -27,7 +27,7 @@ public class OpenShiftPostgresqlMultitenantHibernateSearchIT extends AbstractMul
             .withProperty("cluster.routing.allocation.disk.threshold_enabled", "false")
             .withProperty("xpack.security.enabled", "false");
 
-    @Container(image = "${postgresql.13.image}", port = POSTGRESQL_PORT, expectedLog = "listening on IPv4 address")
+    @Container(image = "${postgresql.latest.image}", port = POSTGRESQL_PORT, expectedLog = "listening on IPv4 address")
     static PostgresqlService database = new PostgresqlService();
 
     @QuarkusApplication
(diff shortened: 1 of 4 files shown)


hibernate/hibernate-reactive/src/test/java/io/quarkus/ts/hibernate/reactive/openshift/OpenShiftPostgresql13HibernateReactiveIT.java
```diff
@@ -26,6 +26,7 @@ public class OpenShiftPostgresql13HibernateReactiveIT extends AbstractDatabaseHi
 
     @QuarkusApplication
     static RestService app = new RestService().withProperties("postgresql.properties")
+            .withProperty("quarkus.datasource.db-version", "13")
```
jedla97: Looking at git history these two seems to be bumped to oldest available and supported by RH and Hibernate.

So I believe we should move to https://catalog.redhat.com/en/software/containers/rhel9/postgresql-15/63f763f779eb1214c4d6fcf6 and rename the tests again to `OpenShiftPostgresql15` 

I looked and the 14 is not on RH image registry so 15 is probably best option to tests older PG version

jedla97 (commented):
I have one comment. With moving the `OpenShiftPostgresqlMultitenantHibernateSearchIT`  to latest version I don't see problem

jedla97 (approved):
Thanks!

jedla97:
run tests

PR #3053, opened 2026-06-23
Fix BearerFilesystemIT native mode failure: use already-expired dummy JWT in onPreStart

gastaldi:
### Summary

`BearerFilesystemIT.missingTokenFileTest` was consistently failing in native mode. Quarkus eagerly validates the bearer token file at startup ([quarkusio/quarkus#55045](https://github.com/quarkusio/quarkus/issues/55045)), so the workaround writes a dummy JWT in `onPreStart`. In native mode the app starts in ~0.4 s, meaning the first test ran while the dummy token's cache was still valid — Quarkus reused the cached token instead of re-reading the deleted file, so the expected `"Cannot find a file with a JWT bearer token at path: ..."` log message never appeared.

**Fix:** write a dummy JWT that is already expired (`exp = now - 1 s`). Quarkus's startup validation accepts the file (it exists and is parseable), but since the token is expired it is never cached — the file is re-read on every request, which is exactly what all four test scenarios require. This also removes the `DUMMY_TOKEN_EXPIRY_SECONDS` constant and the 7-second `Thread.sleep` that the previous attempt required.

To verify manually, run `BearerFilesystemIT` in native mode — all four ordered tests (`missingTokenFile`, `emptyTokenFile`, `malformedToken`, `validToken`) should pass without any sleep.

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/security/keycloak-oidc-client-extended/src/test/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/BearerFilesystemIT.java
+++ b/security/keycloak-oidc-client-extended/src/test/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/BearerFilesystemIT.java
@@ -8,6 +8,7 @@ import static io.restassured.RestAssured.given;
 import java.io.File;
 import java.io.IOException;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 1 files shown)


mocenas (approved):
I'm OK with the changes, but there is some related fail in the CI, please fix that.

fedinskiy (approved):

PR #3060, opened 2026-06-29
Fix daily build

fedinskiy:
### Summary

- Use maven settings consistently in all builds
- Rename platform build to 999-core-3.x-SNAPSHOT
- Remove download logging while setting properties

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -28,24 +28,25 @@ jobs:
           # Save local repo to GH ENV to ensure that it's the same in all steps
           mkdir -p maven-repository
           echo "LOCAL_REPO=$(pwd)/maven-repository" >> $GITHUB_ENV
+          echo "MVN_SETTINGS=$(pwd)/.github/mvn-settings.xml" >> $GITHUB_ENV
       - name: Build Quarkus ${QUARKUS_BUILD}
         run: |
           git clone -b 3.x https://github.com/quarkusio/quarkus.git && cd quarkus
-          ./mvnw -B --no-transfer-progress -s .github/mvn-settings.xml clean install -Dquickly -Dno-test-modules -Prelocations -Dmaven.repo.local=${{ env.LOCAL_REPO }}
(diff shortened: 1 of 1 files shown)


.github/workflows/daily.yaml
```diff
@@ -28,36 +28,37 @@ jobs:
...
+      - name: Build Quarkus Platform main with Quarkus core ${QUARKUS_BUILD} (version 999-core-3.x-SNAPSHOT)
         run: |
           git clone https://github.com/quarkusio/quarkus-platform.git && cd quarkus-platform
 
           # add some common options
           cat <<EOF >> .mvn/maven.config
             --no-transfer-progress
             -Dmaven.repo.local=${{ env.LOCAL_REPO }}
-            -s.github/mvn-settings.xml
+            -s${{ env.SETTINGS }}
```
jedla97: Afaik space should be here

fedinskiy: I thought so, but it leads to failures. Some weird behavior on `maven.config` part.

.github/workflows/daily.yaml
```diff
@@ -28,36 +28,37 @@ jobs:
           # Save local repo to GH ENV to ensure that it's the same in all steps
           mkdir -p maven-repository
           echo "LOCAL_REPO=$(pwd)/maven-repository" >> $GITHUB_ENV
+          echo "SETTINGS=$(pwd)/.github/mvn-settings.xml" >> $GITHUB_ENV
```
jedla97: Also maybe could this variable be like `MVN_SETTINGS` as setting seems something which could GH use in future to pass their settings  for something

.github/workflows/daily.yaml
jedla97: If it's work I think i's ok to leave as it si, but it seems weird

.github/workflows/daily.yaml
fedinskiy: Good idea! Implemented

jedla97 (approved):

PR #3067, opened 2026-07-03
Test coverage for QUARKUS-7341

SlimaneAbzar:
## Summary

JIRA: https://redhat.atlassian.net/browse/QUARKUS-7341
Upstream PR: https://github.com/quarkusio/quarkus/pull/52353
Test plan: https://github.com/quarkus-qe/quarkus-test-plans/pull/308

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

Checklist:

- [x] Commits are well encapsulated and follow the best practices


jedla97:
I'll check this today

security/keycloak-oidc-client-extended/src/test/resources/rar-realm.json
```diff
@@ -0,0 +1,2151 @@
+{
```
jedla97, line 1: Could this be trim to some minimal version as we using in other Keycloak setups.

security/keycloak-oidc-client-extended/src/test/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/RarNegativeIT.java
```diff
@@ -0,0 +1,69 @@
...
+            loginForm.getInputByName("username").setValueAttribute("alice");
+            loginForm.getInputByName("password").setValueAttribute("alice");
+
+            try {
+                loginForm.getButtonByName("login").click();
+            } catch (FailingHttpStatusCodeException e) {
+                assertEquals(401, e.getStatusCode());
+            }
+
+            webClient.getCookieManager().clearCookies();
```
jedla97: Afaik I don't think you need to clear cookies as this is only one test and you clear them at the end of the test.

security/keycloak-oidc-client-extended/src/test/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/AbstractRarIT.java
```diff
@@ -0,0 +1,95 @@
...
+            Page resultPage = loginForm.getButtonByName("login").click();
+            String authorizationDetails = resultPage.getWebResponse().getContentAsString().trim();
+
+            assertNotNull(authorizationDetails);
+            assertTrue(authorizationDetails.contains("type=openid_credential"),
+                    "Expected type=openid_credential but was: " + authorizationDetails);
+            assertTrue(authorizationDetails.contains("credential_configuration_id"),
+                    "Expected credential_configuration_id but was: " + authorizationDetails);
+
+            webClient.getCookieManager().clearCookies();
```
jedla97: Is it necessary to clear the cookies as you don't reuse webClient?

security/keycloak-oidc-client-extended/src/test/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/AbstractRarIT.java
```diff
@@ -0,0 +1,95 @@
...
+        try (WebClient webClient = createWebClient()) {
+            HtmlPage page = webClient.getPage(
+                    app.getURI(Protocol.HTTP).withPath("/rar/token-response/authorization-details").toString());
+
+            HtmlForm loginForm = page.getForms().get(0);
+            loginForm.getInputByName("username").setValueAttribute("alice");
+            loginForm.getInputByName("password").setValueAttribute("alice");
+
+            Page resultPage = loginForm.getButtonByName("login").click();
+            String authorizationDetails = resultPage.getWebResponse().getContentAsString().trim();
```
jedla97, line 52: I'm looking at this and think you should if it's passed as payload to Keycloak instead of storing it as body, Maybe I missing something.

From my POV I see that we should test that these values are passed to Keycloak as Keycloak doing something with them internally.

I play with your code and wasn't able to trigger the filter, only the default values was passed to Keycloak as part of the payload header, not the custom ones. I believe the filter should not have `@AuthorizationCodeFlow`, but it should be on endpoint and your filter should have these 3 annotation
```
@TenantFeature("tenant")
@ApplicationScoped
@Unremovable
```

security/keycloak-oidc-client-extended/src/test/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/rar/RarRedirectFilter.java
```diff
@@ -0,0 +1,45 @@
+package io.quarkus.ts.security.keycloak.oidcclient.extended.restclient.rar;
```
jedla97: Why all of these are in `src/test` and not in `src/main` ? Should TenantFeature prevent it from activating it on different tenant? So it should not be problem to have it with other classes in `src/main`

jedla97 (changes requested):
Hi, sorry for late review. I have few question as I'm not sure this tests the feature as it should, or at least I don't see the flow so maybe I just missing something.

security/keycloak-oidc-client-extended/src/test/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/RarNegativeIT.java
SlimaneAbzar: You're right I'll remove it.

security/keycloak-oidc-client-extended/src/test/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/AbstractRarIT.java
SlimaneAbzar: Thanks, I'll try removing @AuthorizationCodeFlow from the filter and add @Unremovable instead.

security/keycloak-oidc-client-extended/src/test/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/rar/RarRedirectFilter.java
SlimaneAbzar: I followed the step-up auth pattern, but I'll try move to  `src/main`,

security/keycloak-oidc-client-extended/src/test/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/AbstractRarIT.java
```diff
@@ -0,0 +1,84 @@
...
+            .withProperty("quarkus.oidc.auth-server-url", () -> keycloak.getRealmUrl())
+            .withProperties(() -> keycloak.getTlsProperties());
+
+    @Test
+    void testConfigBasedRarAuthorizationDetails() throws IOException {
+        try (WebClient webClient = createWebClient()) {
+            HtmlPage page = webClient.getPage(
+                    app.getURI(Protocol.HTTP).withPath("/rar/token-response/authorization-details").toString());
+
+            HtmlForm loginForm = page.getForms().get(0);
```
jedla97, line 47: Afaik reading one of my and I forget to mention one thing.

I think it would be better to check if the values are actually passed to Keycloak. Something like `page.getBaseURL()` and checked if expected values are there. As in my POV you entering the values from `getAuthorizationDetails` which can be there but there is possibility that these values are not passed to Keycloak

security/keycloak-oidc-client-extended/src/main/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/rar/RarRedirectFilter.java
```diff
@@ -0,0 +1,31 @@
...
+@Unremovable
+public class RarRedirectFilter implements OidcRedirectFilter {
+
+    @Override
+    public void filter(OidcRedirectContext redirectContext) {
+        String authServerUrl = redirectContext.oidcTenantConfig().authServerUrl().orElseThrow();
+
+        JsonArray authorizationDetailsArray = new JsonArray();
+
+        JsonObject obj = new JsonObject();
```
jedla97: You setting these values also in property file. How can we be sure that these are changed by this redirect filter?

I think there are 2 option. Either these will be changed to some different and in that case we get 401 or you restart Quarkus app with different values and change them in this filter to correct one. Something like this https://github.com/quarkus-qe/quarkus-test-suite/blob/main/nosql-db/mongodb/src/test/java/io/quarkus/ts/nosqldb/mongodb/MongoClientMtlsIncorrectSettingsIT.java#L104

I thinks both are valid approaches and I don't have preferences here

jedla97 (commented):
Thanks for update! I think it's look fine, I have just two additional comment

security/keycloak-oidc-client-extended/src/main/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/rar/RarRedirectFilter.java
```diff
@@ -0,0 +1,31 @@
+package io.quarkus.ts.security.keycloak.oidcclient.extended.restclient.rar;
```
jedla97: As you went with creating the new filter for negative scenario, do we need this one? Basically this filter don't change any value compared to `rar.properties` and you don't checking if it was used.

security/keycloak-oidc-client-extended/src/test/java/io/quarkus/ts/security/keycloak/oidcclient/extended/restclient/RarNegativeIT.java
```diff
@@ -0,0 +1,84 @@
...
+    @Test
+    void testWrongRedirectFilterRarTypeProducesError() throws IOException {
+        try (WebClient webClient = createWebClient()) {
+            HtmlPage page = webClient.getPage(
+                    app.getURI(Protocol.HTTP).withPath("/rar-redirect-wrong/token-response/authorization-details")
+                            .toString());
+
+            HtmlForm loginForm = page.getForms().get(0);
+            loginForm.getInputByName("username").setValueAttribute("alice");
+            loginForm.getInputByName("password").setValueAttribute("alice");
```
jedla97, line 73: Could you check here also the url using `page.getBaseURL()`. Also check if the value is same as the one set in filter.

jedla97 (changes requested):
Thanks for update! I have one question and one comment.

jedla97 (approved):
Thanks for updates!

jedla97:
run tests

PR #3082, opened 2026-07-16
Skip DevModeWorkspaceIT.workspaceCanBeEdited on AArch64

gastaldi:
## Summary

- Skip `DevModeWorkspaceIT.workspaceCanBeEdited` on AArch64 runners using `@DisabledIfSystemProperty(named = "ts.arm.missing.services.excludes")`

The Playwright headless Chromium browser consistently times out (30s) waiting for the `#code` editor element on AArch64 runners. Verified this has been failing on every AArch64 run since at least Jul 10 while always passing on x86_64.

## Test plan

- [ ] `DevModeWorkspaceIT` passes on Linux x86_64 JVM
- [ ] `DevModeWorkspaceIT.workspaceCanBeEdited` is skipped on Linux AArch64 JVM


mjurc (changes requested):
This is valid test that tests a feature that should be working. The test also worked and passed with 3.33 release. Can you see why this is failing on aarch64 and fix the test?

gastaldi:
Yeah, you're right, it passes locally. I'll close this and investigate further what's causing this

PR #3094, opened 2026-07-24
Add update policy for repository

jedla97:
### Summary

This is useful especially for local developments currently it trying to download metadata for platform version which is not available. When running without core-only profile it still stay same value as default one. When running the core-only the value for update is never.

Currently every time when I run some tests it's try to download `maven-metadat.xml`. I don't see much point to even try to download the the as they are not existing. Possibly when you run much tests daily it could even cause rate limit (never happen to me yet, but I not running that much tests currently). Also downloading it every run create log noise.

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -62,6 +62,7 @@
         <reruns>2</reruns>
         <oc.reruns>1</oc.reruns>
         <flaky-run-reporter.version>0.1.7</flaky-run-reporter.version>
+        <repository.update.policy>daily</repository.update.policy>
     </properties>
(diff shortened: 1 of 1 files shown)


fedinskiy (approved):
LGTM, but please check the CI

jedla97:
Looking at failures and they are not related, they are same as daily, RAR (keycloak extended) tests and snappy (kafka producer). So merging this

PR #3097, opened 2026-07-27
Test Coverage for QUARKUS-7866 - Add JsonProvider SPI for per-record dynamic JSON field additions

Andy-Y1314:
### Summary
TP: https://github.com/quarkus-qe/quarkus-test-plans/pull/312
JIRA: https://redhat.atlassian.net/browse/QUARKUS-7866

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/logging/jboss/src/main/java/io/quarkus/ts/logging/jboss/CdiJsonProvider.java
+++ b/logging/jboss/src/main/java/io/quarkus/ts/logging/jboss/CdiJsonProvider.java
@@ -1,0 +1,36 @@
+package io.quarkus.ts.logging.jboss;
+
+import jakarta.enterprise.context.ApplicationScoped;
+
+import org.jboss.logmanager.ExtLogRecord;
+import org.jboss.logmanager.Level;
+
+import io.quarkus.arc.lookup.LookupIfProperty;
+import io.quarkus.logging.json.runtime.JsonFormatter.JsonLogGenerator;
+import io.quarkus.logging.json.runtime.JsonProvider;
+
+@ApplicationScoped
+@LookupIfProperty(name = "json-provider-test.enabled", stringValue = "true")
+public class CdiJsonProvider implements JsonProvider {
+    @Override
+    public void writeTo(JsonLogGenerator generator, ExtLogRecord record) throws Exception {
+        generator.add("customField", "customValue");
+        generator.add("loggerNameFromRecord", record.getLoggerName());
+
+        String requestId = record.getMdcCopy().get("requestId");
+        if (requestId != null) {
+            generator.add("requestIdFromMdc", requestId);
+        }
+
+        if (record.getLevel().intValue() >= Level.ERROR.intValue()) {
+            generator.add("isErrorRecord", "true");
+        }
+
+        generator.startObject("nestedObject")
+                .add("nestedField1", "nestedValue1")
+                .add("nestedField2", "nestedValue2")
+                .endObject();
+
+        generator.add("excludedField", "excludedValue");
+    }
+}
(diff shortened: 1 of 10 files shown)


Andy-Y1314:
I split the info and error tests because app.logs() asserts against the accumulated application log. For negative scenarios I needed a new log to verify it.

logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/JsonProviderConsoleInfoIT.java
```diff
@@ -0,0 +1,31 @@
...
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.QuarkusApplication;
+
+@Tag("QUARKUS-7866")
+@QuarkusScenario
+public class JsonProviderConsoleInfoIT {
```
jedla97: Could not this be merged with `JsonProviderConsoleErrorIT`? You accessing different endpoints, so there should not be problem and it only differ with `quarkus.log.console.json.excluded-keys` which you not check in console error either. So I think it can be merged

jedla97: Note: probably same goes for other tests as well.

Also not sure if it's work that way but technically if you set the `quarkus.log.file.enabled` the console output should still stay (not 100% sure about that) so it could be merged to one test class, but I'm not oppossed  to have it in different classes to differentiate between console and file tests

logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/JsonProviderFileErrorIT.java
```diff
@@ -0,0 +1,45 @@
...
+    static void cleanup() throws IOException {
+        app.stop();
+        Files.deleteIfExists(LOG_FILE_DIR.resolve(LOG_FILE_NAME));
+    }
+
+    @Test
+    void shouldWriteErrorFieldsToFile() throws IOException {
+        app.given().get("/json-provider/error");
+
+        String log = Files.readString(LOG_FILE_DIR.resolve(LOG_FILE_NAME));
```
jedla97: Maybe you could check first if the file exists, before reading it

jedla97 (commented):
Overall I think it's look fine, I have just 2-3 suggestion

logging/jboss/src/test/java/io/quarkus/ts/logging/jboss/JsonProviderConsoleInfoIT.java
Andy-Y1314: > Could not this be merged with `JsonProviderConsoleErrorIT`? You accessing different endpoints, so there should not be problem and it only differ with `quarkus.log.console.json.excluded-keys` which you not check in console error either. So I think it can be merged

Yes, I originally had both cases in the same test. The issue is that app.logs() returns the accumulated log output, so after calling /json-provider/info, the MDC field remains in the captured log and the ERROR test cannot assert that it is absent then, hence the 2 separate test files.

Andy-Y1314: > Note: probably same goes for other tests as well.
> 
> Also not sure if it's work that way but technically if you set the `quarkus.log.file.enabled` the console output should still stay (not 100% sure about that) so it could be merged to one test class, but I'm not oppossed to have it in different classes to differentiate between console and file tests

I think we should keep it as is so we can differentiate the tests.

jedla97: > The issue is that app.logs() returns the accumulated log output

In that case after each test you can restart the quarkus app. Something like app.stop() and app.start() it will clear logs. I personally think it's better that have one class for one test

Andy-Y1314: Yes I agree. I'll try that thanks!

jedla97 (approved):
Thanks!

PR #3098, opened 2026-07-29
Fix OIDC RAR tests as KC 26.7 rework it

jedla97:
### Summary

Fixing the RAR test as KC 26.7 rework some thing on their side (this is experimental feature in KC).

Now the KC fail even before redirect, which in my POV is better and correct, instead of allowing the redirect and send the unauthorized information.

Also extend the check of logs in RarIT to check for all set part of RAR request.

Update:
The RHBK using 26.6 and they from experience releasing every 2 version so most likely the 26.8 is next. As the 26.6 no include changes, done in 26.7 It's need temporary it's own realm json file without setting `verifiableCredentials`. I added TODO to change it back.

- [ ] Dependency update
- [x] Refactoring
- [ ] Backport

### Checklist:


jedla97:
run tests

jedla97:
run tests

SlimaneAbzar (approved):
Looks good to me.

rsvoboda:
Interesting, @SlimaneAbzar were these changes mentioned in release notes of Keycloak 26.7 ?

jedla97:
@rsvoboda the changes for verifiable credentials are mentioned in RN https://www.keycloak.org/docs/latest/release_notes/index.html#progress-on-verifiable-credentials-oid4vci-experimental 

btw it is experimental feature on KC side so it can change in future. I think current flow make more sense as if the RAR pass wrong values from Quarkus app it fail and throw 401 instead of the previous flow when the 401 was returned after the user login even with wrong RAR parameters

PR #3104, opened 2026-08-03
Fix Windows-only flaky tests with Awaitility polling

gastaldi:
## Summary

Fixes https://redhat.atlassian.net/browse/QQE-2819

- **`AccessLoggingIT.accessTest`**: Access log flushing is delayed on Windows — the third entry may not be written when the assertion runs immediately after sending requests. Replace direct `assertEquals` with `Awaitility.await()` polling.
- **`UniFailureCacheIT.testSizeBasedEviction`**: Caffeine cache eviction is asynchronous and may not complete immediately on slower/Windows runners. Replace direct assertion with `Awaitility.await()` polling until eviction completes.

Both tests pass consistently on Linux but fail intermittently on Windows due to timing differences in filesystem flushing and background thread scheduling.

## Test plan

- [x] `AccessLoggingIT.accessTest` passes on Windows JVM
- [x] `UniFailureCacheIT.testSizeBasedEviction` passes on Windows JVM
- [x] Both tests still pass on Linux JVM


jcarranzan (approved):

rsvoboda (commented):
LGTM!

PR #3108, opened 2026-08-05
Install Maven settings in test jobs to avoid Maven Central rate-limiting

gastaldi:
## Summary

Copy `.github/mvn-settings.xml` to `~/.m2/settings.xml` in each test job (Linux JVM, Linux Native, Windows JVM, AArch64 JVM, AArch64 Native) so that all Maven invocations — including `mvnw` spawned by the test framework inside generated apps — use the Google Maven Central mirror.

Currently, the settings file is only passed via `-s` to the main `mvn` command, but tests like `CliProdModeJvmIT` create Quarkus apps and run `mvnw build` on them. These spawned processes don't inherit the `-s` flag and hit `repo.maven.apache.org` directly, resulting in `429 Too Many Requests` errors.

Evidence:
- [Daily Build Aug 3 — AArch64 JVM 25 root-modules](https://github.com/quarkus-qe/quarkus-test-suite/actions/runs/30961557841/job/92171950243): `CliProdModeJvmIT.extensionSetA` and `extensionSetB` failed with `Could not transfer artifact ... status code: 429, reason phrase: Too Many Requests`
- [Daily Build Aug 2 — Windows JVM 25](https://github.com/quarkus-qe/quarkus-test-suite/actions/runs/30773081486/job/91567206565): AMQP, QPID, OpenTelemetry modules failed with `Connect to central.sonatype.com:443 failed: Connect timed out`

## Test plan

- [ ] `CliProdModeJvmIT` passes without 429 errors
- [ ] Daily build passes on all platforms

### DIFF
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -93,6 +93,8 @@ jobs:
       - uses: ./.github/actions/cache-quarkus
       - uses: ./.github/actions/prepare-quarkus-cli
       - uses: ./.github/actions/use-docker-mirror
+      - name: Install Maven settings
+        run: cp .github/mvn-settings.xml ~/.m2/settings.xml
       - name: Test in JVM mode
         run: |
           mvn -fae -V -B -nsu --no-transfer-progress clean verify -P ${{ matrix.profiles }} -Dinclude.quarkus-cli-tests -Dts.quarkus.cli.cmd="${PWD}/quarkus-dev-cli"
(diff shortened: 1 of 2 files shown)


.github/workflows/daily.yaml
```diff
@@ -161,6 +165,9 @@ jobs:
           check-latest: true
       - uses: ./.github/actions/cache-quarkus
       - uses: ./.github/actions/prepare-quarkus-cli-win
+      - name: Install Maven settings
+        shell: bash
```
rsvoboda, line 169: This (`shell: bash`) is not in other sections, is it because it's Windows?

rsvoboda (dismissed):
`windows-build-native-latest` uses `-s .github/mvn-settings.xml`

It should be also switch to copy approach

.github/workflows/daily.yaml
jedla97: On Linux runners the bash is default, on windows the PowerShell is default so it§s need to overwritten

jedla97 (commented):
I think this should be also added to PR CI (https://github.com/quarkus-qe/quarkus-test-suite/blob/main/.github/workflows/ci.yml) to prevent this problem

.github/workflows/daily.yaml
gastaldi: Exactly

jedla97 (approved):

rsvoboda (commented):
Great, thanks for the update.

PR #3127, opened 2026-08-18
Switch main branch from Quarkus 3.x to 3.39 as 3.x was deleted

jedla97:
### Summary

Same as https://github.com/quarkus-qe/quarkus-test-framework/pull/1964. It will need new FW snapshot before this. The failure is expected as we need it to build and cache it first. After the merge I'll run the daily to compare it with older one

Please select the relevant options.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Dependency update
- [x] Refactoring
- [ ] Backport
- [ ] New scenario (non-breaking change which adds functionality)
- [ ] This change requires a documentation update
- [ ] This change requires execution against OCP (use `run tests` phrase in comment)
- [ ] This change requires execution with OCP on Aarch64 (use `run arm tests` phrase in comment)

### Checklist:
- [x] Methods and classes used in PR scenarios are meaningful
- [x] Commits are well encapsulated and follow [the best practices](https://cbea.ms/git-commit/)

PR #3128, opened 2026-08-20
Add coverage for QUARKUS-8368 (Hibernate ORM 7.4 upgrade)

jcarranzan:
### Summary

JIRA: https://issues.redhat.com/browse/QUARKUS-8368

Upstream PR: https://github.com/quarkusio/quarkus/pull/54083

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/hibernate/hibernate-orm/src/main/java/io/quarkus/qe/hibernate/temporal/TemporalResource.java
+++ b/hibernate/hibernate-orm/src/main/java/io/quarkus/qe/hibernate/temporal/TemporalResource.java
@@ -1,0 +1,61 @@
+package io.quarkus.qe.hibernate.temporal;
+
+import java.time.Instant;
+
+import jakarta.enterprise.event.Observes;
+import jakarta.inject.Inject;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.PathParam;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.MediaType;
+
+import org.hibernate.SessionFactory;
+
+import io.quarkus.hibernate.orm.PersistenceUnit;
+import io.quarkus.runtime.StartupEvent;
+
+@Path("/temporal")
+public class TemporalResource {
+
+    @PersistenceUnit("temporal-pu")
+    @Inject
+    SessionFactory sessionFactory;
+
+    private volatile Instant snapshotInstant;
+
+    public void startup(@Observes StartupEvent event) throws InterruptedException {
+        sessionFactory.inTransaction(session -> {
+            session.persist(new TemporalProduct(1L, "Original Product"));
+        });
+
+        Thread.sleep(500);
+        snapshotInstant = Instant.now();
+        Thread.sleep(500);
+
+        sessionFactory.inTransaction(session -> {
+            var product = session.find(TemporalProduct.class, 1L);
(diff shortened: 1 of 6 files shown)


jedla97:
For CI fix please wait until https://github.com/quarkus-qe/quarkus-test-suite/pull/3127 is merged + cache is populated

jcarranzan:
> For CI fix please wait until #3127 is merged + cache is populated

Yep, I noticed the change of dropping quarkus branch 3.x and I read the slack thread yesterday, thanks!

gtroitsk:
@jcarranzan I can review your PR tonight

hibernate/hibernate-orm/src/main/java/io/quarkus/qe/hibernate/temporal/TemporalResource.java
```diff
@@ -0,0 +1,61 @@
...
+    SessionFactory sessionFactory;
+
+    private volatile Instant snapshotInstant;
+
+    public void startup(@Observes StartupEvent event) throws InterruptedException {
+        sessionFactory.inTransaction(session -> {
+            session.persist(new TemporalProduct(1L, "Original Product"));
+        });
+
+        Thread.sleep(500);
```
gtroitsk, line 32: `Thread.sleep()` is a bit risky here as it makes the test timing-dependent and potentially flaky.

Could we make the `asOf()` timestamp deterministic by using the actual temporal timestamp generated and stored by the DB? That should be safer than relying on `Instant.now()` and fixed delays, because we would use the timestamp of the actual temporal version created in the DB

Maybe the `HISTORY_TABLE` strategy could help here: [Hibernate temporal data - HISTORY_TABLE](https://docs.hibernate.org/orm/7.4/introduction/html_single/#temporal-data)

gtroitsk (commented):
The Pagination part of the TD looks good. I just have one suggestion regarding the Temporal entity test.

Also, please take a look at the CI failures

hibernate/hibernate-orm/src/main/java/io/quarkus/qe/hibernate/temporal/TemporalResource.java
jcarranzan: When I did it, I checked the upstream code here https://github.com/quarkusio/quarkus/blame/576f6607b4a35201dcf7c0a078cde8064a79896f/integration-tests/hibernate-orm-temporal/src/main/java/io/quarkus/it/temporal/TemporalResource.java#L38 .
 But you're right, and I agree it makes test timing dependent. I'll investigate if it's possible to do it more robustly

jcarranzan: HISTORY_TABLE is a storage strategy (separate history table vs same table),  not a mechanism for reading temporal timestamps.

The only Hibernate 7.4 approach that would avoids Thread.sleep is `@Audited` + `@Changelog`  (uses changeset IDs instead of timestamps), which is a different feature  excluded from the TP scope.
The Thread.sleep pattern follows upstream exactly, in fact I increase from 100 to 500... The feature is incubating with no documented alternative for deterministic temporal testing. Keep as-is.

jcarranzan:
run tests

gtroitsk (approved):
Thank you

Issue #3130, opened 2026-08-20
ai/mcp: BasicMCPIT.roots fails — 'Roots not supported' vs 'No roots found' (clarification needed)


gastaldi:
## Summary

The [Daily Build run 32362778910](https://github.com/quarkus-qe/quarkus-test-suite/actions/runs/32362778910) failed with **9 failing jobs**, all in the `roots` test of the `ai/mcp` module (`BasicMCPIT.roots()`, run by `LocalStreamingHTTPIT`, `StdioIT`, `WebSocketIT`) — deterministically, on every platform (Windows/Linux/AArch64, JVM 17/21/25, Native).

This is a request for clarification on whether these are **intended upstream behavior changes** (so the test should be reworked) or a **regression**.

> **Note:** the initial analysis (below, "Change 1") thought this was a single string-expectation mismatch. CI on the interim PR #3131 revealed a **second, larger change** ("Change 2"): the client API the test used to configure roots has been removed. Both are described below.

## Failures

Under MCP protocol `2026-07-28` and later, **two** distinct behavior changes affect `BasicMCPIT.roots()`:

### Change 1 — capability not advertised before a root is set (line 120)

```
io.quarkus.ts.mcp.BasicMCPIT.roots(BasicMCPIT.java:120)
org.opentest4j.AssertionFailedError: expected: <No roots found> but was: <Roots not supported>
```

`GET /mcp/roots` (`MCPClient.java:187`) invokes the server-side `rootchecker` tool via the langchain4j MCP client. The tool `getRoots(Roots roots)` (`FileServer.java:234`):
```java
if (!roots.isSupported()) {
    return "Roots not supported";
}
return roots.list()... // empty -> "No roots found"
```
The client no longer advertises the `roots` capability before a root is configured, so `roots.isSupported()` went `true → false`, returning `"Roots not supported"` instead of `"No roots found"`.

### Change 2 — `McpClient.setRoots()` removed (line 125)

After fixing the string above, CI failed at the next step:
```
io.quarkus.ts.mcp.BasicMCPIT.roots(BasicMCPIT.java:125)
org.opentest4j.AssertionFailedError: expected: <204> but was: <500>

Caused by: java.lang.UnsupportedOperationException: setRoots is not supported with MCP protocol 2026-07-28 or later
    at dev.langchain4j.mcp.client.DefaultMcpClient.setRoots(DefaultMcpClient.java:946)
```
[...]

PR #3131, opened 2026-08-20
Disable BasicMCPIT.roots pending MCP roots API clarification (#3130)

gastaldi:
## Summary

Disables `ai/mcp` `BasicMCPIT.roots()` because it is broken by an MCP protocol behavior change (MCP protocol `2026-07-28` and later), which the Daily Build now pulls in via the platform BOM.

Two independent changes break the test:

1. **Before any root is configured**, the langchain4j MCP client no longer advertises the `roots` capability, so the server-side `rootchecker` tool returns `"Roots not supported"` instead of `"No roots found"` (fails at line 120).
2. **`McpClient.setRoots()` is no longer supported** — it now throws:
   ```
   java.lang.UnsupportedOperationException: setRoots is not supported with MCP protocol 2026-07-28 or later
       at dev.langchain4j.mcp.client.DefaultMcpClient.setRoots(DefaultMcpClient.java:946)
   ```
   so `POST /mcp/roots` returns HTTP 500 and the test fails at line 125 with `expected: <204> but was: <500>`.

Because the client API the test relied on to configure roots is gone, the test cannot simply have its string expectations adjusted — the correct new expectation (and how roots should be configured under the new protocol) is unknown. This PR disables the test pending clarification in #3130 to unblock the Daily Build, following the repo's `@Disabled("<issue-url>")` convention.

## Context

This started as a simple expectation adjustment, but CI revealed the `setRoots` API removal, so the interim fix was changed to disabling the test.

- Failing Daily Build: https://github.com/quarkus-qe/quarkus-test-suite/actions/runs/32362778910
- Failing PR CI (expectation-only attempt): https://github.com/quarkus-qe/quarkus-test-suite/actions/runs/32399353924
- Related issue: #3130

### DIFF
--- a/ai/mcp/src/test/java/io/quarkus/ts/mcp/BasicMCPIT.java
+++ b/ai/mcp/src/test/java/io/quarkus/ts/mcp/BasicMCPIT.java
@@ -5,6 +5,7 @@ import java.util.List;
 
 import org.apache.http.HttpStatus;
 import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 import io.quarkus.test.bootstrap.RestService;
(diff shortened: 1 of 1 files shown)


fedinskiy (approved):
Thank you!

I would say, we should have create the issue upstream (quarkus-langchain4j) and not in the test suite.

jcarranzan:
Let's see how's going the issue opened discussion, I guess we will need to adapt the test after that.

PR #3153, opened 2026-09-09
Throttle memory leak test polling to avoid port exhaustion

Ladicek:
The test watches a counter that `EventsClientResource` increments on its own, polling it in a tight loop (more than 1000 requests per second), which may exhaust the ephemeral port range (very easy on Windows). If so, the run fails with `BindException: Address already in use: connect`.

Wiht this commit, the test polls every 100 ms instead, which makes the request volume 2 orders of magnitude smaller without any effect on the ability to reproduce the OOM (verified the test still fails on 3.27.4 and passes on 3.27.5). Also, progress is reported by counter delta rather than on exact multiples of 1000, which was frequently stepped over.

### Summary

- [ ] Dependency update
- [ ] Refactoring
- [ ] Backport

### Checklist:

### DIFF
--- a/http/rest-client-reactive/src/test/java/io/quarkus/ts/http/restclient/reactive/RestClientReactiveMemoryLeakIT.java
+++ b/http/rest-client-reactive/src/test/java/io/quarkus/ts/http/restclient/reactive/RestClientReactiveMemoryLeakIT.java
@@ -41,13 +41,19 @@ public class RestClientReactiveMemoryLeakIT {
         assertEquals(200, response.statusCode());
         long count = Long.parseLong(response.getBody().asString());
         assertNotEquals(0, count);
+        long lastLogged = count;
         try {
-            while (count < (200_000)) { // as of 3.36.1, the job fails around 100 000, but let's have some margin for error
+            // as of 3.27.4 and 3.36.1, the job fails around 100 000, let's have some margin for error
+            while (count < 200_000) {
+                // the app increments the counter on its own, polling faster only exhausts ephemeral ports faster
+                Thread.sleep(100);
+
                 response = client.given().get("/client/events");
                 assertEquals(200, response.statusCode());
                 count = Long.parseLong(response.getBody().asString());
-                if (count % 1000 == 0) {
+                if (count - lastLogged >= 10_000) {
                     Log.info("Current count is " + count);
+                    lastLogged = count;
                 }
             }
         } catch (Exception ex) {


Ladicek:
Once this is merged, I'll need to backport to 3.27, 3.33 and 3.40.

http/rest-client-reactive/src/test/java/io/quarkus/ts/http/restclient/reactive/RestClientReactiveMemoryLeakIT.java
```diff
@@ -41,13 +41,19 @@ public void httpServer() {
...
+            // as of 3.27.4 and 3.36.1, the job fails around 100 000, let's have some margin for error
+            while (count < 200_000) {
+                // the app increments the counter on its own, polling faster only exhausts ephemeral ports faster
+                Thread.sleep(100);
+
                 response = client.given().get("/client/events");
                 assertEquals(200, response.statusCode());
                 count = Long.parseLong(response.getBody().asString());
-                if (count % 1000 == 0) {
+                if (count - lastLogged >= 1000) {
```
fedinskiy:
```suggestion
                if (count - lastLogged >= 10_000) {
```

The original logging was intentionally weird so it will not output too many messages. I presume, 10K granularity is good enough for this case

fedinskiy (changes requested):
One more thing: does this patch really fix port exhaustion? Most of connections in this test are happening in the EventsClientResource class, which is not affected by changes

http/rest-client-reactive/src/test/java/io/quarkus/ts/http/restclient/reactive/RestClientReactiveMemoryLeakIT.java
Ladicek: 1000 isn't too bad, but 10_000 should be just fine too.

Ladicek:
> One more thing: does this patch really fix port exhaustion? Most of connections in this test are happening in the EventsClientResource class, which is not affected by changes

I don't know how the client behind `EventsClientResource` (the generated implementation of `EventsClient`) works, it is quite possible it uses pipelining or something.

The client in the test, which I'm changing here, has demonstrably reached the limit of ephemeral ports on Windows (it is weird noone has seen this before, but it feels possible the Windows setup on GitHub Actions is different from our Windows in AWS) and the commit here does fix that.

There are Windows settings one can do (increase the ephemeral port range, reduce the time sockets wait to recycle), but this fix feels best to me.

http/rest-client-reactive/src/test/java/io/quarkus/ts/http/restclient/reactive/RestClientReactiveMemoryLeakIT.java
Ladicek: Done.

fedinskiy (approved):

mjurc (approved):
lgtm, but let's wait for the ci

